PR #40, opened 2021-04-21
Feature/39 axis accessor

### LINKED ISSUE #39: Add AxisAccessor to return boundary information
#### Is your feature request related to a problem? Please describe

xarray does not have a built-in method to calculate the axis boundaries for input files that supply this information

For example, there might or not be boundary information in the dataset's variables (`lat_bnds`/`latitude_bnds` and `lon_bnds`/`longitude_bnds' ).

#### Describe the solution you'd like

Using `@xr.register_dataset_accessor('axis')`, add `AxisAccessor` class with method `get_bounds()`.

Example usage:
```
ds = xr.open_dataset(...)
lat_bnds = ds.axis.get_bounds('lat')
lon_bnds = ds.axis.get_bounds('lon')
```

tomvothecoder:
## Description

This PR adds the `AxisAccessor` class with methods to access axis boundary information.

- Closes #39 

## How Has This Been Tested?

- [ ] Local Pre-commit Checks
- [ ] CI/CD Build

## Checklist

If applicable:

- [ ] New and existing unit tests pass locally with my changes


xcdat/axis.py
```diff
@@ -0,0 +1,102 @@
...
+            matching_bounds_var = self._obj.data_vars.get(var)
+            if matching_bounds_var is not None:
+                break
+
+        if matching_bounds_var is None:
+            return self._calc_bounds(axis)
+
+        return matching_bounds_var
+
+    def _calc_bounds(self, axis: Axis) -> xr.DataArray:
```
tomvothecoder, line 47: This method was based on https://github.com/lee1043/XCDAT/blob/master/compare_cdat_xarray/explore_spatial_averages.ipynb
(`get_lat_bounds`)

xcdat/axis.py
```diff
@@ -0,0 +1,100 @@
...
+        else:
+            axis_bnds[:, 0] = np.concatenate(([90], axis_mid))
+            axis_bnds[:, 1] = np.concatenate((axis_mid, [-90]))
+
+        # TODO: Determine the coords value for "bnds"
+        axis_bnds_var = xr.DataArray(
+            data=axis_bnds,
+            coords={axis: np.array(axis_coords), "bnds": [0, 1]},
+            dims=[axis, "bnds"],
+            attrs={"units": axis_coords.units, "is_calculated": True},
```
tomvothecoder, line 73: `is_calculated` is set to `True` whenever `get_bounds()` has to calculate the bounds

tomvothecoder:
Hi @pochedls and @chengzhuzhang, here's my first attempt at extending xarray using the Dataset accessor decorator.

It was pretty straightforward working with the accessor decorator and writing tests for the class methods. I based the core logic of `get_bounds()` on `get_lat_bounds()` from this [notebook](https://github.com/lee1043/XCDAT/blob/master/compare_cdat_xarray/explore_spatial_averages.ipynb), which was nice because it provided me with some initial direction.

If you both can review the code and comments, that would be great!

xcdat/axis.py
pochedls: I think this will not work correctly for longitude (since longitude limits are not +/- 90). I also think the algorithm will need to be aware of whether the longitude limits are [0:360] or [-180:180]. 

The method in the example above was meant to be pretty minimal (and not necessarily robust). I think we probably need to look at [getBounds](https://github.com/CDAT/cdms/blob/master/Lib/axis.py#L2459), which is more sophisticated.

tomvothecoder:
Hey @jasonb5, it was suggested it might be a good idea for you to review this code also. I can't add you as a reviewer until you accept the collaborator invite.

xcdat/axis.py
tomvothecoder: Thanks, after doing a call stack trace starting from `getBounds`, it leads to [`genGenericBounds`](https://github.com/CDAT/cdms/blob/dd41a8dd3b5bac10a4bfdf6e56f6465e11efc51d/Lib/axis.py#L2171) which contains the core logic to handle both axis.

xcdat/axis.py
```diff
@@ -0,0 +1,204 @@
...
+            matching_bounds_var = self._obj.data_vars.get(var)
+            if matching_bounds_var is not None:
+                break
+
+        if matching_bounds_var is None:
+            return self._generate_bounds(axis)
+
+        return matching_bounds_var
+
+    def _generate_bounds(self, axis: Axis, width: float = 1) -> xr.DataArray:
```
tomvothecoder, line 58: @chengzhuzhang @pochedls This method and its children methods are based on CDAT's [`genGenericBounds` ](https://github.com/CDAT/cdms/blob/dd41a8dd3b5bac10a4bfdf6e56f6465e11efc51d/Lib/axis.py#L2171) method.

If you do a comparison, you'll see some resemblance in the code. However, I've broken up the function into modular methods, and most of the conditional statements have been refactored for readability.

xcdat/axis.py
```diff
@@ -0,0 +1,204 @@
...
+        :return: [description]
+        :rtype: np.ndarray
+
+        # TODO: If possible, refactor this method to simplify conditionals
+        """
+        min_neg_val = bounds_2d[0, 0]
+        min_pos_val = bounds_2d[0, 1]
+        second_max_pos_val = bounds_2d[-1, 0]
+        max_pos_val = bounds_2d[-1, 1]
+        max_degree_interval = 360.0
```
tomvothecoder, line 163: Let me know if these variable names make sense. I just checked what the values are and named them accordingly.

xcdat/axis.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def _calc_lat_bounds(self, bounds_2d: np.ndarray) -> np.ndarray:
+        """Calculate latitude boundaries and avoids floating point errors.
+
+        :param bounds_2d: [description]
+        :type bounds_2d: np.ndarray
+        :return: [description]
+        :rtype: np.ndarray
+
+        # TODO: If possible, refactor this method to simplify conditionals
```
tomvothecoder, line 157: If we can further refactor the conditionals, that would be nice here.

xcdat/axis.py
```diff
@@ -0,0 +1,204 @@
...
+                abs(min_neg_val - np.floor(min_neg_val + 0.5))
+                < abs(min_pos_val - min_neg_val) * 0.01
+            )
+            end_bound_near_int: bool = (
+                abs(max_pos_val - np.floor(max_pos_val + 0.5))
+                < abs(max_pos_val - second_max_pos_val) * 0.01
+            )
+            if (
+                start_bound_near_int or end_bound_near_int
+            ) and min_neg_val * max_pos_val < 0:
```
tomvothecoder, line 183: Not sure what to name the boolean variable for `min_neg_val * max_pos_val < 0`

jasonb5:
@tomvothecoder Just accepted the invite.

Might be wrong but I think you have the `_calc_lat_bounds` and `_calc_lon_bounds` reversed, lat should be [-90, 90] and lon [0, 360] or [-180, 180]. Using `get_bounds("lon")` returns the wrong left and right endpoints (90, -90).

I think adding an additional method like `gen_bounds(self, *axes: Axis) -> xr.Dataset` that calculates any missing bounds or a subset might be nice.

xcdat/axis.py
```diff
@@ -0,0 +1,204 @@
...
+            raise TypeError(f"Dataset {axis} coordinates has no attribute 'units'")
+        if "degree" not in units_attr:
+            raise ValueError(
+                f"Dataset {axis} coordinates 'units' attr, {units_attr} does not contain 'degree'"
+            )
+
+        if axis == "lon" and len(bounds.shape) == 2:
+            bounds = self._calc_lat_bounds(bounds)
+        if axis == "lat":
+            bounds = self._calc_lon_bounds(bounds)
```
jasonb5, line 83: Should these method calls be swapper?

xcdat/axis.py
jasonb5: IMHO min_left/right and max_left/right might read better.

tomvothecoder: Thanks for the suggestion, I agree and adopted this convention

pochedls:
@tomvothecoder - what do you think about adding dask as a requirement in this dev environment? I wanted to test this by looping over CMIP datasets using `open_mfdataset()`, but I think that function requires dask.

chengzhuzhang:
One item from today's discussion is that, in CDMS is able to do a longitude conversion [-180 to 180] ->[0, 360] which is convenient to deal with multiple datasets with these two different longitude range. Not sure if this is relevant to this PR, or we should discuss further when data slicing is being worked on . But for utilities like slicing a domain, this feature will come in handy.

tomvothecoder:
> One item from today's discussion is that, in CDMS is able to do a longitude conversion [-180 to 180] ->[0, 360] which is convenient when dealing with multiple datasets with these two different longitude ranges. Not sure if this is relevant to this PR, or we should discuss further when data slicing is being worked on . But for utilities like slicing a domain, this feature will come in handy.

Thanks Jill, I'm fixing up the unit tests and will see if I can get to it in this PR. Otherwise we can discuss the details further for a new PR.

tomvothecoder:
Hey @jasonb5, this is ready for a final review. 

I've done some refactoring to make the names of the methods and variables easier to understand. I also added comments in areas that might be confusing, and the entire class is covered with tests.

xcdat/axis.py
```diff
@@ -0,0 +1,245 @@
...
+        :param axis: "lat" or "lon" axis
+        :type axis: Axis
+        :param generate: If True, generate the bounds regardless if it exists in the
+        dataset or not. If False and the bounds already exist, return the existing
+        bounds, defaults to False
+        :type generate: bool, optional
+        :return: Axis bounds
+        :rtype: xr.DataArray
+        """
+        if generate:
```
tomvothecoder, line 66: Explicit flag for generating bounds

xcdat/axis.py
```diff
@@ -0,0 +1,245 @@
...
+
+        for var in bounds_vars:
+            matching_bounds_var = self._dataset.data_vars.get(var)
+            if matching_bounds_var is not None:
+                logger.info(f"Returning existing {axis} bounds from the dataset.")
+                return matching_bounds_var
+
+        raise ValueError(
+            f"{axis} bounds not found in the dataset. Pass generate=True to generate bounds."
+        )
```
tomvothecoder, line 81: If bounds don't exist and they user doesn't pass `generate=True`, it will throw a ValueError.

xcdat/axis.py
```diff
@@ -0,0 +1,245 @@
...
+        if "degree" not in units_attr:
+            raise ValueError(
+                f"Dataset {axis} coordinates value for 'units' ('{units_attr}'), does not contain 'degree' substring."
+            )
+
+        # Perform adjustments based on boundary values as needed
+        if axis == "lat":
+            bounds = self._adjust_lat_bounds(bounds)
+        if axis == "lon" and len(bounds.shape) == 2:
+            bounds = self._adjust_lon_bounds(bounds)
```
tomvothecoder, line 111: These methods are now called `_adjust` instead of ` _calculate` for clarity.

xcdat/axis.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+@xr.register_dataset_accessor("axis")
+class AxisAccessor:
+    """A class used to represent an AxisAccessor (xarray dataset accessor)
+
+    :param xarray_obj: The Dataset object to be extended
+    :type xarray_obj: xr.Dataset
+
+    Examples
```
tomvothecoder, line 16: Added examples for how to use the accessor

xcdat/axis.py
```diff
@@ -0,0 +1,243 @@
...
+        "lon": {
+            "coords": ["lon", "longitude"],
+            "bounds_vars": ["lon_bnds", "longitude_bnds"],
+        },
+    }
+
+    def __init__(self, xarray_obj: xr.Dataset):
+        self._dataset = xarray_obj
+
+    def get_bounds(self, axis: Axis, generate: bool = False) -> xr.DataArray:
```
tomvothecoder, line 52: Will be defaulting generate to `True` to always generate bounds if they do not exist, and updating param name for clarity

xcdat/axis.py
tomvothecoder: Moving this below the logic to check for matching bounds

xcdat/axis.py
tomvothecoder: Removing this to avoid frustrating runtime errors for most use cases.

xcdat/axis.py
```diff
@@ -0,0 +1,270 @@
...
+            # For (-180, 180), if either bound is near an integer value, round both integers
+            # Otherwise it is not needed if all values are positive (0, 360)
+            if (
+                min_left_val_near_int or max_right_val_near_int
+            ) and min_bound_left * max_bound_right < 0:
+                bounds[0, 0] = np.floor(min_bound_left + 0.5)
+                bounds[0, 1] = np.floor(max_bound_right + 0.5)
+            else:
+                if max_bound_right > min_bound_left:
+                    bounds[-1, 1] = round(min_bound_left + max_degree_interval, 4)
```
tomvothecoder, line 266: I had to add `round()` here or I cannot test the result since it can be < -0.0009999999

tomvothecoder: This might explain why the bounds are exact in some datasets?

jasonb5: Would it be better to move `round` out of the implementation and into the test? This might surprise the user if they have a longitude axis starting from a small number <4 decimals, not sure how common this is.

Example: This produces bounds where the last interval is not complete.
```python
ds = xr.Dataset(None, coords={"lon": xr.DataArray(np.arange(0.05125, 360, 0.5), attrs={"units": "degrees_west"})})
x = ds.axis.get_bounds("lon")
x[:,1]-x[:,0]
```

tomvothecoder: Thanks for the suggestion. I agree, this is a special test case and it would be better to round in the test.

tomvothecoder: Fixed by removing `round()` and using[ `numpy.allclose()`](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html#numpy.allclose) to compare two float arrays

PR #81, opened 2021-07-06
Add wrappers for opening datasets and data variables

### LINKED ISSUE #77: Support for time units of "months since..."
### Is your feature request related to a problem? If yes, Please describe

One feature of CDAT is its ability to flexibly read in a large number of time units. In testing some code, I realize that xarray does not natively support "months since ..."

```
import xarray as xr

fn = 'rss_4.0_ttt_197901-202012.nc'
ds = xr.open_dataset(fn)
```

> ValueError: unable to decode time units 'months since 1979-01-01' with "calendar 'standard'". Try opening your dataset with decode_times=False or installing cftime if it is not installed.

### Describe your proposed solution
Ideally, we would be able to open and decode files like this, perhaps with other packages (e.g., [nctime](http://prodiguer.github.io/nctime/index.html)). Could a fix for this issue (and potentially other fixes) be included in xcdat utilities (e.g., [open_datasets](https://github.com/tomvothecoder/xcdat/blob/main/xcdat/utils.py#L16))? 

### Describe alternatives solutions you've considered
One obvious alternative is to force the user to use the `decode_times=False` option and deal with the time axis on their own or deal with the axis information after loading the dataset. 

### Additional context
If we had a utility to help load data, we could also add axis information (e.g., bounds) when opening the dataset and include the option of stripping out extraneous variables, e.g., `tas = xcdat.open_dataset(fn, variable='tas')`.

Additional tasks based on 7/21 meeting
- [x] Take into account dataset's calendar when decoding time

### LINKED ISSUE #80: Add Dataset and data variable wrappers to apply common operations upon opening
### Is your feature request related to a problem? If yes, Please describe

CDAT users commonly work on a single `TransientVariable` of a dataset, which is equivalent to a single xarray `Dataset` data variable of type `DataArray`. However, an xarray data variable does not link to other data variables. This means that the extracted data variable does not have access to any existing axis bounds (also data variables).

### Describe your proposed solution

Acceptance Criteria
1. Add `open_dataset` function to open a dataset and generate bounds if they don't exist. 
   - [x] Extensible for additional operations.
   - [x] Add call to generate bounds if it doesn't exist
2. Add `open_variable` function to open a data variable with axis bounds (passed from parent Dataset). Replicates CDAT `TransientVariable`.
   - [x] Extensible for additional operations
   - [x] Add DataArray accessor to add additional attributes to DataArray class (to store bounds info)
   - [x] Show all attributes of accessor class (just use `__dict__`)
   - [x] Take into account dataset's calendar (EDIT: not needed, but keep calendar type in encoding if available)
      - I don’t think the calendar type affects the algorithm for generating a date range. Non-CF units “months since…” or “years since…” don’t factor the number of days in a month.
      - pd.date_range() generates proleptic_gregorian datetime values

### Describe alternatives solutions you've considered

- On creation of the variable (easiest for the user): tas = ds.tas
  - **Not possible because this is native xarray syntax, which requires extending via subclassing (not recommended) or xarray accessors.**
- On dataset open: ds = xcdat.load_dataset(filename) [this would embed the bounds in the relevant variables so that they are included in ds.tas]
  - **Not possible, I found Dataset data variables don't preserve custom attributes via DataArray accessors**
 
### Additional context

- Extracts some tasks for #77

tomvothecoder:
## Description

- Closes #80 
- Closes #77 

## Checklist

If applicable:


xcdat/dataset.py
```diff
@@ -0,0 +1,43 @@
...
+
+    >>> from xcdat import open_dataset
+    >>> ds = open_dataset("file_path")
+    """
+    ds = xr.open_dataset(path)
+
+    ds.axis.get_bounds("lat", allow_generating=True)
+    ds.axis.get_bounds("lon", allow_generating=True)
+
+    return ds
```
tomvothecoder, line 43: You can refer to the docs here: https://xcdat.readthedocs.io/en/feature-80-open-dataset/generated/xcdat.dataset.html or look directly at this docstring for usage examples.

xcdat/dataset.py
```diff
@@ -0,0 +1,43 @@
...
+    in-memory copy you are manipulating in xarray is modified: the original file
+    on disk is never touched.
+
+    Examples
+    --------
+    Import and call module:
+
+    >>> from xcdat import open_dataset
+    >>> ds = open_dataset("file_path")
+    """
```
tomvothecoder, line 37: Example usage for `open_dataset()`. Viewable in the docs [here](https://xcdat.readthedocs.io/en/feature-80-open-dataset/generated/xcdat.dataset.html).

xcdat/variable.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Get lat bounds attribute:
+
+    >>> tas.bounds.lat
+
+    Get lon bounds attribute:
+
+    >>> tas.bounds.lon
+
+    Get time bounds attribute:
```
tomvothecoder, line 47: Usage examples for `open_variable()`. Viewable in docs [here](Example usage for `open_dataset()`. Viewable in the docs [here](https://xcdat.readthedocs.io/en/feature-80-open-dataset/generated/xcdat.dataset.html).

xcdat/variable.py
```diff
@@ -0,0 +1,103 @@
...
+            missing_bnds.append(attr)
+
+    if missing_bnds:
+        raise ValueError(
+            f"The Dataset is missing the following axis bounds: {', '.join(missing_bnds)}. "
+            "Make sure to open the Dataset using `xcdat.dataset.open_dataset()` to "
+            "autogenerate bounds before passing to this function."
+        )
+    else:
+        data_var.bounds._copy_from_parent(dataset)
```
tomvothecoder, line 68: At the moment, `open_variable()` only checks for missing `lat_bnds` or `lon_bnds` in the Dataset because there is no function to generate missing `time_bnds`.

xcdat/dataset.py
```diff
@@ -0,0 +1,43 @@
...
+    ds = xr.open_dataset(path)
```
pochedls: @tomvothecoder - One xarray feature that I expect we will frequently rely on is the [open_mfdataset](http://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html) functionality. I'm wondering if we could automatically decide whether to use `open_dataset` or `open_mfdataset` by checking if the `path` parameter is a file or a directory, e.g.,

```
import os
...
if os.path.isdir(path):
    ds = xr.open_mfdataset(path + '*')
elif os.path.isfile(path):
    ds = xr.open_dataset(path)
else:
    raise ValueError('Input path (' + path + ') is invalid.')
```

xcdat/dataset.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    ds = xr.open_dataset(path, decode_times=False)
+    ds = _decode_time_units(ds)
+
+    ds.axis.get_bounds("lat", allow_generating=True)
+    ds.axis.get_bounds("lon", allow_generating=True)
+    return ds
+
+
+def _decode_time_units(ds: xr.Dataset):
```
tomvothecoder, line 48: This function handles #77

xcdat/dataset.py
```diff
@@ -0,0 +1,102 @@
+"""Dataset module for Dataset related functions."""
+import pandas as pd
+import xarray as xr
+
+from xcdat import axis  # noqa: F401
+
+
+def open_dataset(path: str) -> xr.Dataset:
```
tomvothecoder, line 8: This function handles #80

xcdat/dataset.py
```diff
@@ -0,0 +1,102 @@
...
+    on disk is never touched.
+
+    Examples
+    --------
+    Import and call module:
+
+    >>> from xcdat import open_dataset
+    >>> ds = open_dataset("file_path")
+    """
+    ds = xr.open_dataset(path, decode_times=False)
```
pochedls, line 40: @tomvothecoder - One xarray feature that I expect we will frequently rely on is the [open_mfdataset](http://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html) functionality. I'm wondering if we could automatically decide whether to use `open_dataset` or `open_mfdataset` by checking if the `path` parameter is a file or a directory, e.g., something like this:

```
import os
...
if os.path.isdir(path):
    ds = xr.open_mfdataset(path + '/*')
elif os.path.isfile(path):
    ds = xr.open_dataset(path)
else:
    raise ValueError('Input path (' + path + ') is invalid.')
```

pochedls: This might require adding dask as a [dependency](https://github.com/tomvothecoder/xcdat/blob/main/conda-recipe/meta.yaml)

tomvothecoder: Thanks for bringing this up @pochedls.

I need to familiarize myself with `open_mfdataset()` to see if there are any caveats/complexities in regards to generating bounds and fixing time units. I'm assuming we'd use this function to join multiple files within a single dataset as a single xarray `Dataset` object, which means they share the same coordinates and bounds?

pochedls: @tomvothecoder - yes you are exactly right. A large fraction of climate data spans multiple netCDF files (each file typically has a chunk of time, e.g., 1970 - 1979, 1980 - 1989, etc.).

xcdat/dataset.py
```diff
@@ -0,0 +1,221 @@
...
+            attrs=dataset["time"].attrs,
+        )
+        decoded_time.encoding = {
+            "source": dataset.encoding.get("source"),
+            "dtype": time.dtype,
+            "original_shape": decoded_time.shape,
+            "units": units_attr,
+            # pandas.date_range() returns "proleptic_gregorian" by default
+            "calendar": time.attrs.get("calendar", "proleptic_gregorian"),
+        }
```
tomvothecoder, line 218: This might need a closer look. From my understanding, the number of days in a month should not matter if you are generating date ranges in months or years. Therefore, the calendar type has no influence on the operation.

tomvothecoder:
Hey @pochedls and @jasonb5, this PR is ready for final review. Thanks!

xcdat/bounds.py
```diff
@@ -1,46 +1,60 @@
-"""Coordinate module that contains coordinate related functions (e.g., getting bounds)."""
```
tomvothecoder, line 1: This is one of the main files to focus on

xcdat/dataset.py
```diff
@@ -0,0 +1,221 @@
+"""Dataset module for functions related to an xarray.Dataset."""
```
tomvothecoder, line 1: This is one of the main files to focus on

xcdat/variable.py
```diff
@@ -0,0 +1,60 @@
+"""Variable module for functions related to an xarray.Dataset data variable"""
```
tomvothecoder, line 1: This is one of the main files to focus on

xcdat/bounds.py
```diff
@@ -1,46 +1,60 @@
...
+"""Bounds module for functions related to coordinate bounds."""
```
tomvothecoder, line 1: This is one of the main files to focus on

tomvothecoder:
I just found out an interesting behavior related to the bounds accessor class and its attributes.

When you open a DataArray/Dataset with an accessor class and add attributes, then make a copy of it, the attributes don't propagate up to the copy (which means you lose them). **This sounds like it will cause problems with data integrity as you work with variables, make copies, pass to functions, try to access the bounds, etc.**

### Code example:
```
# Open and set the bounds accessor class attributes
------------------------------------------------------
>>> ts = open_variable(ds, "ts")

# Access an attribute
------------------------------------------------------
>>> print(ts.bounds.lat)
<xarray.DataArray 'lat_bnds' (lat: 145, bnds: 2)>
array([[-90.   , -89.375],
       [-89.375, -88.125],
       [-88.125, -86.875],
       ...,
       [ 86.875,  88.125],
       [ 88.125,  89.375],
       [ 89.375,  90.   ]])
Coordinates:
  * lat      (lat) float64 -90.0 -88.75 -87.5 -86.25 ... 86.25 87.5 88.75 90.0
Dimensions without coordinates: bnds

# Make a copy of the variable
------------------------------------------------------
>>> ts_copy = ts.copy()

# Try to access the bounds attribute again. It's lost now.
------------------------------------------------------
>>> print(ts_copy.bounds.lat)
None

```

### Alternate Solution
The alternative solution is to store bounds inside the DataArray **coordinates**  instead of attributes of the bounds accessor class.

Pros:
- Bounds are persistent even when copying since they are stored as coords (not cached accessor property) -- big deal
- Use native xarray syntax to access bounds instead decorator (`ts.lat_bnds` vs. `ts.bounds.lat`)

Cons:
- xarray might store bounds as data variables (not sure if that's default or NetCDF behavior) -- not a big deal though
- Have to propagate the bounds dimension (can be "bnds" or "bounds") -- not hard to check and set dynamically
- Have to check which bounds are set -- `cf_xarray` can handle this

```
# 1. Make sure bounds are stored in the Dataset's coordinates
# ------------------------------------------------------------
# Not sure if xarray reads in bounds as data variables instead of coordinates by default, or the NetCDF files store 
# them that way.

>>> ds.coords
Coordinates:
  * time       (time) datetime64[ns] 1850-01-16T12:00:00 ... 2005-12-16T12:00:00
    time_bnds  (time, bnds) datetime64[ns] 1850-01-01 1850-02-01 ... 2006-01-01
  * lat        (lat) float64 -90.0 -88.75 -87.5 -86.25 ... 86.25 87.5 88.75 90.0
    lat_bnds   (lat, bnds) float64 -90.0 -89.38 -89.38 ... 89.38 89.38 90.0
  * lon        (lon) float64 0.0 1.875 3.75 5.625 ... 352.5 354.4 356.2 358.1
    lon_bnds   (lon, bnds) float64 -0.9375 0.9375 0.9375 ... 357.2 357.2 359.1

# 2. Copy bounds into each data variable's coordinates
# ------------------------------------------------------------
ts = ds.ts.copy()
ts = ts.expand_dims(bnds=np.array([0,1]))
ts = ts.assign_coords({"lon_bnds": ds.lon_bnds, "lat_bnds": ds.lat_bnds, "time_bnds": ds.time_bnds})

# 3. Access bounds using default xarray syntax 
# ------------------------------------------------------------
>>> ts.coords
Coordinates:
  * bnds       (bnds) int64 0 1
  * time       (time) datetime64[ns] 1850-01-16T12:00:00 ... 2005-12-16T12:00:00
  * lat        (lat) float64 -90.0 -88.75 -87.5 -86.25 ... 86.25 87.5 88.75 90.0
  * lon        (lon) float64 0.0 1.875 3.75 5.625 ... 352.5 354.4 356.2 358.1
    lon_bnds   (lon, bnds) float64 -0.9375 0.9375 0.9375 ... 357.2 357.2 359.1
    lat_bnds   (lat, bnds) float64 -90.0 -89.38 -89.38 ... 89.38 89.38 90.0
    time_bnds  (time, bnds) datetime64[ns] 1850-01-01 1850-02-01 ... 2006-01-01

>>> ts.lat_bnds # or ts.coords["lat_bnds"]
<xarray.DataArray 'lat_bnds' (lat: 145, bnds: 2)>
array([[-90.   , -89.375],
       [-89.375, -88.125],
       [-88.125, -86.875],
       ...,
       [ 86.875,  88.125],
       [ 88.125,  89.375],
       [ 89.375,  90.   ]])
Coordinates:
  * bnds      (bnds) int64 0 1
  * lat       (lat) float64 -90.0 -88.75 -87.5 -86.25 ... 86.25 87.5 88.75 90.0
    lat_bnds  (lat, bnds) float64 -90.0 -89.38 -89.38 ... 89.38 89.38 90.0
```

Related GitHub Issues:
- https://github.com/pydata/xarray/issues/3268
- https://github.com/pydata/xarray/issues/1475

PR #87, opened 2021-07-22
Feature/46 spatial averaging

### LINKED ISSUE #46: Add grid-aware geospatial averaging (weighted)

Flexible (e.g., in averaging order), accurate, weighted averaging over a precise domain 

Acceptance criteria:
- [x] Averaging options include weighted, unweighted, or custom weights
- [x] Assigns masked data zero-weight
- [x] Ensure we can convert between 0:360 longitude and -180:180 longitude
- [x] Allow users to specify a specific averaging region (i.e., routine calculates proper weights for sub-region)
- [x] Performs horizontal averaging, but can handle higher dimensional data (i.e., x(time, altitude, lat, lon) -> x(time, altitude) after averaging)
- [x] Can average over latitude, longitude, or both axes
- [x] Operates on datasets with common or cf_compliant metadata (i.e., inclusion of lat/latitude and lon/longitude)
- [x] ~~Axes to be averaged can be specified in CDAT notation (i.e., axis='xy')~~ going with "lat" and "lon" instead
- [ ] ~~#114~~ This can be done later

### Additional context

CDAT equivalent:
- [genutil.averager](https://github.com/CDAT/genutil/blob/master/Lib/averager.py#L856)
- [cdutil.region](https://github.com/CDAT/cdutil/blob/master/cdutil/region.py)
- [genutil.area_weights](https://github.com/CDAT/genutil/blob/master/Lib/averager.py#L178)
- [cdms2.getWeights](https://github.com/CDAT/cdms/blob/master/Lib/grid.py#L449)
- [cdms.getBounds](https://github.com/CDAT/cdms/blob/master/Lib/axis.py#L1481)

Prototype: 
- [Spatial Averaging in xarray and CDAT](https://github.com/lee1043/XCDAT/blob/master/compare_cdat_xarray/explore_spatial_averages.ipynb)

Resources
- https://gist.github.com/dennissergeev/60bf7b03443f1b2c8eb96ce0b1880150

pochedls:
## Description

- Closes #46

## How Has This Been Tested?

- [ ] Local Pre-commit Checks
- [ ] CI/CD Build

## Checklist

If applicable:

- [ ] New and existing unit tests pass locally with my changes


xcdat/averager.py
```diff
@@ -0,0 +1,458 @@
...
+    lon = np.array(lon)  # convert to array
+    # convert to specified longitude axis orientation
+    if totype == 'to360':
+        inds = np.where(lon < 0)[0]
+        lon[inds] = lon[inds] + 360
+    elif totype == 'to180':
+        inds = np.where(lon > 180)[0]
+        lon[inds] = lon[inds] - 360
+    # switch back to original data type
+    # REVIEW: Is there an easier way to be flexible to the datatype provided?
```
pochedls, line 120: There must be a better way to do this - right?

xcdat/averager.py
```diff
@@ -0,0 +1,458 @@
...
+                                   coords=xdim.coords)
+            dim_wts.name = dim + '_wts'
+            axis_weights[dim] = dim_wts
+        return axis_weights
+
+    # just check a valid weight option is provided
+    # if invalid, throw error, else continue (with weighted logic)
+    # REVIEW: Should we throw a different exception type?
+    if weight not in ['weighted', 'unweighted']:
+        raise ValueError('Incorrect weight option. Choose either "weighted"' +
```
pochedls, line 204: What exception type should this be?

tomvothecoder:
Not sure what happened, but the `main` branch reverted back to an old commit. Luckily my local `main` was up to date so I force pushed back the changes.

Another odd thing is that your PR is showing many old commits and there's a large diff with `main`, even though `main` has the same changes. I'll see if I can fix this.

pochedls:
> Not sure what happened, but the `main` branch reverted back to an old commit. Luckily my local `main` was up to date so I force pushed back the changes.
> 
> Another odd thing is that your PR is showing many old commits and there's a large diff with `main`, even though `main` has the same changes. I'll see if I can fix this.

Sorry about the trouble here. I think this was because I initially did `git push -f` instead of `git push -f origin feature/46-spatial-averaging`. I thought I needed the `-f` flag from [this thread](https://stackoverflow.com/questions/39399804/updates-were-rejected-because-the-tip-of-your-current-branch-is-behind-its-remot).

tomvothecoder:
@pochedls Ahh, no worries. Just turned off force push capabilities on `main`.  Let me try switching branches back and forth to see if the commit list and diff update.

xcdat/averager.py
tomvothecoder: Here's going to 180: https://stackoverflow.com/a/53561230
xarray also provides the same formula: https://xarray.pydata.org/en/stable/generated/xarray.DataArray.assign_coords.html
```
da.coords['lon'] = (da.coords['lon'] + 180) % 360 - 180
da = da.sortby(da.lon)
```
You can do something similar for going to 360.

The benefit of using this formula is that you don't to filter using `.where()`, which might make it quicker.

xcdat/averager.py
tomvothecoder: `ValueError` is correct

xcdat/averager.py
```diff
@@ -0,0 +1,458 @@
...
+        Dataarray that has been reduced via a weighted average.
+
+    Notes
+    -----
+    Needs to be evaluated for speed and compared to native xarray and CDAT
+    functionality.
+
+    """
+    # get masked data
+    masked = np.isnan(xvariable)
```
pochedls, line 322: This step is a huge hit on performance. It can be skipped if there are no missing values. xarray may have built-in functionality to flag this (I think for some arrays `xvariable.mask` will return an array or a boolean, but I'm not sure how bullet-proof this is). I also think it is possible to include a masked array with a dataarray, so it may be necessary to include logic to check for (and work on) masked arrays, too.

xcdat/averager.py
tomvothecoder: Here's a refactored version of this function.

Notes:

- Generally, it is better to work directly on the `xarray.DataArray`  instead of converting it into a `numpy.ndarray`. You maintain the attributes of the `DataArray` and don't need to perform type conversion. If xarray does not support an operation, you can still perform `numpy.ndarray` operations directly on the `xarray.DataArray` due to the interoperability.

My refactored function:
- Pass the entire xarray DataArray instead of just the `lon` coordinate. 
- The mid-point is -180, not 180 like the original `_switch_lon_bounds()`.

https://github.com/tomvothecoder/xcdat/blob/911419a8580688cbdad4a362d72e6e804b387272/xcdat/average_tv.py#L18-L77

xcdat/averager.py
```diff
@@ -0,0 +1,458 @@
...
+    min_grid_lon_bounds = np.min(grid_lon_bounds, axis=1)
+    max_grid_lon_bounds = np.max(grid_lon_bounds, axis=1)
+    # iterate over regional bounds and check to see if they fall within
+    # any grid cell bounds.
+    in_bounds = True
+    for lon in region_lon_bounds:
+        inds = np.where((min_grid_lon_bounds <= lon) &
+                        (max_grid_lon_bounds >= lon))[0]
+        if len(inds) == 0:
+            in_bounds = False
```
tomvothecoder, line 85: This may be a performance bottleneck since it is looping through each element of the ndarray, then performing a call against two numpy ndarrays using a `.where()`.

xcdat/averager.py
```diff
@@ -0,0 +1,458 @@
...
+    # If .bounds had a method to return a list of bounds entries, we could create weights for all dimensions
+# May want to keep the axis weights or full weight array (i.e., add this information into the variable dataarray properties)
+# Is xvariable a good name?
+# Need to write unit tests
+# Can remove logging around bounds found in dataset
+# Functions are too specific to "lat" and "lon" (may have issues if dims are latitude / longitude)
+# Should remove these notes and outline...
+
+
+def _validate_region_lon_bounds(grid_lon_bounds, region_lon_bounds):
```
tomvothecoder, line 32: Here's a refactored version that uses xarray syntax (pretty much the same as numpy)

https://github.com/tomvothecoder/xcdat/blob/1471ac8b4881379de37f9ea964596df708a745b8/xcdat/average_tv.py#L79-L133

xcdat/averager.py
```diff
@@ -0,0 +1,458 @@
...
+    if weight not in ['weighted', 'unweighted']:
+        raise ValueError('Incorrect weight option. Choose either "weighted"' +
+                         ' or "unweighted"')
+
+    # note that xvariables.bounds does not return a list
+    # so for now this is hard-coded
+    bounds = {}
+    bounds['time'] = xvariable.bounds.time.copy(deep=True)
+    bounds['lat'] = xvariable.bounds.lat.copy(deep=True)
+    bounds['lon'] = xvariable.bounds.lon.copy(deep=True)
```
tomvothecoder, line 212: With my recent merge of the `feature/80-open-dataset` branch, bounds are now referenced as coordinates.

I should probably turn the chunk of code below for retrieving bounds dynamically into a standalone method (cf_xarray has only one for Datasets since DataArrays weren't designed with coordinate bounds in mind).

https://github.com/tomvothecoder/xcdat/blob/1471ac8b4881379de37f9ea964596df708a745b8/xcdat/average_tv.py#L42-L54

xcdat/averager.py
tomvothecoder: You can extract a masked array using [xarray.DataArray.isnull](https://xarray.pydata.org/en/v0.19.0/generated/xarray.DataArray.isnull.html), which fills missing values with `np.nan`.

If we need to replace `np.nan` with another value, there is [xarray.DataArray.fillna](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.fillna.html).

> I also think it is possible to include a masked array with a dataarray, so it may be necessary to include logic to check for (and work on) masked arrays, too.

I'm not  sure if this is possible given that `DataArrays` can only store a single `np.ndarray` for its main data. The masked `DataArray` can only be stored in the coordinates, which doesn't sound right.

Reference: http://xarray.pydata.org/en/stable/generated/xarray.DataArray.to_masked_array.html

xcdat/averager.py
```diff
@@ -0,0 +1,458 @@
...
+    #         that does this operation.
+    ntile = []
+    for i, dim in enumerate(weights.dims):
+        if len(weights[dim]) != len(xvariable[dim]):
+            ntile.append(len(xvariable[dim]))
+        else:
+            ntile.append(1)
+    weights = np.tile(weights, ntile)
+    # zero out missing data
+    weights[masked] = 0.
```
tomvothecoder, line 341: Can be replaced with http://xarray.pydata.org/en/stable/generated/xarray.DataArray.fillna.html

xcdat/averager.py
```diff
@@ -0,0 +1,458 @@
...
+        if not _validate_region_lon_bounds(xvariable.bounds.lon, lon_bounds):
+            if np.max(xvariable.lon) <= 180:
+                lon_bounds = _switch_lon_bounds(lon_bounds, totype='to180')
+            else:
+                lon_bounds = _switch_lon_bounds(lon_bounds, totype='to360')
+            # after region bounds are adjusted, check to ensure they are
+            # consistent with the dataarray axes
+            if not _validate_region_lon_bounds(xvariable.bounds.lon,
+                                               lon_bounds):
+                # REVIEW: Should we throw a more specific error?
```
tomvothecoder, line 443: ValueError is correct. We can create our own custom xcdat errors if needed, but low priority.

xcdat/averager.py
tomvothecoder, line 204: Input validation should be at the top of the function.

tomvothecoder:
I am closing this PR in favor of #95, which has been squashed to a single commit and rebased on `main`.

Issue #85, opened 2021-07-16
Support Climatology bounds


chengzhuzhang:
Feature request copied from @golaz:

### Is your feature request related to a problem? If yes, Please describe
Climatology files have a special type of time bounds:
```
    double climatology_bounds(time, nbnd) ;
            climatology_bounds:long_name = "time interval endpoints" ;
            climatology_bounds:units = "days since 0001-01-01 00:00:00" ;
            climatology_bounds:calendar = "noleap" ;
```
While they look like regular bounds, they are more subtle. For more details, see :http://cfconventions.org/cf-conventions/cf-conventions.html
```
dimensions:
  time=4;
  nv=2;
variables:
  float temperature(time,lat,lon);
    temperature:long_name="surface air temperature";
    temperature:cell_methods="time: minimum within years time: mean over years";
    temperature:units="K";
  double time(time);
    time:climatology="climatology_bounds";
    time:units="days since 1960-1-1";
  double climatology_bounds(time,nv);
data:  // time coordinates translated to date/time format
  time="1960-4-16", "1960-7-16", "1960-10-16", "1961-1-16" ;
  climatology_bounds="1960-3-1",  "1990-6-1",
                     "1960-6-1",  "1990-9-1",
                     "1960-9-1",  "1990-12-1",
                     "1960-12-1", "1991-3-1" ;
```

Focusing on the first set of bounds: "1960-3-1", "1990-6-1"

The climatology spans the years 1960 to 1990 (inclusive), and the months of March, April, May.

It would be best if xCDAT could handle regular time series bounds, as well as climatology bounds transparently. There will be a need for an auxiliary function that computes the time span between the bounds (to use as weights for time averaging). It’s doable, but a little trickier for climatology bounds.

One additional complexity is that climatology bounds can also apply to diurnal averages.

We want to be able to perform time averaging operations on climatology data just like regular time series data. For example:

- Average 10 yr climo into 50 yr climo.

- Average monthly climo into seasonal or annual.

- Average hourly diurnal climo into 6-hourly diurnal climo.

-  Etc.

If xCDAT is designed in a flexible manner from the on-set, all of that becomes possible without any special treatment. If not, it will be a pain to add later (and lead to a more convoluted and harder to maintain  code.

gleckler1:
@chengzhuzhang @tomvothecoder   I see this is still a low priority.  Can we briefly discuss this at the next meeting?

tomvothecoder:
[...]

PR #107, opened 2021-09-21
Add temporal averaging functionality (time series avg, climatology and departure)

### LINKED ISSUE #47: Add temporal averaging functionality (time series avg, climatology and departure)

Temporal averaging for time series, climatology, and departures for different time frequencies with the option for weighted or unweighted means.

### Acceptance criteria

Bounds
- [x] #90  
- [x] Uses time bounds to recompute time, or throw error if it doesn't exist.

Weights
- [x] Weighted means (xcdat default)
  - [x] If time bounds exist, generates weights using it
  - [x] If time bounds don't exist, raise error that time bounds must be added.
  - [x] Considers leap years (uses datetime component `days_in_month`). More info [here](http://xarray.pydata.org/en/stable/examples/monthly-means.html)
  - [x] Handles getting weights for all time intervals (hourly, daily, monthly, seasonal, yearly)
- [x] Unweighted means

Timeseries Averaging
 - [x] annual
 - [x] monthly
 - [x] season
 - [x] custom season
 - [x] day
 - [x] hour
 - [ ]  nHour (nice-to-have)
 
Climatology
- [x] month (aka ANNUALCYCLE) -- monthly means for each month of the year for all years
- [x] season (aka SEASONALCYCLE) -- means for the 4 predefined seasons for all years
  - [x] Handle DJF season
      - [x] continuous (previous year December, remove incomplete seasons) (xcdat default)
      - [x] discontinuous (same year December) (xarray default, there is no continuous variant for seasonal grouping)
- [x] custom season
- [x] day -- daily means for each day of the year for all years
- [ ] diurnalNN (nice-to-have)

Departure
- [x] Handles monthly, seasonal, and annual climatology
- [x] Handles hourly climatology

Edge Cases
- [ ] Time and output time for climatology requires special handling. The month and day should be centered, but the year needs to be selected (first year, last year?). The mid year may not be an integer.
- [ ] Confirm decoding non-CF time units produces the correct calendar type, otherwise use existing calendar

### Additional context

https://stackoverflow.com/questions/59234745/is-there-any-easy-way-to-compute-seasonal-mean-with-xarray
https://cdat.llnl.gov/documentation/utilities/utilities-1.html
Prototypes:
- [Annual Cycle: CDAT vs. Xarray](https://github.com/lee1043/XCDAT/blob/master/compare_cdat_xarray/annual_cycle.ipynb)
- [Annual Cycle Climatology Test: CDAT vs. Xarray](https://github.com/lee1043/XCDAT/blob/master/compare_cdat_xarray/annual_cycle_climatology.ipynb)
- [Annual Cycle Departure: CDAT vs. Xarray](https://github.com/lee1043/XCDAT/blob/master/compare_cdat_xarray/annual_cycle_departure.ipynb)

tomvothecoder:
## Description

- Closes #47 

This PR adds temporal averaging functionality, including calculating time series averaging, climatology, and departures.

### Summary of Changes
- Operates on a data variable within a dataset, and returns a dataset
- Stores metadata regarding temporal averaging in the data variable's attributes
- Supported grouping frequencies:
  - Time series average:
    * "year", "season", "month", "day", "hour"
        * "year" groups by year
        * "season" groups by (year, season)
        * "month" groups by (year, month)
        * "day" groups by (year, month, day)
        * "hour" groups by (year, month, day, hour)
  - Climatology and departures:
    * "season", "month", "day"
        * "season" groups by season
        * "month" groups by month
        * "day" groups by (month, day)
- Support for weighted and unweighted averaging (defaults to True)
- Support for centering time coordinates using time bounds (defaults to true)
- Configuration specific to "season" frequency
  -  Predefined seasons:
     - "dec_mode" - mode for the seasons that includes December ("DJF" or "JFD") (defaults to "DJF")
     - "drop_incomplete_djf" - if "dec_mode" is "DJF", drop incomplete "DJF" seasons (defaults to False)
  - Custom seasons: 
     - "custom_seasons" - list of lists containing the 3-letter string representation of months (e.g., "Jan")

## Validation
If you are interested in the validation of this feature against CDAT, please checkout the notebooks found here: https://github.com/XCDAT/xcdat_test/pull/28

## Future Enhancements
Feature Specification Documentation: https://docs.google.com/document/d/1klHh5LLYcmNSopvptSVYmcgtUVLFnLMueRhfCF6tbaU/edit#

- Time series averaging frequency - "nHour"
- Climatology frequency - "diurnalNN"
- #201 
- #202 
- #85 
- Update seasonal cycle departures to group on year and season like CDAT for seasonal anomalies
- Handling missing data
  - User may want to set criteria for excluding locations entirely from a climatology if the amount of missing data is excessive or unrepresentative.
  - CDAT allowed the user to set two criteria for excluding a location (i.e., setting it to “missing”)  to mask out locations with insufficient data. 

## Checklist

If applicable:


xcdat/temporal_avg.py
```diff
@@ -0,0 +1,844 @@
...
+                f"Incorrect `djf_type` argument. Supported DJF types include: {djf_types}"
+            )
+
+        self.mode = mode
+        self.freq = freq
+        self.is_weighted = is_weighted
+        self.djf_type = djf_type
+
+    def _handle_custom_seasons(self, custom_seasons: List[List[str]]) -> List[str]:
+        """Validates the user-specified list of custom seasons.
```
tomvothecoder, line 399: Change description to `Validates the list of custom seasons and handles creating them.'

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,844 @@
...
+
+        The method checks to see if the custom seasons include the 12 predefined
+        month strings.
+
+        Afterwards, it concatenates each sublist into a string to represent
+        the custom season. For example, ["Jan", "Feb", "Mar] becomes
```
tomvothecoder, line 405: Fix missing quotation mark

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,899 @@
...
+DateTimeComponent = Literal["hour", "day", "month", "season", "year"]
+
+# DJF Season Types
+# ================
+# Type alias for the type of DJF seasons.
+DJFType = Literal["cont", "discont"]
+# Tuple of DJF season type params.
+DJF_TYPES = get_args(DJFType)
+
+
```
tomvothecoder, line 44: Remove space

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,913 @@
...
+# Type alias for supported grouping frequencies.
+Frequency = Literal["hour", "day", "month", "season", "year"]
+#: Tuple of supported grouping frequencies.
+FREQUENCIES = get_args(Frequency)
+
+# Configuration specific to the "season" frequency.
+SeasonConfig = TypedDict(
+    "SeasonConfig",
+    {
+        "winter_season": Literal["DJF", "JFD"],
```
pochedls, line 32: I think we should make this `boreal_winter_season`

tomvothecoder: @pochedls Is this the technical term you found for this key?

pochedls: It seems like we should base this off [meteorological winter](https://en.wikipedia.org/wiki/Season#Meteorological), which unfortunately is different in each hemisphere, which is why I appended "boreal." I think it is okay if this is a little long, since we can define the default value (DJF) which I assume won't change that much (I guess if you are passing in a climatology you'd have to change the default value?).

tomvothecoder: `boreal_winter_season` works since there doesn't seem to be a more generalized term for that season across hemispheres.

> (I guess if you are passing in a climatology you'd have to change the default value?).
If the climatology is calculated using xcdat, this attribute is stored in the variable DataArray for reference.
Otherwise, the user would have to change the value if it should be "JFD".

chengzhuzhang: boreal_winter_season is accurate. or `NH_winter`, which is short for `Northern_Hemisphere_Winter`,  how about `dec_mod` shorter?

tests/test_xcdat.py
```diff
@@ -1,14 +1,17 @@
...
 
 from tests.fixtures import generate_dataset, lat_bnds
 from xcdat.xcdat import XCDATAccessor
 
 
 class TestXCDATAccessor:
-    # NOTE: We don't have to test this class in-depth because its methods
+    # NOTE: We don't have to perform in-depth testing because its methods
     # are abstracted public methods from other classes, which already has more
     # comprehensive testing.
```
tomvothecoder, line 13:
```suggestion
    # NOTE: We don't have to perform in-depth testing because the methods
    # are abstracted public methods from other classes, which already has more
    # comprehensive testing.
```

xcdat/temporal_avg.py
tomvothecoder: Thanks @chengzhuzhang. `dec_mode` is a good alternative so that the key doesn't have to be associated with a "winter" season.

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,950 @@
...
+            'operation': 'temporal_avg',
+            'mode': 'climatology',
+            'freq': 'season',
+            'groupby': 'season',
+            'weighted': 'True',
+            'centered_time': 'True',
+            'dec_mode': 'DJF',
+            'drop_incomplete_djf': 'True'
+        }
+        """
```
tomvothecoder, line 273: Here are docstring examples for calling temporal averaging.

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,950 @@
...
+            'operation': 'temporal_avg',
+            'mode': 'departure',
+            'frequency': 'season',
+            'groupby': 'season',
+            'weighted': 'True',
+            'centered_time': 'True',
+            'dec_mode': 'DJF',
+            'drop_incomplete_djf': 'True'
+        }
+        """
```
tomvothecoder, line 333: Here are docstring examples for calculating departures.

tomvothecoder:
Hi @pochedls, @lee1043, and @chengzhuzhang, this PR is ready for review. It takes precedence over PR #145 for the v0.2.0 release.

I added PR comments pointing to code examples in the docstrings that you can try if you checkout the branch. Let me know if you run into any issues, and I'm always open to suggestions. Thanks!

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,950 @@
...
+        >>> ds_year = ds.temporal.temporal_avg(
+        >>>     "timeseries",
+        >>>     "year",
+        >>>     data_var="ts"
+        >>> )
+        >>> ds_year.ts
+
+        Get weighted temporal average for a variable ("season" frequency):
+
+        >>> ds_season_with_djf = ds_season.temporal.temporal_avg(
```
pochedls, line 225: I'm confused by the dataset being passed in. Should this be: `ds_season_with_djf = ds.temporal.temporal_avg(`? Or should I have done some work on the original raw dataset `ds`?

pochedls: Another issue with seasonal frequency (but not month/annual) is that I am getting the following error:
```
fn = '/p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2/historical/r1i1p1f1/Amon/tas/gn/v20190308/tas_Amon_CESM2_historical_r1i1p1f1_gn_185001-201412.nc'
ds = xcdat.open_dataset(fn)
ds = ds.xcdat.temporal_avg("timeseries", "season")
```

> TypeError: '_ElementwiseFunctionArray' object does not support item assignment

tomvothecoder: > I'm confused by the dataset being passed in. Should this be: `ds_season_with_djf = ds.temporal.temporal_avg(`? Or should I have done some work on the original raw dataset `ds`?

That is a typo, it should be `ds_season_with_djf = ds.temporal.temporal_avg(...`.


> Another issue with seasonal frequency (but not month/annual) is that I am getting the following error:
> 
> ```
> fn = '/p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2/historical/r1i1p1f1/Amon/tas/gn/v20190308/tas_Amon_CESM2_historical_r1i1p1f1_gn_185001-201412.nc'
> ds = xcdat.open_dataset(fn)
> ds = ds.xcdat.temporal_avg("timeseries", "season")
> ```
> 
> > TypeError: '_ElementwiseFunctionArray' object does not support item assignment

I opened up an issue in the xarray repo (https://github.com/pydata/xarray/issues/6015) and thought I implemented a workaround for this using `.load()` on the time bounds. I'll take a look, thanks for the example Steve.

tomvothecoder: You may be on an outdated version of the branch. I ran the same example on the latest commit and it works.

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,986 @@
...
+        # The weights for time coordinates calculated based on a frequency.
+        self._weights: Optional[xr.DataArray] = None
+
+    def temporal_avg(
+        self,
+        mode: Mode,
+        freq: Frequency,
+        data_var: Optional[str] = None,
+        weighted: bool = True,
+        centered_time: bool = True,
```
pochedls, line 109: Consider renaming this `center_time`. `centered_time` (past tense) suggests the time has already been centered, but a value of `True` will result in `center_time()` being called.

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,986 @@
...
+        ds = self._dataset.copy()
+        if self._centered_time:
+            ds = self.center_time(ds)
+
+        # Perform data variable level operations.
+        dv = get_data_var(ds, data_var)
+        ds[dv.name] = self._averager(dv)
+
+        # Keep the original data var for calculating departures.
+        if mode == "climatology":
```
pochedls, line 286: I'm not sure what the best solution is here. I think it may be better to leave the variable (e.g., `tas`) as-is and add a new variable (e.g., `tas_climatology`). This seems more intuitive to me, but it would be useful to hear others' thoughts. I'm not far enough in to see what the logic is for departures, but this might help streamline the logic in that function (you'd likely need to add an argument specifying the departure type, monthly or seasonal): 

```
climo_vname = ds.variable_id + '_climatology'
# calculate the climatology if it doesn't exist
if climo_vname not in list(ds.keys()):
    self = self._temporal_avg("climatology", "month")
```

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,986 @@
...
+            ds[f"{dv.name}_original"] = dv.copy()
+
+        return ds
+
+    def departure(self, data_var: Optional[str] = None) -> xr.Dataset:
+        """Calculates departures (anomalies) for a data variable's climatology.
+
+        In climatology, “anomalies” refer to the difference between observations
```
pochedls, line 294: Suggest

> In climatology, “anomalies” refer to the difference between the value during a given time interval (e.g., the January average surface air temperature) and the long-term average value for that time interval (e.g., the average surface temperature over the last 30 Januaries).

xcdat/temporal_avg.py
tomvothecoder: Thanks, I must have copied the suggestion from one our meetings without thinking about the tense.

xcdat/temporal_avg.py
tomvothecoder: Hey Steve, I'll bring this up in the XCDAT meeting to gather input.

> I'm not sure what the best solution is here. I think it may be better to leave the variable (e.g., tas) as-is and add a new variable (e.g., tas_climatology). This seems more intuitive to me, but it would be useful to hear others' thoughts. 

My original thought process from Slack:
- `tas` - the variable that is actively being worked on (climatology, departures, spatial avg, etc.). We must maintain the “original” name ts so that xcdat can properly infer operations using the xcdat_infer attr.
- `tas_original` - the actual original, unmodified variable. Used for calculating departures, which gets stored in ts.

I was thinking if we wanted to support xcdat inference of data variables using the `xcdat_infer` attribute, we need to work on `tas`. One example is if we ran spatial averaging on `tas`, which does not add a new data variable like `tas_spatial_avg`. 
```python
ds = xcdat.open_dataset("tas")

ds.attrs["xcdat_infer"]  # "tas"

# "tas" is inferred from "xcdat_infer"
ds_spatial = ds.xcdat.spatial_avg(...)

# "tas" is inferred from "xcdat_infer"
ds_climo = ds_spatial.xcdat.temporal_avg(...)
```

Despite this, I am now having second thoughts about the `xcdat_infer` system altogether. It seems to be adding feature bloat and an unnecessary overhead when working with the API. It is more straightforward to have the user explicitly pass the key of the data variable, even if it might be redundant for datasets with only one data var. The behavior would at least remain consistent, regardless if the dataset was opened via xarray or XCDAT, or if a single or multiple data vars exist.

```python
ds = xcdat.open_dataset("tas")

# "tas_spatial_avg" gets added, "tas" remains untouched
ds_spatial = ds.xcdat.spatial_avg(data_var="tas")

# "tas_climo" gets added, "tas" and "tas_spatial_avg" remains untouched
ds_climo = ds_spatial.xcdat.temporal_avg(data_var="tas_spatial")
```

> this might help streamline the logic in that function so that you could call departure directly without calculating the climatology first

This answers a thought I had about whether we needed to call climatology first, or if departure should work without calling climatology (I noticed CDAT supports this).

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,988 @@
...
+        end_year = data_var.time.dt.year.data[-1]
+
+        incomplete_seasons = (f"{start_year}-01", f"{start_year}-02", f"{end_year}-12")
+        for year_month in incomplete_seasons:
+            try:
+                data_var.loc[dict(time=year_month)] = np.nan
+                data_var = data_var.dropna("time")
+
+                self._time_bounds.loc[dict(time=year_month)] = np.nan
+                self._time_bounds = self._time_bounds.dropna("time")
```
pochedls, line 649: This probably doesn't affect performance, but I think these `.dropna("time")` calls could be placed outside the loop. Would something like this might make sense:

```
        try:
            for year_month in incomplete_seasons:
                data_var.loc[dict(time=year_month)] = np.nan
                self._time_bounds.loc[dict(time=year_month)] = np.nan
            data_var = data_var.dropna("time")
            self._time_bounds = self._time_bounds.dropna("time")
        except KeyError:
            logger.debug("Dataset does not contain {month}.")
            continue
```

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,988 @@
...
+        Returns
+        -------
+        pd.DataFrame
+            The DataFrame of time coordinates for the "season" frequency with
+            obsolete columns dropped.
+        """
+        if self._mode == "timeseries":
+            df_season = df_season.drop("month", axis=1)
+        elif self._mode in ["climatology", "departure"]:
+            df_season = df_season.drop(["year", "month"], axis=1)
```
pochedls, line 850: This is a note to myself to come back to this...it is not clear to me right now why you would drop the year if the mode is departure.

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,988 @@
...
+        Parameters
+        ----------
+        data_var : xr.DataArray
+            The data variable with some incomplete DJF seasons.
+
+        Returns
+        -------
+        xr.DataArray
+            The data variable with all complete DJF seasons.
+        """
```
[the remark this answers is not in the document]
lee1043: Great point! I second this. We may not need to provide QC for user's time series but having this noted would be helpful.

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,988 @@
...
+        weights: xr.DataArray = time_grouped / time_grouped.sum()  # type: ignore
+
+        self._validate_weights(data_var, weights)
+        return weights
+
+    def _validate_weights(self, data_var: xr.DataArray, weights: xr.DataArray):
+        """Validates the sums of the weights for each group equals 1.0 or 0.0.
+
+        A sum of 0.0 means that all values in the group are np.nan (e.g.,
+        dropping incomplete "DJF" seasons).
```
pochedls, line 909: I wasn't sure if this case can occur, since DJF is dropped from the DataArray.

xcdat/temporal_avg.py
```diff
@@ -0,0 +1,988 @@
...
+        self._time_bounds: xr.DataArray = ds.time_bnds.copy()
+
+        # Get departures by calculating the difference between observations and
+        # the climatology.
+        departures = ds.copy()
+        dv_og = ds[f"{dv_climo.name}_original"].copy().rename("ts")
+        dv_obs = self._group_observations(dv_og)
+
+        with xr.set_options(keep_attrs=True):
+            dv_departures = dv_obs - dv_climo
```
pochedls, line 380: I think something is going awry in how I am attempting to calculate seasonal departures:

```
fn = '/p/user_pub/work/E3SM/1_0/historical/1deg_atm_60-30km_ocean/atmos/180x360/time-series/mon/ens1/v5/TS_185001_201412.nc'
mode = 'timeseries'
freq = 'season'
weighted = True
centered_time = True
# custom_seasons = [["Jan", "Feb", "Mar"], ["Apr", "May", "Jun"], ["Jul", "Aug", "Sep"], ["Oct", "Nov", "Dec"]]
season_config = {"dec_mode": "DJF", "drop_incomplete_djf": True,
                 "custom_seasons": None}
ds = xcdat.open_dataset(fn, data_var='TS')
ds = ds.xcdat.spatial_avg()
ds_climo = ds.temporal.temporal_avg("climatology", "season")
ds_departure = ds_climo.temporal.departure()
ds.close()
```

The output is not what I expect. I expect it to have roughly size `len(ds.TS)/4`, but I am getting the full time series length (1980 months). 

One other thing that is no problem, but I didn't realize at first is that the seasonal output starts with MAM. I expected it to start with DJF. I think this is fine, but noting it in case it is a surprise to others.

pochedls: Okay – this appears to remove the seasonal anomaly from each month, but an additional step is required to get the seasonal anomaly:

```
ds_departure_seasonal = ds_departure.temporal.temporal_avg("timeseries", "season")
```

xcdat/temporal_avg.py
pochedls: Got it. This is only called from the `_process_season_dataframe` function.

chengzhuzhang:
It seems `center_time` is not working from my end:
```
fn = '/p/user_pub/work/E3SM/1_0/historical/1deg_atm_60-30km_ocean/atmos/180x360/time-series/mon/ens1/v5/TS_185001_201412.nc'
ds = xcdat.open_dataset(fn, data_var='TS')
ds.xcdat.center_time()
ds.TS.time[0].values #this gives array(cftime.DatetimeNoLeap(1850, 2, 1, 0, 0, 0, 0, has_year_zero=True),
      dtype=object)
```

I think the expected behavior with center_time is to yield (1850,1,15/16) as the first time value? It is possible that I'm not use it correctly...
Following is the time bounds
[cftime.DatetimeNoLeap(1850, 1, 1, 0, 0, 0, 0, has_year_zero=True),
        cftime.DatetimeNoLeap(1850, 2, 1, 0, 0, 0, 0, has_year_zero=True)]

xcdat/temporal_avg.py
chengzhuzhang: Seasonal output starting with DJF perhaps is more intuitive. I would propose to change the order ..

tomvothecoder:
> fn = '/p/user_pub/work/E3SM/1_0/historical/1deg_atm_60-30km_ocean/atmos/180x360/time-series/mon/ens1
> #or fn = '/p/user_pub/e3sm/zhang40/xcdat_test_e3sm'
> /v5/TS_185001_201412.nc'
> ds = xcdat.open_dataset(fn, data_var='TS')
> ds.xcdat.center_time()
> ds.TS.time[0].values #this gives array(cftime.DatetimeNoLeap(1850, 2, 1, 0, 0, 0, 0, has_year_zero=True),
>       dtype=object)

Hey @chengzhuzhang, `ds.xcdat.center_time()` returns a dataset, so you need to update the variable `ds`.
```python
import xcdat

fn = '/p/user_pub/work/E3SM/1_0/historical/1deg_atm_60-30km_ocean/atmos/180x360/time-series/mon/ens1/v5/TS_185001_201412.nc'
ds = xcdat.open_dataset(fn, data_var='TS')
ds = ds.xcdat.center_time()

ds.TS.time[0].values
# array(cftime.DatetimeNoLeap(1850, 1, 16, 12, 0, 0, 0, has_year_zero=True), dtype=object)
```

xcdat/temporal_avg.py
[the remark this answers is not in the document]
tomvothecoder: That's exactly right, I will add that in the notes. Thanks Steve!

xcdat/temporal_avg.py
tomvothecoder: Nice optimization, better to perform the operation once if possible.

tomvothecoder: The call to `.dropna("time")` assumes a continuous time series. Might need to address this by dropping specific coordinate points instead.

chengzhuzhang:
> 
> Hey @chengzhuzhang, `ds.xcdat.center_time()` returns a dataset, so you need to update the variable `ds`.
> 
Thank you, Tom. `center_time` is very useful. I'm wondering is it straightforward to make it as an argument in open_dataset().

xcdat/temporal_avg.py
tomvothecoder: The logic for this method has been updated to only drop the leading and trailing incomplete seasons.
https://github.com/XCDAT/xcdat/blob/b3e975f6d803a8314d4bd95ef5681c9c43794dec/xcdat/temporal_avg.py#L644-L661

xcdat/temporal_avg.py
tomvothecoder: Thanks for catching this incorrect description. DJF seasons should be dropped prior to calculating weights, rather than masked with `np.nan`.

I incorrectly wrote two DJF tests for `_calculate_weights()` that had `np.nans` representing the dropped seasons (they should be removed instead of masked).

xcdat/temporal_avg.py
tomvothecoder: > Okay – this appears to remove the seasonal anomaly from each month, but an additional step is required to get the seasonal anomaly:
> 
> ```
> ds_departure_seasonal = ds_departure.temporal.temporal_avg("timeseries", "season")
> ```

As mentioned on Slack, this can be addressed by adding this call to the end of the `departure()` method with the same `freq`.

> One other thing that is no problem, but I didn't realize at first is that the seasonal output starts with MAM. I expected it to start with DJF. I think this is fine, but noting it in case it is a surprise to others.

> Seasonal output starting with DJF perhaps is more intuitive. I would propose to change the order .. Actually, I think it is fine to start with MAM. NVM...

The seasonal output starts at MAM because when the leading and trailing incomplete DJF seasons are removed. The month for the first timestamp is March, so the grouping starts at MAM.

A one-liner call to xarray's sort method can be implemented.

tomvothecoder:
> > Hey @chengzhuzhang, `ds.xcdat.center_time()` returns a dataset, so you need to update the variable `ds`.
> 
> Thank you, Tom. `center_time` is very useful. I'm wondering is it straightforward to make it as an argument in open_dataset().

No problem, glad to hear! Yes, adding  `center_time` as a kwarg to `open_dataset()` is straightforward so I'll do that.

tests/test_temporal_avg.py
```diff
@@ -0,0 +1,3207 @@
...
+                "operation": "temporal_avg",
+                "mode": "departures",
+                "freq": "season",
+                "weighted": "True",
+                "center_times": "False",
+                "groupby": "season",
+                "dec_mode": "DJF",
+                "drop_incomplete_djf": "True",
+            },
+        )
```
tomvothecoder, line 1021:
```suggestion
        expected["ts"] = xr.DataArray(
            data=np.zeros((12, 4, 4)),
            coords={
                "lat": ds.lat,
                "lon": ds.lon,
                "time": ds.time,
            },
            dims=["time", "lat", "lon"],
            attrs={
                "operation": "temporal_avg",
                "mode": "departures",
                "freq": "season",
                "weighted": "True",
                "center_times": "False",
                "groupby": "season",
                "dec_mode": "DJF",
                "drop_incomplete_djf": "True",
            },
        )
```

tests/test_temporal_avg.py
```diff
@@ -0,0 +1,3207 @@
...
+        # Update time bounds with centered time coordinates.
+        time_bounds = ds.time_bnds.copy()
+        time_bounds["time"] = expected.time
+        expected["time_bnds"] = time_bounds
+
+        # Compare result of the method against the expected.
+        result = ds.temporal.center_times(ds)
+        assert result.identical(expected)
+
+
```
tomvothecoder, line 1489:
```suggestion


# TODO: Update TestAverager tests to use other numbers rather than 1's for
# better test reliability and accuracy. This may require subsetting.
```

tests/test_temporal_avg.py
```diff
@@ -0,0 +1,3207 @@
...
+        # Drop the incomplete DJF seasons
+        expected = expected.isel(time=slice(2, -1))
+        expected["ts"] = xr.DataArray(
+            data=np.ones(2),
+            coords={"time": [datetime(2000, 3, 1), datetime(2000, 4, 1)]},
+            dims=["time"],
+        )
+        assert result.identical(expected)
+
+    def test_does_not_drop_incomplete_seasons_dont_exist(self):
```
tomvothecoder, line 1672:
```suggestion
    def test_does_not_drop_incomplete_seasons_if_none_exist(self):
```

tomvothecoder (commented):
Some minor review updates

durack1:
@tomvothecoder this looks great, I wonder are all the various flavours of calendar managed by XCDAT?

durack1:
Ok this is great, it seems all the `cdtime` calendars are included in `cftime`, so we're good, great!

So the `cftime` library includes (https://unidata.github.io/cftime/api.html#cftime.num2date)
```python
calendar: describes the calendar used in the time calculations. All the values currently
defined in the [CF metadata convention](http://cfconventions.org/cf-conventions/cf-conventions#calendar)
are supported. Valid calendars ‘standard’, ‘gregorian’, ‘proleptic_gregorian’ ‘noleap’,
‘365_day’, ‘360_day’, ‘julian’, ‘all_leap’, ‘366_day’. Default is ‘standard’, which is a
mixed Julian/Gregorian calendar.
```

Just as an FYI, dropping some `cdtime` links here for documentation:
https://github.com/CDAT/cdtime/blob/master/Src/cdtimemodule.c#L1422-L1430 <- `cdtime` calendar types,

```python
    PyDict_SetItemString(d, "error", PyCdtime_ErrorObject); /* export exception */
    DefineLongConstant(d, "StandardCalendar", (long)cdStandard);
    DefineLongConstant(d, "GregorianCalendar", (long)cdStandard);
    DefineLongConstant(d, "JulianCalendar", (long)cdJulian);
    DefineLongConstant(d, "MixedCalendar", (long)cdMixed);
    DefineLongConstant(d, "NoLeapCalendar", (long)cdNoLeap);
    DefineLongConstant(d, "Calendar360", (long)cd360);
    DefineLongConstant(d, "ClimCalendar", (long)cdClim);
    DefineLongConstant(d, "ClimLeapCalendar", (long)cdClimLeap);
    DefineLongConstant(d, "DefaultCalendar", (long)cdMixed); /* initialize the default calendar */
```
Which calls
https://github.com/CDAT/libcdms/blob/master/include/cdms.h#L149-L167
```python
#define cdStandardCal   0x11
#define cdClimCal        0x0
#define cdHasLeap      0x100
#define cdHasNoLeap    0x000
#define cd365Days     0x1000
#define cd360Days     0x0000
#define cdJulianCal  0x10000
#define cdMixedCal   0x20000

typedef enum cdCalenType {
	cdStandard    = ( cdStandardCal | cdHasLeap   | cd365Days),
	cdJulian      = ( cdStandardCal | cdHasLeap   | cd365Days | cdJulianCal),
	cdNoLeap      = ( cdStandardCal | cdHasNoLeap | cd365Days),
	cd360         = ( cdStandardCal | cdHasNoLeap | cd360Days),
	cdClim        = ( cdClimCal     | cdHasNoLeap | cd365Days),
	cdClimLeap    = ( cdClimCal     | cdHasLeap   | cd365Days),
	cdClim360     = ( cdClimCal     | cdHasNoLeap | cd360Days),
	cdMixed       = ( cdStandardCal | cdHasLeap   | cd365Days | cdMixedCal)
}  cdCalenType;
```

tomvothecoder:
Hey everyone, this PR is ready for a final round of review. Here are the associated validation notebooks that I mentioned on our email chain: https://github.com/XCDAT/xcdat_test/pull/28

@lee1043, @chengzhuzhang, and @pochedls -- feel free to try out the API and look through the notebooks
@jasonb5 -- it will be nice if you can review the code for any optimizations, improvements, and fixes

I'm hoping to have this PR ready to be merged by our meeting next Wednesday. Thanks!

durack1:
Folks I just came across this, might be worth a quick look to see alternatives https://ncar.github.io/esds/posts/2021/yearly-averages-xarray/

@chengzhuzhang @tomvothecoder @pochedls

jasonb5 (changes requested):
Rest of the code LGTM.

tomvothecoder:
> Hi @tomvothecoder ,
> 
> My testing goes smoothly so far. Only have two high level comments:
> 
>     * It would be really handy, if this feature can include the capability to calculate weighted temporal average. Right now in `xarray`, one still has to generates the `weights` and then do something like `(d * weights).groupby("time.month").sum(dim="time")` I think `temporal_average` can cover this case straightforwardly? perhaps we can add a `mean` mode in addition or enable this if neither "climatology" nor "timeseries" is supplied?
> 
>     * And have a time coordinate, as well as a multi-index coordinate is a bit confusing. The additional step to convert the multi-index coordinates back to a new time coordinate seems a good way to clear this confusing. But it will require some additional work...

I decided to carry these issues over to #201 and #202.

chengzhuzhang:
Sounds good!

PR #152, opened 2021-11-12
Refactor spatial averaging with more robust handling of longitude spanning prime meridian

### LINKED ISSUE #150: Refactor Spatial Averaging: More Robust Handling of Longitude Spanning Prime Meridian
### What versions of software are you using?

-   Package Version: CDAT v0.1.0

### What are the steps to reproduce this issue?
```
fn = '/p/css03/esgf_publish/CMIP6/CMIP/CSIRO/ACCESS-ESM1-5/historical/r10i1p1f1/Amon/tas/gn/v20200605/tas_Amon_ACCESS-ESM1-5_historical_r10i1p1f1_gn_185001-201412.nc'
ds = xcdat.open_dataset(fn)
tsa = ds.xcdat.spatial_avg(lat_bounds=(-30, 30), lon_bounds=(0, 360)).tas.load()
print(tsa[0].values)
```
> 297.44541842495704

But the correct value is: 297.150202135895 (Difference of 0.31 K)

I think this can be traced to [this line](https://github.com/XCDAT/xcdat/blob/main/xcdat/spatial_avg.py#L343): `domain_bounds` should be `dom_bounds` in this case.

This will fix the issue in this case, but I realize there are specific situations that will give an (slightly) incorrect solution when the specified regional bounds are within a grid cell that spans across the prime meridian (e.g., -1 to 1). 

### Any other comments?

This code should be refactored to more robustly deal with averaging along the longitude axis (it needs to robustly handle domain bounds and regional selections that span the prime meridian). An approach I would like to try is to:

1. Convert longitude axes to 0 to 360 domain
2. If a grid cell boundary crosses the prime meridian, concatenate a version of that grid cell spanning 0 and 360 (e.g., [-1, 1] and [359, 361])
3. Limit the bounds of each to within 0 to 360 (e.g., [0, 1] and [359, 360]) so that the domain is the same, but now represented by two grid cells
4. Calculate weights based on this new dimension bounds array
5. Add the weight from each of the repeated grid cells together (so that the weight array is the length of the longitude dimension)

pochedls:
## Description

- Closes #150 
- Refactors spatial averaging functionality to more robustly handle domains that span the prime meridian (the current code returns a (slightly) incorrect result in some specific situations where a grid cell crosses a prime meridian). 

## Checklist

If applicable:


xcdat/spatial_avg.py
```diff
@@ -203,6 +203,27 @@ def _validate_axis(
 
         return axis
 
+    def _validate_domain_bounds(self, domain_bounds: xr.DataArray):
```
pochedls, line 206: In refactoring this code, I realize I was making the assumption that domain bounds were ordered from low to high, (e.g., `[[0, 1], [1, 2], ..., [359, 360]]`) and not the opposite or mixed (e.g., `[[1, 0], [2, 1], ..., [360, 358]]`). I *think* this assumption is also consistent with [cf-conventions](https://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries):

> For a coordinate variable such as lat(lat) with associated boundary variable `latbnd(x,2)`, the interval endpoints must be ordered consistently with the associated coordinate, e.g., for an increasing coordinate, `lat(1) > lat(0)` implies `latbnd(i,1) >= latbnd(i,0)` for all `i` (figure 1).

xcdat/spatial_avg.py
```diff
@@ -203,6 +203,27 @@ def _validate_axis(
...
+
+        Parameters
+        ----------
+        domain_bounds: xr.DataArray
+            The bounds of an axis
+
+        Raises
+        ------
+        TypeError
+            If the ``bounds`` of a grid cell are not ordered low-to-high.
```
pochedls, line 217: I think ``bounds`` should be ``domain_bounds``?

xcdat/spatial_avg.py
```diff
@@ -324,83 +345,136 @@ def _get_weights(
...
+                        # grid cells ([0, 1] and [359, 360]).
+                        pmb_index, dom_bounds = self._align_longitude_to_360_axis(
+                            dom_bounds
                         )
+
+                        # If the region bounds are equal, it means all
+                        # longitude values are specified (e.g., (360, 360) ==
+                        # (0, 360)). As a result, set the region lon bounds to
+                        # [0, 360].
+                        if reg_bounds[1] - reg_bounds[0] == 0:
```
pochedls, line 372: This is replacing `_set_equal_lon_bounds_to_domain` since it is used once and is only a couple lines. But this causes the code coverage to go down - I'm not exactly sure how to test that this specific section of code runs as needed.

xcdat/spatial_avg.py
```diff
@@ -324,83 +345,136 @@ def _get_weights(
...
+
                 if region_bounds is not None:
                     reg_bounds = np.array(region_bounds).astype(float)
                     if dim == "lon":
-                        # Forces longitude bounds linearity for elements where
-                        # the upper value is greater than the lower value (e.g.,
-                        # across prime meridian or dateline) to avoid issues
-                        # in operations/calculations that operate using the lon
-                        # bounds.
-                        dom_bounds, reg_bounds = self._force_lon_linearity(
```
pochedls, line 335: This use to try to make sure that the `dom_bounds` and `reg_bounds` lined up (but in specific cases did not work). This refactoring ensures that both `dom_bounds` and `reg_bounds` are in a 0 -> 360 coordinate system (the new relevant function is `_align_longitude_to_360_axis`).

xcdat/spatial_avg.py
```diff
@@ -324,83 +345,136 @@ def _get_weights(
...
                 dim_wts.name = dim + "_wts"
                 axis_weights[dim] = dim_wts
 
         weights = self._combine_weights(axis, axis_weights)
         return weights
 
-    def _force_lon_linearity(
-        self, domain_bounds: xr.DataArray, region_bounds: np.ndarray
-    ) -> Tuple[xr.DataArray, np.ndarray]:
+    def _align_longitude_to_360_axis(self, domain_bounds: xr.DataArray):
```
pochedls, line 406: Replace `_force_lon_linearity` with `_align_longitude_to_360_axis`. It essentially ensures that the domain bounds neatly span from 0 to 360 degrees. If there is a grid cell that crosses the prime meridian (e.g., [-1, 1]) it breaks this into to different grid cells spanning [0, 1] and [359, 360]. In this case, the original grid cell index is returned with the function. The domain_bounds shape goes from [nlon, 2] to [nlon+1, 2]. When these weights are calculated (e.g., 1 - 0 = 1 and 360 - 359 = 1) they are summed and both weights are included with the grid cell that was indexed (so that the weight vector is length nlon).

xcdat/spatial_avg.py
pochedls: @tomvothecoder - do you know how I can test this? This runs in `get_weights`, which returns weights (not `reg_bounds`) - so I can't directly see if the function is appropriately adjusting the domain longitude bounds here.

tomvothecoder: I usually add VSCode breakpoints to the lines I want to debug to check variable values at runtime. This involves a debug script with a call to get_weights(), then clicking "Debug Line" for that call.

If your editor doesn't support this, you can install ipdb in your conda environment and import it within this file. Then you can add ipdb breakpoints to pause at that line, step forward, step into , etc.

pochedls: I am pretty confident this is working as I want it to (I could debug to verify)...but the issue is that the codecov report indicated that this isn't covered (I assume it wants a unit test that covers this portion of code?).

pochedls: This is the [codecov report](https://app.codecov.io/gh/XCDAT/xcdat/compare/152/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=XCDAT#D1L373).

xcdat/spatial_avg.py
tomvothecoder: Sounds good since it follows CF conventions.

If the bounds aren't following this convention, should the user fix the bounds before spatial averaging?
We can also add a conditional to check that all min bound values are less than all max bounds values and raise ValueError if any of them aren't.

xcdat/spatial_avg.py
tomvothecoder: Correct

xcdat/spatial_avg.py
pochedls: > If the bounds aren't following this convention, should the user fix the bounds before spatial averaging?

I think they will need to, because they will get a ValueError if their bounds are not ordered correctly when calling spatial averaging. 

> We can also add a conditional to check that all min bound values are less than all max bounds values and raise ValueError if any of them aren't.

I think this is what is done [right here](https://github.com/XCDAT/xcdat/blob/c56569440c1d79c8b578008dc087f949e9d37582/xcdat/spatial_avg.py#L220).

tomvothecoder: > I think this is what is done right here.

That check is performed on the user specified tuple representing the region bounds, and not the domain bounds found in the Dataset. 

We can add a check here when we retrieve the domain bounds. https://github.com/XCDAT/xcdat/blob/c56569440c1d79c8b578008dc087f949e9d37582/xcdat/spatial_avg.py#L329-L338

However, we can also avoid hand-holding and let the user take responsibility for making sure bounds are CF compliant.

tomvothecoder: Ignore all comments I have just made, this makes sense now

tomvothecoder:
@pochedls I noticed some of my commits from #154 were merged directly to this branch before #154 was merged. Not a big deal, but there will probably some merge conflicts to handle.

tomvothecoder:
PR #154 has been merged into this branch so I'll do a final review of this PR.

I noticed you opened up #163. I think we should open up a separate branch and PR for that one.

tests/test_spatial_avg.py
```diff
@@ -454,124 +454,252 @@ def test_returns_area_weights_for_region_within_lon(self):
...
+            dims=["lon"],
+        )
+        result = self.ds.spatial._get_longitude_weights(
+            domain_bounds=self.ds.lon_bnds.copy(),
+            region_bounds=np.array([0.0, 360.0]),
+        )
+
+        xr.testing.assert_allclose(result, expected)
+
+    def test_weights_all_longitudes_for_explicit_equal_region_bounds(self):
```
tomvothecoder, line 543:
```suggestion
    def test_weights_for_equal_region_bounds_representing_entire_lon_domain(self):
```

tests/test_spatial_avg.py
```diff
@@ -454,124 +454,252 @@ def test_returns_area_weights_for_region_within_lon(self):
...
     @pytest.fixture(autouse=True)
     def setup(self):
         self.ds = generate_dataset(cf_compliant=True, has_bounds=True)
 
-    def test_forces_linearity_for_rows_in_domain_bounds_where_lower_val_is_larger_than_upper(
-        self,
-    ):
-        domain_bounds = xr.DataArray(
-            name="lon_bnds",
+    def test_returns_weights_for_region_in_lon(self):
```
tomvothecoder, line 462:
```suggestion
    def test_returns_weights_for_region_in_lon_domain(self):
```

tests/test_spatial_avg.py
```diff
@@ -454,124 +454,252 @@ def test_returns_area_weights_for_region_within_lon(self):
...
+
+        xr.testing.assert_allclose(result, expected)
+
+
+class TestGetLatitudeWeights:
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        self.ds = generate_dataset(cf_compliant=True, has_bounds=True)
+
+    def test_weights_for_region_within_lat(self):
```
tomvothecoder, line 563:
```suggestion
    def test_weights_for_region_in_lat_domain(self):
```

tests/test_spatial_avg.py
```diff
@@ -743,17 +871,16 @@ def setup(self):
         }
 
     def test_weights_for_single_axes_is_the_same(self):
```
tomvothecoder, line 873:
```suggestion
    def test_weights_for_single_axis_is_identical(self):
```

tests/test_spatial_avg.py
```diff
@@ -743,17 +871,16 @@ def setup(self):
...
-        )
+        axis_weights = self.axis_weights
+        del axis_weights["lon"]
+
+        result = self.ds.spatial._combine_weights(axis_weights=self.axis_weights)
         expected = self.axis_weights["lat"]
 
         assert result.identical(expected)
 
     def test_weights_for_multiple_axes_is_a_matrix_multiplication(self):
```
tomvothecoder, line 882:
```suggestion
    def test_weights_for_multiple_axis_is_a_matrix_multiplication(self):
```

tests/test_spatial_avg.py
```diff
@@ -396,7 +396,7 @@ class TestGetWeights:
     def setup(self):
         self.ds = generate_dataset(cf_compliant=True, has_bounds=True)
 
-    def test_returns_area_weights_for_region_within_lat_and_lon(self):
+    def test_area_weights_for_region_within_lat_and_lon(self):
```
tomvothecoder, line 399:
```suggestion
    def test_weights_for_region_within_lat_and_lon_domains(self):
```

tests/test_spatial_avg.py
```diff
@@ -415,7 +415,7 @@ def test_returns_area_weights_for_region_within_lat_and_lon(self):
 
         xr.testing.assert_allclose(result, expected)
 
-    def test_returns_area_weights_for_region_within_lat(self):
+    def test_area_weights_for_region_within_lat(self):
```
tomvothecoder, line 418:
```suggestion
    def test_area_weights_for_region_in_lat_domain(self):
```

tests/test_spatial_avg.py
```diff
@@ -434,7 +434,7 @@ def test_returns_area_weights_for_region_within_lat(self):
 
         xr.testing.assert_allclose(result, expected)
 
-    def test_returns_area_weights_for_region_within_lon(self):
+    def test_area_weights_for_region_within_lon(self):
```
tomvothecoder, line 437:
```suggestion
    def test_weights_for_region_in_lon_domain(self):
```

xcdat/spatial_avg.py
```diff
@@ -194,15 +195,36 @@ def _validate_axis(
...
+        TypeError
+            If the ``domain_bounds`` of a grid cell are not ordered low-to-high.
+        """
+        index_bad_cells = np.where(domain_bounds[:, 1] - domain_bounds[:, 0] < 0)[0]
+        if len(index_bad_cells) > 0:
+            raise ValueError(
+                "The bounds have unexpected ordering: "
+                "A lower bound has a greater value than an "
+                "upper bound"
+            )
```
tomvothecoder, line 226:
```suggestion
        if len(index_bad_cells) > 0:
            raise ValueError(
                "The bounds have unexpected ordering. A lower bound has a "
                "greater value than an upper bound."
            )
```

xcdat/spatial_avg.py
```diff
@@ -303,109 +325,228 @@ def _get_weights(
...
+        return weights
+
+    def _get_latitude_weights(
+        self, domain_bounds: xr.DataArray, region_bounds: xr.DataArray
+    ) -> xr.DataArray:
+        """Gets weights for the latitude axis.
+
+        This method scales the domain to a region (if selected). It also scales
+        the area between two lines of latitude scales as the difference of the
+        sine of latitude bounds.
```
tomvothecoder, line 429:
```suggestion
        This method scales the domain to a region (if selected). It also scales
        the area between two lines of latitude as the difference of the
        sine of latitude bounds.
```

xcdat/spatial_avg.py
```diff
@@ -303,109 +325,228 @@ def _get_weights(
...
+                "Longitude bounds aren't inclusively between 0 and 360. "
+                "Use `_swap_lon_axis()` before calling `_align_longitude_to_360_axis()`."
+            )
+
+        p_meridian_index = np.where(domain_bounds[:, 1] - domain_bounds[:, 0] < 0)[0]
+        if len(p_meridian_index) == 0:
+            p_meridian_index = None
+        elif len(p_meridian_index) > 1:
+            raise ValueError("More than one grid cell spans prime meridian.")
+        elif len(p_meridian_index) == 1:
```
tomvothecoder, line 521:
```suggestion
        if p_meridian_index.size == 0:
            p_meridian_index = None
        elif p_meridian_index.size > 1:
            raise ValueError("More than one grid cell spans prime meridian.")
        elif p_meridian_index.size == 1:
```

tomvothecoder (changes requested):
Just some minor suggestions that can be merged directly by accepting them. After that, it should be good to squash and merge.

PR #161, opened 2021-11-18
Update xcdat.open_mfdataset time decoding logic

### LINKED ISSUE #158: open_mfdataset not decoding times correctly
### What versions of software are you using?

-   Package Version: main branch (latest)

### What are the steps to reproduce this issue?
```
import xarray as xr
import xcdat

dpath = "/p/css03/cmip5_css02/data/cmip5/output1/CMCC/CMCC-CESM/historical/mon/atmos/Amon/r1i1p1/tas/1/*nc"

# xcdat
ds = xcdat.open_mfdataset(dpath, "tas")
print('xcdat inferred tas shape: ', ds.tas.shape)
ds.close()

# xarray
ds = xr.open_mfdataset(dpath)
print('xarray inferred tas shape: ', ds.tas.shape)
ds.close()

# xarray with decode_times=False
ds = xr.open_mfdataset(dpath, decode_times=False)
print('xarray (no time decoding) inferred tas shape: ', ds.tas.shape)
ds.close()
```
### What happens? Any logs, error output, etc?

The code excerpt above shows that xcdat yields a `tas` time series that has only 12 time steps, but xarray has 1872. We can reproduce the xcdat result by adding `decode_times=False` (which is included in the `xcdat.open_mfdataset()` logic [here](https://github.com/XCDAT/xcdat/blob/main/xcdat/dataset.py#L172).

> xcdat inferred tas shape:  (12, 48, 96)
> xarray inferred tas shape:  (1872, 48, 96)
> xarray (no time decoding) inferred tas shape:  (12, 48, 96)

### What were you expecting to happen?

`xcdat.open_mfdataset()` should be able to load and decode the entire time series (1872 time steps). 

### Any other comments?

I think (this is a guess) this may arise because the time units in each netCDF in this directory have different units (but perhaps the same time values). We may need to add a function that checks for this and reverts to `decode_times=False`. I imagine this could get complicated in some situations, but I think starting by turning this off in some circumstances would likely be the most ideal behavior.

pochedls:
## Description
- Refactor `xcdat.open_mfdataset()` to use `xarray`'s in-built time decoding if the dataset has cf_compliant time units
- Closes #158 

## Checklist

If applicable:

### DIFF
--- a/xcdat/dataset.py
+++ b/xcdat/dataset.py
@@ -220,11 +220,9 @@ def has_cf_compliant_time(
 ) -> Optional[bool]:
     """Determine if a dataset has time coordinates with CF compliant units.
 
-    This function opens a single dataset either from a single string path, glob
-    string path, a list, or a list of lists. If the dataset does not contain a
-    time dimension, None is returned. Otherwise, the units attribute is
-    extracted from the time coordinates to determine whether it is CF or
-    non-CF compliant.
+    If the dataset does not contain a time dimension, None is returned.
+    Otherwise, the units attribute is extracted from the time coordinates to
+    determine whether it is CF or non-CF compliant.
 
     Parameters
     ----------
--- a/xcdat/dataset.py
+++ b/xcdat/dataset.py
@@ -1,6 +1,7 @@
 """Dataset module for functions related to an xarray.Dataset."""
+from functools import partial
 from glob import glob
-from typing import Any, Dict, Hashable, List, Optional, Union
+from typing import Any, Callable, Dict, Hashable, List, Optional, Tuple, Union
 
 import pandas as pd
 import xarray as xr
@@ -11,23 +12,25 @@ from xcdat.logger import setup_custom_logger
 
 logger = setup_custom_logger(__name__)
 
+#: List of non-CF compliant time units.
+NON_CF_TIME_UNITS: List[str] = ["months", "years"]
+
 
 def open_dataset(
-    path: str, data_var: Optional[str] = None, **kwargs: Dict[str, Any]
+    path: str,
+    data_var: Optional[str] = None,
+    decode_times: bool = True,
+    **kwargs: Dict[str, Any],
 ) -> xr.Dataset:
     """Wrapper for ``xarray.open_dataset()`` that applies common operations.
 
     Operations include:
 
-    - Decode both CF and non-CF compliant time units if the Dataset has a time
-      dimension
+    - Optional decoding of time coordinates with CF or non-CF compliant units if
+      the Dataset has a time dimension
     - Add missing bounds for supported axis
     - Option to limit the Dataset to a single regular (non-bounds) data
       variable, while retaining any bounds data variables
-
-    ``decode_times`` is statically set to ``False``. This enables a check
-    for whether the units in the time dimension (if it exists) contains CF or
-    non-CF compliant units, which determines if manual decoding is necessary.
 
     Parameters
     ----------
@@ -35,6 +38,10 @@ def open_dataset(
         Path to Dataset.
     data_var: Optional[str], optional
         The key of the data variable to keep in the Dataset, by default None.
+    decode_times: bool
+        If True, decode times encoded in the standard NetCDF datetime format
+        into datetime objects. Otherwise, leave them encoded as numbers.
+        This keyword may not be supported by all the backends, by default True.
     kwargs : Dict[str, Any]
         Additional arguments passed on to ``xarray.open_dataset``. Refer to the
         [1]_ xarray docs for accepted keyword arguments.
@@ -67,18 +74,19 @@ def open_dataset(
 
     >>> from xcdat.dataset import open_dataset
     >>> ds = open_dataset("file_path", data_var="tas")
-
-    Keep multiple variables in the Dataset:
-
-    >>> from xcdat.dataset import open_dataset
-    >>> ds = open_dataset("file_path", data_var=["ts", "tas"])
     """
-    ds = xr.open_dataset(path, decode_times=False, **kwargs)
+    if decode_times:
+        cf_compliant_time: Optional[bool] = has_cf_compliant_time(path)
+        if cf_compliant_time is False:
+            # XCDAT handles decoding time values with non-CF units.
+            ds = xr.open_dataset(path, decode_times=False, **kwargs)
+            ds = decode_non_cf_time(ds)
+        else:
+            ds = xr.open_dataset(path, decode_times=True, **kwargs)
+    else:
+        ds = xr.open_dataset(path, decode_times=False, **kwargs)
+
     ds = infer_or_keep_var(ds, data_var)
-
-    if ds.cf.dims.get("T") is not None:
-        ds = decode_time_units(ds)
-
     ds = ds.bounds.add_missing_bounds()
     return ds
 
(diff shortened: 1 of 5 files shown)


tomvothecoder:
The check for CF compliance in `_check_dataset_for_cf_compliant_time()` is repeated in `decode_time_units()`.
https://github.com/XCDAT/xcdat/blob/b010930ecf96117c793ce7fc04d89f5074ba8271/xcdat/dataset.py#L329-L341
We should remove these checks in `decode_time_units()` and refactor `_check_dataset_for_cf_compliant_time()` so that it works for both a single dataset (through `xcdat.open_dataset()`) and multi-file datasets (through `xcdat.open_mfdataset()`).

Comments from #158 
> The code excerpt above shows that xcdat yields a tas time series that has only 12 time steps, but xarray has 1872. We can reproduce the xcdat result by adding decode_times=False (which is included in the xcdat.open_mfdataset() logic here.

I expected `xr.open_mfdataset(decode_times=False)` to still be able to join all non-decoded timestamps, but apparently it doesn't. Instead, it only keeps the last file's timestamp for some reason (probably below).

> "I think (this is a guess) this may arise because the time units in each netCDF in this directory have different units (but perhaps the same time values)."

I believe your explanation might be right, since each file has their own units ("days since ..."). This might be causing xarray to not join the timestamps across files.

---
Another important note is that the updates in this **PR only handles multi-file datasets with CF compliant time.** 
We still have to handle multi-file datasets that have non-CF compliant time, which force us to use `decode_times=False` and thus leading to a dropping of timestamps.

Example:
```
    # if cf_compliant, let xarray decode the time units
    # otherwise, decode using decode_time_units
    if cf_compliant:
        ds = xr.open_mfdataset(paths, decode_times=True, data_vars=data_vars, **kwargs)
    else:
        # FIXME: If non-CF compliant, using `decode_times=False` will still only
        # keep the last file's timestamps. There's an issue with how
        # `decode_times=False` works in `xr.open_mfdataset()`.
        ds = xr.open_mfdataset(paths, decode_times=False, data_vars=data_vars, **kwargs)
        if ds.cf.dims.get("T") is not None:
            ds = decode_time_units(ds)
```

tomvothecoder:
I'm running into a weird xarray issue in PR #107 involving datasets with CF compliant time units that are decoded using `xr.decode_cf()` and masking specific time bounds coordinates for incomplete DJF seasons. 

I opened up the issue on the xarray repo (https://github.com/pydata/xarray/issues/6015). The updates in this branch should allow us to avoid using `xr.decode_cf()` through `decode_time_units()`.

pochedls:
> Another important note is that the updates in this PR only handles multi-file datasets with CF compliant time.
We still have to handle multi-file datasets that have non-CF compliant time, which forces us to use decode_times=False and thus leading to a dropping of timestamps. I'm trying to see why xarray only keeps the last file's timestamps in certain cases (e.g., diff units).

Okay - I did not realize this - I see that this is not working for non-CF compliant multi-file datasets:

```
flist = ['/p/user_pub/climate_work/pochedley1/cmip6_msu/tmt_Amon_CanESM5_historical_r1i1p1f1_gr_185001-201412.nc',
         '/p/user_pub/climate_work/pochedley1/cmip6_msu/tmt_Amon_CanESM5_ssp370_r1i1p1f1_gr_201501-210012.nc']
ds = xcdat.open_mfdataset(flist)
print(ds.time)
```

> <xarray.DataArray 'time' (time: 3012)>                                                                                                          
> array(['1800-01-01T00:00:00.000000000', '1800-02-01T00:00:00.000000000',
>        '1800-03-01T00:00:00.000000000', ..., '2050-10-01T00:00:00.000000000',
>        '2050-11-01T00:00:00.000000000', '2050-12-01T00:00:00.000000000'],
>       dtype='datetime64[ns]')
> Coordinates:
>   * time     (time) datetime64[ns] 1800-01-01 1800-02-01 ... 2050-12-01
> Attributes:
>     bounds:        time_bnds
>     units:         months since 1800
>     calendar:      noleap
>     axis:          T
>     realtopology:  linear

The decoded time only goes to 2050 (it should go to 2100). It seems like this is an issue with `decode_time_units`. **Is that the correct interpretation?** Note that when I try to open either of these files using `xcdat.open_dataset()` I get a time vector that (incorrectly) begins in January 1800.

tomvothecoder:
> The decoded time only goes to 2050 (it should go to 2100). It seems like this is an issue with decode_time_units. Is that the correct interpretation? Note that when I try to open either of these files using xcdat.open_dataset() I get a time vector that (incorrectly) begins in January 1800.

I will investigate this issue using your example files once `/user_pub` is back online.

tomvothecoder:
> The decoded time only goes to 2050 (it should go to 2100). It seems like this is an issue with decode_time_units. Is that the correct interpretation? Note that when I try to open either of these files using xcdat.open_dataset() I get a time vector that (incorrectly) begins in January 1800.

Yes, the issue is related to `decode_time_units()`. 

For your example, which includes non-CF time units, the reference date of January 1800 is being set as the start for the time series. Instead, it should be January 1800 + the first coordinate value (600 months in this case when `decode_times=False`).

```
>>> ds_xr1 = xr.open_dataset(flist[0], decode_times=False)
>>> ds_xr1.time.units
'months since 1800'
>>> ds_xr1.time.data[0]
600.0
```

Related code:
https://github.com/XCDAT/xcdat/blob/69dd39d249fdf63a78e39ba44dca2e968b714e83/xcdat/dataset.py#L345-L353

I'll rollout a separate PR for this fix, thanks for the catch!

tomvothecoder:
Continuing work in PR #173, which will merge into this PR.

xcdat/dataset.py
```diff
@@ -161,22 +185,219 @@ def open_mfdataset(
...
+    first_file: Optional[Union[pathlib.Path, str]] = None
+    if isinstance(path, str) and "*" in path:
+        first_file = glob(path)[0]
+    elif isinstance(path, str) or isinstance(path, pathlib.Path):
+        first_file = path
+    elif isinstance(path, list):
+        if any(isinstance(sublist, list) for sublist in path):
+            first_file = path[0][0]  # type: ignore
+        else:
+            first_file = path[0]  # type: ignore
```
tomvothecoder, line 261: I updated this to support `pathlib.Path` and nested list of strings/`pathlib.Paths`.

tomvothecoder (approved):
This PR looks good to me now. I added an additional fix, which I've pointed out in a comment.

Feel free to merge after you do a final review.

pochedls:
@tomvothecoder: I am assuming these types of messages for `pytest` are fine?

> ....2021-12-23 14:15:35,270 [DEBUG]: dataset.py(infer_or_keep_var:461) >> This dataset contains more than one regular data variable ('tas', 'ts'). If desired, pass the `data_var` kwarg to reduce down to one regular data var.
> 2021-12-23 14:15:35,270 [DEBUG]: dataset.py(infer_or_keep_var:461) >> This dataset contains more than one regular data variable ('tas', 'ts'). If desired, pass the `data_var` kwarg to reduce down to one regular data var.
> ......2021-12-23 14:15:35,328 [DEBUG]: dataset.py(get_inferred_var:534) >> The data variable 'ts' was inferred from the Dataset attr 'xcdat_infer' for this operation.
> 2021-12-23 14:15:35,328 [DEBUG]: dataset.py(get_inferred_var:534) >> The data variable 'ts' was inferred from the Dataset attr 'xcdat_infer' for this operation.

tomvothecoder:
@pochedls Yeah, those are debug level messages that appear in the test suite. I might end up removing it because it will throw an error if `xcdat_infer` isn't set anyways and they get distracting in the test suite.

pochedls:
@tomvothecoder - This looks good to me. Could you merge it (or I can merge it with some guidance from you after the holidays)? Thanks for all your contributions to this bug/PR!

tomvothecoder:
@pochedls Thanks for the final review! I'll handle rebasing and merging the PR today.

PR #164, opened 2021-11-19
Adds horizontal regridding

### LINKED ISSUE #44: Add horizontal regridding
Blocked by #71 

Create time average of data with proper temporal weighting accounting for calendar peculiarities (e.g., from monthly mean, seasonal mean, or annual mean from monthly/sub-monthly data) 

Acceptance criteria:
- [ ] Decide if we want to wrap third-party library calls into utility method(s), or have user rely on library APIs directly
-  [ ] Ensure we can convert between 0:360 longitude and -180:180 longitude

- Does SciPy do this? 
- Wrap xESMF/ESMF? 
- Wrap TempestRemap?

#### Additional context

CDAT/CDMS uses SCRIP2, ESMF, and regrid2

Prototype: 
- [Regrid example](https://github.com/lee1043/XCDAT/blob/master/compare_cdat_xarray/Regrid_ex1.ipynb)

jasonb5:
## Description

Adds horizontal regridding via xESMF and ported regrid2 for grid types.

Grid types:
- Rectilinear
- Curvilinear

**Unstructured grids will be a future feature.**

The plan is to move regrid2 from c to pure numpy.

- Closes #44 

## TODO
- [x] Clean up gaussian grid code
- [x] Document/test gaussian grid code
- [x] Add global mean and zonal grid + documentation/tests
- [x] Add additional testing for xESMF backend
- [x] Port regrid2 from CDMS2 (Plan is for pure python)

## Checklist

If applicable:


chengzhuzhang:
Thank you so much for the PR @jasonb5 ! I will start some testing tomorrow.

durack1:
#jasonb5 this sounds great, am I right to assume that `xesmf` is the regridder, or we also have the cdms2 regrid2 library also available? If `xesmf` is available, this is great as it should deal with the complex ocean grids

jasonb5:
@durack1 That's correct we have both the `xesmf` and `regrid2` libraries available.

xcdat/regridder/accessor.py
```diff
@@ -0,0 +1,89 @@
...
+        Examples
+        --------
+        Import:
+
+        >>> import xcdat
+        >>> from xcdat.regridder import grid
+
+        Open a dataset and limit to a single variable:
+
+        >>> ds = xcdat.open_dataset("path/to/file.nc", var="tas")
```
pochedls, line 68: Note here (and elsewhere) that we no longer set a default `var`.

xcdat/regridder/accessor.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Create destination grid:
+
+        >>> out_grid = xcdat.regridder.grid.create_uniform_grid(-90, 90, 4.0, -180, 180, 5.0)
```
pochedls, line 72: Since you import `from xcdat.regridder import grid`, I think this can simply be `out_grid = grid.create_uniform_grid(-90, 90, 4.0, -180, 180, 5.0)`

xcdat/regridder/accessor.py
```diff
@@ -0,0 +1,89 @@
...
+            Dataset containing destination grid.
+        tool : str
+            Name of the regridding tool.
+        **options : Any
+            These options are passed to the tool being used for regridding.
+            See specific regridder documentation for available options.
+
+        Returns
+        -------
+        xr.Dataset
```
pochedls, line 51: I think this returns a DataArray:

```
# Imports
import xcdat
from xcdat.regridder import grid

# dataset to use
dpath = '/p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2/historical/r1i1p1f1/Amon/tas/gn/v20190308/*.nc'

# open dataset
ds = xcdat.open_mfdataset(dpath)

# create output grid
out_grid = grid.create_uniform_grid(-90, 90, 4.0, -180, 180, 5.0)

# regrid
ds2 = ds.regridder.regrid("tas", out_grid, tool="xesmf", method="bilinear")

# check type
type(ds2)
```

> xarray.core.dataarray.DataArray

xcdat/regridder/accessor.py
```diff
@@ -0,0 +1,89 @@
...
+class DatasetRegridderAccessor:
+    """xarray dataset accessor for access to regridding."""
+
+    def __init__(self, ds: xr.Dataset):
+        self._ds = ds
+
+    def regrid(
+        self,
+        data_var: str,
+        dst_grid: xr.Dataset,
```
pochedls, line 25: Should we have this default to one of the regridders? Maybe xesmf?

chengzhuzhang:
Hey @jasonb5, I created a jupyter notebook to cross validating horizontal regridding. There is one issue that I couldn't understand why. The note books compared linear/conservative regriding using 3 cases:
- cdms2
- xesmf and xarray
- xcdat. 
With xcdat, comparison between linearly interpolated data and conservatively regridded data  shows a much larger difference (maximum difference is 6.07 K for Sea surface temperature SST ln[37]). But the maximum difference is 0.77 ln[22]using xesmf and xarray. The xcdat result probably is erroneous, but I can't immediately tell why. Would you take a look at this https://github.com/XCDAT/xcdat_test/blob/master/validation/regridding/regrid_cdat_xarray_xcdat.ipynb
The data for testing is also available here: /p/user_pub/e3sm/zhang40/xcdat_test_e3sm

Thank you!

jasonb5:
@chengzhuzhang Thanks for putting this notebook together, I'll take a look at the issue.

tomvothecoder (changes requested):
My initial review consisting of mostly docstring updates, naming conventions, and type annotation updates.

pochedls:
I am working on some tests for horizontal regridding and I am hitting an early issue. I think the regridder is interpreting a height coordinate as a dimension for 3d [time, lat, lon] fields:

```
import xcdat

ngrid = xcdat.regridder.grid.create_uniform_grid(-90, 90, 4.0, -180, 180, 5.0)

fn = '/p/css03/cmip5_css01/data/cmip5/output1/NSF-DOE-NCAR/CESM1-BGC/historical/mon/atmos/Amon/r1i1p1/v20130213/tas/tas_Amon_CESM1-BGC_historical_r1i1p1_185001-200512.nc'
ds = xcdat.open_dataset(fn)
ds2 = ds.regridder.regrid("tas", ngrid, tool="xesmf", method="bilinear")
```

> KeyError: 'Z bounds were not found, they must be added.'

In the loop to get the bounds – do we need bounds for non horizontal fields (e.g., time)?

pochedls:
> I am working on some tests for horizontal regridding and I am hitting an early issue. I think the regridder is interpreting a height coordinate as a dimension for 3d [time, lat, lon] fields:
> 
> ```
> import xcdat
> 
> ngrid = xcdat.regridder.grid.create_uniform_grid(-90, 90, 4.0, -180, 180, 5.0)
> 
> fn = '/p/css03/cmip5_css01/data/cmip5/output1/NSF-DOE-NCAR/CESM1-BGC/historical/mon/atmos/Amon/r1i1p1/v20130213/tas/tas_Amon_CESM1-BGC_historical_r1i1p1_185001-200512.nc'
> ds = xcdat.open_dataset(fn)
> ds2 = ds.regridder.regrid("tas", ngrid, tool="xesmf", method="bilinear")
> ```
> 
> > KeyError: 'Z bounds were not found, they must be added.'
> 
> In the loop to get the bounds – do we need bounds for non horizontal fields (e.g., time)?

@jasonb5 - I am wondering if we could replace portions of [this code](https://github.com/xCDAT/xcdat/blob/578e35867e0282f21cbcfdee4a50ef040042ba1d/xcdat/regridder/xesmf.py#L104-L118) with a call to [bounds.add_missing_bounds](https://github.com/xCDAT/xcdat/blob/main/xcdat/bounds.py#L95) in order to address this issue. The function only adds bounds for supported axes (T, X, and Y) - I don't think we need bounds for the vertical axis.

jasonb5:
@pochedls `add_missing_bounds` doesn't need to be called. This portion of code was suppose to preserve bounds for non-spatial axes, I'll just wrap it in a `try/except` so if the bounds exist they're added to the new dataset otherwise they're skipped.

tomvothecoder:
Hi @jasonb5, Jill pointed out a potential bug in her validation notebook at our last xCDAT meeting. 

We were aiming to have this PR merged by 5/18 (Wed), but we can push it back a few days for any final fixes.
Do you think this PR is on track to be reviewed and merged by 5/20 (Fri)?

jasonb5:
@tomvothecoder Fri (5/20) will work. I should have my last masking changes in today and am reviewing the issue Jill found.

jasonb5:
@chengzhuzhang @tomvothecoder 
The issue in Jill's example is related to the bounds. The bounds in the source don't appear to be correct.
```python
[[-89., -90],
 [-88., -89],
 ...
 [89., 88.],
 [90., 89.]]
```
In the xarray/xesmf example the bounds in the file are not used and rather are generated from cdms2's bounds (which look the same as above) but they're changed with the following code.
 ```python
lon_b_1 = numpy.unique(numpy.sort(lon_b_var1.flatten())) # reform bounds array
```
This reorders the bounds.
```python
[-90, -89, -88, ..., 88, 89, 90]
```
In the xCDAT example, the bounds from the file are used, creating the max difference of `6.073` rather than `0.7771759` from the xarray/xesmf example.
If you drop the source bounds and generate new ones in the xCDAT example, the results are the same as the xarray/xesmf example.
```python
dsx1 = dsx1.drop_vars(["lat_bnds", "lon_bnds"])
dsx1 = dsx1.bounds.add_missing_bounds()
dsx2 = dsx2.drop_vars(["latitude_bnds", "longitude_bnds"])
dsx2 = dsx2.bounds.add_missing_bounds()
```

chengzhuzhang:
@jasonb5 thank you so much to find the cause of the discrepancy! I think it might be a good feature to consider for xcdat to rectify this kind of latitude bounds? I have to check if this case is common enough. 

I have to take some time off this week, but hope not to delay the release schedule. @tomvothecoder @jasonb5 I'm wondering if we can have `bounds` to handle this case?

tomvothecoder:
Nice job figuring out the source of the issue @jasonb5.

> @jasonb5 thank you so much to find the cause of the discrepancy! I think it might be a good feature to consider for xcdat to rectify this kind of latitude bounds? I have to check if this case is common enough.

This issue is similar to the situation that Steve presented, which involved being unable to open multi-file datasets with conflicting lat bounds as a result of poor data quality.

This one also seems tough to handle in a generalized way, but it might still be a good exercise to think of a possible solution.

Off the top of my head, we'd have to implement some sophisticated pre-checks in the initial opening of the dataset/multi-file dataset to validate the correctness of the bounds within the file (correctness needs to be defined). This might lead to a lot of implementation complexities, edge cases to cover, and performance hits since validation is applied to all files including those without issues.

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+    else:
+        lower = bounds[:, 1]
+        upper = bounds[:, 0]
+
+    return lower, upper
+
+
+def map_latitude(src: xr.DataArray, dst: xr.DataArray) -> Tuple[List, List]:
+    """
+    Map source to destination latitude.
```
tomvothecoder, line 35:
```suggestion
    """Map source to destination latitude.
```

Add description in docstring on how the algorithm/logic works

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+        weight = np.sin(np.deg2rad(north_bounds)) - np.sin(np.deg2rad(south_bounds))
+
+        weights.append(weight.values.reshape(contrib.shape[0], 1))
+
+    return mapping, weights
+
+
+def pertub(value):
+    """
+    Pertub a value.
```
tomvothecoder, line 74:
```suggestion
    """Pertub a value.
```

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+        south_bounds = np.maximum(dst_south[i], src_south[contrib])
+
+        weight = np.sin(np.deg2rad(north_bounds)) - np.sin(np.deg2rad(south_bounds))
+
+        weights.append(weight.values.reshape(contrib.shape[0], 1))
+
+    return mapping, weights
+
+
+def pertub(value):
```
tomvothecoder, line 72: Missing type annotation for `value` and return type

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+    """
+    Pertub a value.
+
+    Parameters
+    ----------
+    value :
```
tomvothecoder, line 78: Add type annotation

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+        Value to pertub.
+    """
```
tomvothecoder, line 80:
```suggestion
        Value to pertub.
     
     Returns
     -------
     <RETURN_TYPE>
         Description
    """
```

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+        if shifted_src_west[-1] > west_most:
+            shifted_src_west[-1] += -360.0
+            shifted_src_east[-1] += -360.0
+
+    return shifted_src_west, shifted_src_east, shift
+
+
+def map_longitude(src: xr.DataArray, dst: xr.DataArray) -> Tuple[List, List]:
+    """
+    Map source to destination longitude.
```
tomvothecoder, line 168:
```suggestion
    """Map source to destination longitude.
```
Add description in docstring on how the algorithm/logic works

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+            Dataset with variable on the destination grid.
+
+        Raises
+        ------
+        KeyError
+            If data variable does not exist in the Dataset.
+
+        Notes
+        -----
+
```
tomvothecoder, line 479:
```suggestion
```

xcdat/regridder/xesmf.py
```diff
@@ -0,0 +1,192 @@
...
+            method,
+            periodic=periodic,
+            extrap_method=extrap_method,
+            extrap_dist_exponent=extrap_dist_exponent,
+            extrap_num_src_pnts=extrap_num_src_pnts,
+        )
+
+    def horizontal(self, data_var: str, ds: xr.Dataset) -> xr.Dataset:
+        """
+        Regrid ``data_var`` in ``ds`` to output grid.
```
tomvothecoder, line 119:
```suggestion
        """Regrid ``data_var`` in ``ds`` to output grid.
```

Add description in docstring on how the algorithm/logic works

xcdat/regridder/xesmf.py
```diff
@@ -0,0 +1,192 @@
...
+        periodic: bool = False,
+        extrap_method: str = None,
+        extrap_dist_exponent: float = None,
+        extrap_num_src_pnts: int = None,
+        **options,
+    ):
+        """
+        Wrapper class for xESMF regridder class.
+
+        Available options: ``method``, ``periodic``, ``extrap_method``, ``extrap_dist_exponent``, ``extrap_num_src_pnts``
```
tomvothecoder, line 36:
```suggestion
```

The Parameters section covers these

xcdat/regridder/grid.py
```diff
@@ -0,0 +1,530 @@
...
+    # bounds = (180.0 / np.pi) * np.arcsin(bessel_zeros)
+
+    return points, weights
+
+
+def _create_gaussian_axis(nlats: int) -> Tuple[xr.DataArray, xr.DataArray]:
+    """Create Gaussian axis.
+
+    Creates a Gaussian axis of `nlats`.
+
```
tomvothecoder, line 213:
```suggestion
    """Creates a Gaussian axis of `nlats`.
```

xcdat/regridder/grid.py
```diff
@@ -0,0 +1,530 @@
...
+    )
+
+    grid = grid.bounds.add_missing_bounds()
+
+    return grid
+
+
+def create_global_mean_grid(grid: xr.Dataset) -> xr.Dataset:
+    """
+    Creates a global mean grid.
```
tomvothecoder, line 405:
```suggestion
    """Creates a global mean grid.
```

xcdat/regridder/grid.py
```diff
@@ -0,0 +1,530 @@
...
+                },
+                attrs=grid.bounds.get_bounds("lon").attrs.copy(),
+            ),
+        },
+    )
+
+
+def create_zonal_grid(grid: xr.Dataset) -> xr.Dataset:
+    """
+    Creates a zonal grid.
```
tomvothecoder, line 473:
```suggestion
    """Creates a zonal grid.
```

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+        Applies regridding to input data.
+
+        Parameters
+        ----------
+        input_data : np.ndarray
+            Input multi-dimensional array on source grid.
+        axis_sizes : Dict[str, int]
+            Mapping of axis name e.g. ("X", "Y", etc) to output sizes.
+        ordered_axis_names : List[str]
+            List of axis name in order of dimenions of ``input_data``.
```
tomvothecoder, line 348:
```suggestion
            List of axis name in order of dimensions of ``input_data``.
```

xcdat/regridder/xesmf.py
```diff
@@ -0,0 +1,192 @@
...
+        periodic : bool
+            Treat longitude as periodic. Used for global grids.
+        extrap_method : str
+            Extrapolation method. Options are
+               - inverse_dist
+               - nearest_s2d
+        extrap_dist_exponent : float
+            The exponent to raise the distance to when calculating weights for the extrapolation method.
+        extrap_num_src_pnts : int
+            The number of source points to use for the extrapolation methods that use more than one source point.
```
tomvothecoder, line 61:
```suggestion
        extrap_dist_exponent : float
            The exponent to raise the distance to when calculating weights for 
            the extrapolation method.
        extrap_num_src_pnts : int
            The number of source points to use for the extrapolation methods 
            that use more than one source point.
```

Fit docstring within 80 character line ruler

xcdat/regridder/xesmf.py
```diff
@@ -0,0 +1,192 @@
...
+            Dataset with variable on the destination grid.
+
+        Raises
+        ------
+        KeyError
+            If data variable does not exist in the Dataset.
+
+        Notes
+        -----
+
```
tomvothecoder, line 140:
```suggestion
```

tomvothecoder (changes requested):
Hey Jason, thanks for the PR. I appreciate that you followed the code style and formatting that is consistent with other `xcdat`'s APIs. Great work!

I had some minor review suggestions. Some are nit-picking, but it should help improve consistency/usage some more.

I didn't really review the implementation itself since I am confident in your implementation and unit testing. I plan on doing that in the future so I become more versed in the regridding API.

tomvothecoder:
@pochedls and I discussed his review process. He is currently looping through files and regridding with CDAT and xCDAT, then comparing the results. He will provide an update with his findings.

Based on Steve's progress, we might need to push back the target merge date (hopefully by mid to end of next week). We should prioritize making sure this API is ready for external testers in `v1.0.0rc1` around 6/1/22, instead of rushing a merge.

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+    def __init__(
+        self, input_grid: xr.Dataset, output_grid: xr.Dataset, **options: Dict[str, Any]
+    ):
+        """
+        Pure python implementation of the regrid2 horizontal regridder from
+        CDMS2's regrid2 module.
+
+        Regrid data from ``input_grid`` to ``output_grid``.
+
+        Available options: None
```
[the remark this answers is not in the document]
pochedls: Yes - it took me a minute to see this.

pochedls:
> @pochedls `add_missing_bounds` doesn't need to be called. This portion of code was suppose to preserve bounds for non-spatial axes, I'll just wrap it in a `try/except` so if the bounds exist they're added to the new dataset otherwise they're skipped.

@jasonb5 - I think this same issue also applies to regrid2, though the fix for esmf seems to be working with no problem.

pochedls:
## Test 1: Surface air temperature bilinear regrid

### Design:

* **Data**: CMIP6 tas files (one dataset per model)
* **Tool / Method**: ESMF / Bilinear
* **Target grid**: `xcdat.regridder.grid.create_uniform_grid(-88.75, 88.76, 2.5, 1.25, 360., 2.5)`
* **Other details**: Assume periodicity True for both xcdat/CDAT, test with `np.allclose()` [and equal_nan=True]

### Result:
* Agreement within tolerance in 62 of 63 instances
* xcdat faster in 53 of 63 instances
* Model that disagreed was ICON-ESM-LR (see note on unstructured grid under take-aways)

## Test 2: Surface air temperature conservative regrid

### Design: 

As with Test 1, but for conservative regridding

### Result:
* Agreement within tolerance in 57 of 63 instances
* xcdat faster in 55 of 63 instances
* Note that I originally dropped and re-created bounds (in xcdat), but this led to much larger disagreement
* Models that disagreed included 4 model families:
    * CNRM-CM6-1, CNRM-CM6-1-HR, CNRM-ESM2-1, FGOALS-f3-L, ICON-ESM-LR, MCM-UA-1-0
    * For CNRM-CM6-1 (and likely the other CNRM models), it seems like there may be an issue at the poles (I suspect xcdat may not be correct)
    * For FGOALS-f3-L and MCM-UA-1-0, it seems like the issue is at the prime meridian (even though I specified periodic=True)
    * ICON-ESM-LR (see note re: unstructured grid)

## Test 3: Sea ice bilinear regrid

### Design

As with Test 1, but for sea ice

### Result

* Agreement within tolerance in 25 of 58 instances
* Many models had errors in xcdat (e.g., `ValueError: ESMC_FieldRegridStore failed with rc = 506. Please check the log files (named "*ESMF_LogFile").`):
    * AWI-ESM-1-1-LR, CAS-ESM2-0, CESM2, CESM2-FV2, CESM2-WACCM, CESM2-WACCM-FV2, CIESM, CMCC-CM2-HR4, CNRM-CM6-1, CNRM-CM6-1-HR, CNRM-ESM2-1, EC-Earth3, EC-Earth3-AerChem, EC-Earth3-CC, EC-Earth3-Veg, EC-Earth3-Veg-LR, GFDL-CM4, HadGEM3-GC31-MM, ICON-ESM-LR, IPSL-CM6A-LR, IPSL-CM6A-LR-INCA, MPI-ESM1-2-HR, NorESM2-MM, SAM0-UNICON
    * Example file with error: `/p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2/historical/r10i1p1f1/SImon/siconc/gn/v20190313/siconc_SImon_CESM2_historical_r10i1p1f1_gn_200001-201412.nc`
* Models that disagreed included 7 model families:
    * BCC-CSM2-MR, BCC-ESM1, FGOALS-f3-L, FGOALS-g3, FIO-ESM-2-0, IPSL-CM5A2-INCA, MIROC-ES2H, NorESM2-LM, TaiESM1

## Test 4: Sea ice conservative regrid

### Design

As with Test 3, but with conservative re-gridding

### Result

* Agreement within tolerance in 10 of 58 instances
* Many models had errors in xcdat (e.g., `ValueError: ESMC_FieldRegridStore failed with rc = 506. Please check the log files (named "*ESMF_LogFile").`):
    * AWI-ESM-1-1-LR, BCC-CSM2-MR, BCC-ESM1, CAS-ESM2-0, CESM2, CESM2-FV2, CESM2-WACCM, CESM2-WACCM-FV2, CIESM, CMCC-CM2-HR4, CMCC-CM2-SR5, CMCC-ESM2, CNRM-CM6-1, CNRM-CM6-1-HR, CNRM-ESM2-1, EC-Earth3, EC-Earth3-AerChem, EC-Earth3-CC, EC-Earth3-Veg, EC-Earth3-Veg-LR, FGOALS-f3-L, FGOALS-g3, FIO-ESM-2-0, GFDL-CM4, HadGEM3-GC31-MM, ICON-ESM-LR, IPSL-CM5A2-INCA, IPSL-CM6A-LR, IPSL-CM6A-LR-INCA, MPI-ESM-1-2-HAM, MPI-ESM1-2-HR, MPI-ESM1-2-LR, MRI-ESM2-0, NESM3, NorESM2-LM, NorESM2-MM, TaiESM1
* Models that disagreed included 8 model families:
    * ACCESS-CM2, ACCESS-ESM1-5, CAMS-CSM1-0, CanESM5, CanESM5-CanOE, HadGEM3-GC31-LL, MIROC-ES2H, MIROC6, NorCPM1, SAM0-UNICON, UKESM1-0-LL

## Take-aways:

* Data on unstructured grids (e.g., `/p/css03/esgf_publish/CMIP6/CMIP/MPI-M/ICON-ESM-LR/historical/r1i1p1f1/Amon/tas/gn/v20210215/`) yields an error (`TypeError: buffer is too small for requested array`) that is not readily interpretable. We might be able to determine rules to provide a better exception (e.g., CDAT states: `AttributeError: 'TransientGenericGrid' object has no attribute 'toCurveGrid'`).
* One model seems to have a *possible* xcdat-related issue with conservative re-gridding at the poles (CNRM-CM6-1)
* Two models seem to have *possible* xcdat-related issues with conservative re-gridding at the prime meridian (FGOALS-f3-L and MCM-UA-1-0)
* Need to further diagnose the exceptions when using sea ice (probably related to the diverse grids sea ice fields are on)
* Same for models that do not agree (within tolerance) for sea ice and conservative grids. In some cases, I think the difference may simply be in the NaN field.

setup.py
```diff
@@ -20,6 +20,7 @@
     "numpy",
     "pandas",
     "xarray",
+    "pangeo-xesmf",
```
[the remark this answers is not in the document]
jasonb5: `pangeo-xesmf` is specific to pypi, this is already covered in `conda-env/dev.yml` and `conda-env/ci.yml` with `xesmf`.

pochedls:
I wanted to add some specific examples for datasets with exceptions / errors in xcdat. To reproduce individual errors, you can run the following code with each dpath:

```
import xcdat

ngrid = xcdat.regridder.grid.create_uniform_grid(-88.75, 88.76, 2.5, 1.25, 360., 2.5)
dpath = '...'
vid = 'siconc'
xcdat_tool = 'xesmf'
xcdat_rmethod = 'conservative'

ds = xcdat.open_mfdataset(dpath + '*.nc')
ds2 = ds.regridder.horizontal(vid, ngrid, tool=xcdat_tool, method=xcdat_rmethod, periodic=True)
ds.close()
```

* **TypeError: buffer is too small for requested array** OR **MemoryError: Unable to allocate 413. GiB for an array with shape (235403, 235403) and data type float64**
   * `/p/css03/esgf_publish/CMIP6/CMIP/AWI/AWI-ESM-1-1-LR/historical/r1i1p1f1/SImon/siconc/gn/v20200212/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/MPI-M/ICON-ESM-LR/historical/r1i1p1f1/SImon/siconc/gn/v20210215/`
* **ValueError: Tool "Receive multiple variables for key 'longitude': {'lon', 'longitude'}. Expected only one. Please pass a list ['longitude'] instead to get all variables matching 'longitude'." does not exist, valid choices ['xesmf', 'regrid2']**
   * `/p/css03/esgf_publish/CMIP6/CMIP/BCC/BCC-CSM2-MR/historical/r1i1p1f1/SImon/siconc/gn/v20200218/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/BCC/BCC-ESM1/historical/r1i1p1f1/SImon/siconc/gn/v20200218/`
* **ValueError: Tool 'lon_b' does not exist, valid choices ['xesmf', 'regrid2']**
   * `/p/css03/esgf_publish/CMIP6/CMIP/CAS/CAS-ESM2-0/historical/r1i1p1f1/SImon/siconc/gn/v20201225/`
   * `/p/css03/scratch/cmip6/CMIP/CAS/FGOALS-f3-L/historical/r1i1p1f1/SImon/siconc/gn/v20191031/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/CAS/FGOALS-g3/historical/r4i1p1f1/SImon/siconc/gn/v20210108/`
* **ValueError: ESMC_FieldRegridStore failed with rc = 506. Please check the log files (named "\*ESMF_LogFile").**
   * `/p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2/historical/r10i1p1f1/SImon/siconc/gn/v20190313/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2-FV2/historical/r1i1p1f1/SImon/siconc/gn/v20191120/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2-WACCM/historical/r1i1p1f1/SImon/siconc/gn/v20190227/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2-WACCM-FV2/historical/r1i1p1f1/SImon/siconc/gn/v20191120/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/CMCC/CMCC-CM2-HR4/historical/r1i1p1f1/SImon/siconc/gn/v20200904/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/CMCC/CMCC-CM2-SR5/historical/r10i1p2f1/SImon/siconc/gn/v20220401/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/CMCC/CMCC-ESM2/historical/r1i1p1f1/SImon/siconc/gn/v20210114/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/CNRM-CERFACS/CNRM-CM6-1/historical/r10i1p1f2/SImon/siconc/gn/v20190125/`
   * `/p/css03/scratch/cmip6/CMIP/CNRM-CERFACS/CNRM-CM6-1-HR/historical/r1i1p1f2/SImon/siconc/gn/v20191021/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/CNRM-CERFACS/CNRM-ESM2-1/historical/r11i1p1f2/SImon/siconc/gn/v20200408/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/EC-Earth-Consortium/EC-Earth3/historical/r101i1p1f1/SImon/siconc/gn/v20200919/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/EC-Earth-Consortium/EC-Earth3-AerChem/historical/r1i1p1f1/SImon/siconc/gn/v20200624/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/EC-Earth-Consortium/EC-Earth3-CC/historical/r10i1p1f1/SImon/siconc   * `/gn/v20210630/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/EC-Earth-Consortium/EC-Earth3-Veg/historical/r10i1p1f1/SImon/siconc   * `/gn/v20210523/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/EC-Earth-Consortium/EC-Earth3-Veg-LR/historical/r1i1p1f1/SImon/siconc/gn/v20200919/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/EC-Earth-Consortium/EC-Earth3-Veg-LR/historical/r1i1p1f1/SImon/siconc/gn/v20200919/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/FIO-QLNM/FIO-ESM-2-0/historical/r1i1p1f1/SImon/siconc/gn/v20191127/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/MOHC/HadGEM3-GC31-MM/historical/r1i1p1f3/SImon/siconc/gn/v20191207/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/IPSL/IPSL-CM5A2-INCA/historical/r1i1p1f1/SImon/siconc/gn/v20200729/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/IPSL/IPSL-CM6A-LR/historical/r10i1p1f1/SImon/siconc/gn/v20180803/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/IPSL/IPSL-CM6A-LR-INCA/historical/r1i1p1f1/SImon/siconc/gn/v20210216/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/HAMMOZ-Consortium/MPI-ESM-1-2-HAM/historical/r1i1p1f1/SImon/siconc/gn/v20190627/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/MPI-M/MPI-ESM1-2-HR/historical/r10i1p1f1/SImon/siconc/gn/v20190710/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/MPI-M/MPI-ESM1-2-LR/historical/r10i1p1f1/SImon/siconc/gn/v20190710/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/MRI/MRI-ESM2-0/historical/r10i1p1f1/SImon/siconc/gn/v20210909/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/NUIST/NESM3/historical/r1i1p1f1/SImon/siconc/gn/v20190704/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/NCC/NorESM2-LM/historical/r2i1p1f1/SImon/siconc/gn/v20190920/`
   * `/p/css03/esgf_publish/CMIP6/CMIP/AS-RCEC/TaiESM1/historical/r1i1p1f1/SImon/siconc/gn/v20200630/`
* **UFuncTypeError: ufunc 'subtract' cannot use operands with types dtype('O') and dtype('<m8[ns]')**
   * `/p/css03/esgf_publish/CMIP6/CMIP/THU/CIESM/historical/r1i1p1f1/SImon/siconc/gn/v20200420/`
* **AttributeError: cf_xarray can't wrap attribute 'dims' because there are multiple values for 'X'. There is no unique mapping from 'X' to a value in 'dims'.**
   * `/p/css03/esgf_publish/CMIP6/CMIP/NOAA-GFDL/GFDL-CM4/historical/r1i1p1f1/SImon/siconc/gn/v20180701/`
* **MergeError: conflicting values for variable 'vertices_longitude' on objects to be combined. You can skip this check by specifying compat='override'.**
   * `/p/css03/esgf_publish/CMIP6/CMIP/NCC/NorESM2-MM/historical/r1i1p1f1/SImon/siconc/gn/v20191108/`

xcdat/regridder/grid.py
```diff
@@ -0,0 +1,530 @@
+from typing import Tuple
+
+import numpy as np
+import xarray as xr
+
+# First 50 zeros for the bessel function
+BESSEL_LOOKUP = [
```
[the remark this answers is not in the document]
jasonb5: Taken from original regrid2 code (https://github.com/CDAT/cdms/blob/dd41a8dd3b5bac10a4bfdf6e56f6465e11efc51d/regrid2/Src/_regridmodule.c#L3390-L3402) I'll add a link in the code.

pochedls:
I wanted to add some specific examples for datasets where an exception / error is not thrown, but the re-gridding results look questionable. 

## Case 1: conservative tas CNRM-CM6-1
**Description**: Large drop (~5K) in temperature near south pole
**Dataset**: `/p/css03/esgf_publish/CMIP6/CMIP/CNRM-CERFACS/CNRM-CM6-1/historical/r10i1p1f2/Amon/tas/gr/v20190125/`
**Other cases**: Similar results for other CNRM models

## Case 2: conservative tas FGOALS-f3-L
**Description**: Large drop (~25K) in temperature near prime meridian in tropics
**Dataset**: `/p/css03/esgf_publish/CMIP6/CMIP/CAS/FGOALS-f3-L/historical/r1i1p1f1/Amon/tas/gr/v20190927/`
**Other cases**: MCM-UA-1-0 has a temperature of 522 K near the prime meridian

## Sea ice fields (conservative or bilinear)

For these fields xcdat and CDAT aren't very by eye (and when they are *xcdat* looks like it is doing the right thing). In some cases, the differences seem to be slight differences in NaN / masking. I'm not totally sure how to evaluate the sea ice fields.

durack1:
@pochedls the sea-ice changing mask complicates the ocean grids further. Another way to tackle the same grid issue is switchout `siconc` (I assume you are using) with `tos` which in many instances will provide you with the same grid to test, but not the varying mask (if you are indeed looking at time varying data). If you're using `siconca` then apologies for the distraction

jasonb5:
@pochedls Thanks for reporting these, I'll begin looking into them.

setup.py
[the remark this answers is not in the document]
tomvothecoder: Got it, thanks

chengzhuzhang:
@pochedls Nice tests! I took brief look at the datasets in case 1: CNRM-CM6-1. This model output itself doesn't have lat lon bounds installed, so I assume xcdat auto generated bounds for it. But need to look into it carefully to understand. The case 2 FGOALS-f3-L has suspicious bounds.. Too bad, there are more problems with sea-ice field, the `rc506` issue is concerning, I saw some similar issues filed at xesmf repo. Anyway, which cases do you like me to look into? Perhaps we won't resolve all of them for this release.

pochedls:
> @pochedls the sea-ice changing mask complicates the ocean grids further. Another way to tackle the same grid issue is switchout `siconc` (I assume you are using) with `tos` which in many instances will provide you with the same grid to test, but not the varying mask (if you are indeed looking at time varying data). If you're using `siconca` then apologies for the distraction

Using tos doesn't really change the result compared to siconc, though it is important to to account for masking before compating xcdat/CDAT. Setting `tos[tos.mask] = 0` in CDAT and `tos[np.isnan(tos)] = 0` in xcdat improves agreement when using `np.allclose()`.

conda-env/dev.yml
```diff
@@ -15,6 +15,8 @@ dependencies:
     - numpy=1.22.3
     - pandas=1.4.1
     - xarray=2022.3.0
+    - xesmf=0.6.2
```
tomvothecoder, line 18: @chengzhuzhang found that `xcdat` and our conda environments break with this error:
```
File ~/miniconda3/envs/xcdat_test_dev/lib/python3.10/site-packages/xesmf/smm.py:7, in <module>
      4 import warnings
      5 from pathlib import Path
----> 7 import numba as nb
      8 import numpy as np
      9 import sparse as sps

ModuleNotFoundError: No module named 'numba'
```

@jasonb5, the conda-forge feedstock recipe for `xesmf=0.6.2` doesn't include `numba` as a dependency. This doesn't affect the PyPI release (`pangeo-xesmf`). I opened up an issue in pangeo-data/xESMF so that can fix it: https://github.com/pangeo-data/xESMF/issues/170

**In the interim, we need to include `numba` in our conda envs and https://github.com/conda-forge/xcdat-feedstock/issues/2.**

setup.py
```diff
@@ -21,6 +21,7 @@
...
```
tomvothecoder, line 24: We should delete the `install_requires` section in `setup.py` so that we don't use PyPI dependencies in our conda-env environments.

I foudn that the xEMSF release on PyPI (`pangeo-xesmf`) includes `numba` as a sub-dependency, while the conda-forge version (`xemsf`) does not. This means the conda-forge version breaks if `numba` isn't included in the environment or as a dependency of `xcdat`.

https://github.com/pangeo-data/xESMF/issues/170

chengzhuzhang:
@jasonb5 hey, I was looking at cases 1 and 2 from @pochedls's diagnostics https://github.com/xCDAT/xcdat/pull/164#issuecomment-1132279922
- Both datasets has `height` coordinate (which don't really have values), and xcdat is trying loop through axes to add bounds, and had the error "KeyError: "No results found for 'Z'."". My workaround is to drop `height` then do the regrinding. I think this may need to be fixed before merging this PR. 

- the `create_uniform_grid` seems generating inconsistent results compared with cdms2:
   e.x. `ngrid = xcdat.regridder.grid.create_uniform_grid(-88.75, 88.75, 2.5, 1.25, 360., 2.5)`  gives (71, 144) shape with lat range from (-88.75 to 86.25) 
while `ngrid2 = cdms2.grid.createUniformGrid(-88.75, 72, 2.5, 1.25, 144, 2.5)` gives (72, 144) grids with lat range from (-88.75, 88.75)
I would expect both command should result matching grids...

- As Steve has pointed out, both cases has large difference (around beginning/end of latitude/longitude) in resulted regridded fields between xcdat and cdms2. xcdat probably provides incorrect answer. Together with the other errors in sea ice case, we perhaps won't be address all of them in this PR. I think it would be great we can address bullet 1 and perhaps together with bullet 2 (if there is an issue) to have this PR merged for the upcoming release, and can take time to work on the kinks? @jasonb5 @tomvothecoder @pochedls does this sound okay to you?

jasonb5:
@chengzhuzhang This sounds good to me, I'll start working on the first two points.

Now that JupyterHub is back up I'll be looking at the remaining files today and should have an idea of what's going on.

pochedls:
Just briefly weighing in on a couple of issues. 

* I'm was surprised @chengzhuzhang got a Key Error. This was not happening last week (but I just did a `git pull`) and now get this error if that helps.
* Regarding the grid generation, notice that I had a slightly different argument for the upper bound ([here](https://github.com/xCDAT/xcdat/pull/164#issuecomment-1132027459)) in xcdat. This is because @jasonb5 is using `numpy.arange` to generate the axes of the grid ([here](https://github.com/xCDAT/xcdat/blob/aaaa6dc70713d81e8f489e7181bdf0a420efeaea/xcdat/regridder/grid.py#L337)). I think this behavior is probably okay and would generally be expected by Python users. I have my own `create_grid` function (below) that I want to propose (pass in an array of lat and lon), but I thought we could add that down the road since there is a lot going on in this PR. 

```
def create_grid(lat: np.ndarray, lon: np.ndarray) -> xr.Dataset:
    """
    Creates a rectilinear for a given latitude and longitude array.
    Sets appropriate attributes for lat/lon axis.

    Parameters
    ----------
    lat : np.ndarray
        Array of latitude values
    lon : np.ndarray
        Array of latitude values

    Returns
    -------
    xr.Dataset
        Dataset lat/lon grid.

    Examples
    --------
    Create uniform 2.5 x 2.5 degree grid:
    >>> lat = np.arange(-88.75, 90, 2.5)
    >>> lon = np.arange(1.25, 360, 2.5)
    >>> xcdat.regridder.grid.create_grid(lat, lon)
    """

    latAxis = xr.DataArray(name='lat', data=lat, dims=['lat'], attrs={"units": "degrees_north", "axis": "Y"})
    lonAxis = xr.DataArray(name='lon', data=lon, dims=['lon'], attrs={"units": "degrees_east", "axis": "X"})

    grid = xr.Dataset(coords={"lat": latAxis, "lon": lonAxis})
    grid = grid.bounds.add_missing_bounds()

    return grid
```

jasonb5:
@tomvothecoder This time frame sounds good, there's not too much more to fix. Then we can work through new issues case by case.

@pochedls I'm inclined to just add this function now, it's pretty simple. What was the adjusted value you used for the upper bounds?

Do we still want the behavior of `create_uniform_grid` to exactly match CDAT's or just document that a small change to the bounds value is required?

pochedls:
> @pochedls I'm inclined to just add this function now, it's pretty simple. What was the adjusted value you used for the upper bounds?

Re: adding this function. This sounds great - thanks! `np.arange()` is not inclusive for the upper bound, so you just have to choose a value slightly larger than the upper bound (which is what I did in creating ngrid [here](https://github.com/xCDAT/xcdat/pull/164#issuecomment-1132027459)). We could do this generically by changing [this line](https://github.com/xCDAT/xcdat/blob/aaaa6dc70713d81e8f489e7181bdf0a420efeaea/xcdat/regridder/grid.py#L337) to:

```
return xr.DataArray(
        name=name, data=np.arange(start, stop+delta/2., delta), dims=[name], attrs=attrs
    )

```

> Do we still want the behavior of `create_uniform_grid` to exactly match CDAT's or just document that a small change to the bounds value is required?

I don't have strong feelings about this. As the documentation is written, I would kind of expect the stopping value to be included in the lat/lon axis. So maybe it is worth doing this. I'll leave this to others to decide (is this an issue of what is intuitive to CDAT users versus what is Pythonic?). 
```    stop : float
        Stopping latitude point.
```

Also, I noticed that you may want to change "latitude" to "latitude/longitude" [here](https://github.com/xCDAT/xcdat/blob/aaaa6dc70713d81e8f489e7181bdf0a420efeaea/xcdat/regridder/grid.py#L321).

chengzhuzhang:
I don't have a preference if to match exactly the behavior of `create_uniform_grid` to cdat. but really like the idea to generically handle the upper bounds, by Steve's suggestion:
`return xr.DataArray(
        name=name, data=np.arange(start, stop+delta/2., delta), dims=[name], attrs=attrs
    )`

jasonb5:
This upper bound solution is good with me, I will add it.

chengzhuzhang:
Okay, I tried to look at case 1 and case2 [here](https://github.com/xCDAT/xcdat/pull/164#issuecomment-1132279922) a little bit more. The founding from @pochedls is that in both case, at the boundary of the data (polar and/or prime meridian), xcdat is off and in-correct compared to cdat. I think the problem is the way how weights are calculated at these data boundaries, compare with the so called`conservative` method in xesmf verses that in cdms2. 

There are two ways of normalization when producing weights in ESMF: https://earthsystemmodeling.org/docs/release/ESMF_7_0_1/esmpy_doc/html/NormType.html
for xesmf:
`conservative` corresponds to `DSTAREA(Destination area normalization)` in ESMF
`conservative_norm` corresponds to `FRACAREA(Fraction area normalization)` in ESMF
The difference primarily show up at the data boundary, as depicted here: https://github.com/JiaweiZhuang/xESMF/issues/17

I suspect the `conservative` option used in cdms2 actually is more of like using `conservative_norm` in xesmf, by which in my test, both cases provide more closer results. @pochedls maybe you could try using this option and loop through the models to see if we can get better match?

pochedls:
@chengzhuzhang: `conservative_normed` makes things look a lot better. The differences are reduced to order ~0.1 K for both CNRM and FGOALS (tas conservative). It still doesn't pass the test for `np.allclose()`, but it is no longer obvious that xcdat is doing the wrong thing...

chengzhuzhang:
@pochedls Thank you for testing and it is great news! 
I'm not sure if there is a good way to let the users know the underlying normalization mechanism for the conservative method. We can recommend `conservative_normed` for most cases in documentation?

pochedls:
> For the remaining 0.1 K difference, I'm tagging @taylor13 to see if we can get some guidance.
> 
> Hi Karl, @pochedls and I are working on cross-validation for conservative method from ESMF, comparing the implementation in xcdat and cdat. Using the conservative method (weights normalized by fraction area, mentioned [here](https://github.com/xCDAT/xcdat/pull/164#issuecomment-1146285637)) from ESMF, the difference between xcdat and cdms2 reduces to ~0.1 K for `tas` , but it doesn't pass the `np.allclose()` test (for two model families, i think the differences are much smaller for other models). In this case, could you please advice us if we can consider the xcdat implementation is accurate enough, or if there is some special treatment in cdms2 that we should apply? Thank you in advance...

As a sanity check, we could calculate the global average time series and difference it with the global average time series after regridding. For conservative regridding, the difference should be zero.

jasonb5:
@chengzhuzhang @pochedls 
I dug around the CDAT code to figure out what `norm_type` was being used. It looks like CDAT never sets a specific value so it defaults to `DSTAREA` (https://earthsystemmodeling.org/esmpy_doc/release/latest/html/regrid.html#ESMF.api.regrid.Regrid)

chengzhuzhang:
@pochedls I'm thinking in the same line, good opportunity to exercise the spatial average routine :)

taylor13:
Is the 0.1 K difference a difference in spatial mean or a typical difference (at the boundary or everywhere)?

chengzhuzhang:
> Is the 0.1 K difference a difference in spatial mean or a typical difference (at the boundary or everywhere)?

My test with FGOAL and CNRM models, show that the difference only at the boundary. 

Using xcdat, I did spatial averaging comparing regridded datasets and original datasets, the global mean difference is around 0.01 K for FGOAL, and 0.002 K for CNRM.

taylor13:
regarding https://github.com/xCDAT/xcdat/pull/164#issuecomment-1146215206 , if I correctly understand the suggestion, a user would specify the "stopping " coordinate value.  I'm worried that this might be confused by some users who in the example considered (a 2.5 deg grid starting at -87.75), might set the "stopping" latitude as 90.  Then when you add delta/2 to this you get 91.25, and if the code wants to know whether to stop or not it takes the last latitude it created (88.75) and adds 2.5 to get 91.25, which depending on truncation errors would say to stop or not to stop.  We clearly want the code to stop at 87.75, not 91.25.

Wouldn't adding a smaller increment to "stop" be better (e.g., delta/3 or delta/10)?

Also, is the "stopping" value used for any other purpose (e.g., to create the bounds on the last grid cell)?  If so, we'd have to be a bit more clever.

taylor13:
I'd say the global mean differences given in https://github.com/xCDAT/xcdat/pull/164#issuecomment-1146464572 is too large to ignore.  I'll try to look at this in more depth over the next few days.

chengzhuzhang:
> @chengzhuzhang @pochedls I dug around the CDAT code to figure out what `norm_type` was being used. It looks like CDAT never sets a specific value so it defaults to `DSTAREA` (https://earthsystemmodeling.org/esmpy_doc/release/latest/html/regrid.html#ESMF.api.regrid.Regrid)


Thank you for looking.. in this case, I'm wondering how CDAT is handling normalization internally..

chengzhuzhang:
> > Is the 0.1 K difference a difference in spatial mean or a typical difference (at the boundary or everywhere)?
> 
> My test with FGOAL and CNRM models, show that the difference only at the boundary.
> 
> Using xcdat, I did spatial averaging comparing regridded datasets and original datasets, the global mean difference is around 0.01 K for FGOAL, and 0.002 K for CNRM.

Adding test results for the same two models, using cdat. the global mean difference is around 0.01K for FGOAL, and 0.001 K for CNRM. In these two cases, the conservation in xcdat is very close for FGOAL, but worse for CNRM.

xcdat/regridder/accessor.py
```diff
@@ -0,0 +1,284 @@
...
+            raise KeyError(
+                f"{standard_name} axis could not be correctly identified in the Dataset"
+            )
+
+        try:
+            axis_bnds = self._ds.bounds.get_bounds(axis.name)
+        except KeyError:
+            axis_bnds = None
+
+        return axis, axis_bnds
```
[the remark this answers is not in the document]
jasonb5: This should be resolved in the latest.

xcdat/regridder/grid.py
```diff
@@ -0,0 +1,530 @@
...
+            "lat": lat_axis,
+            "lon": lon_axis,
+        }
+    )
+
+    grid = grid.bounds.add_missing_bounds()
+
+    grid["lat_bnds"] = lat_bounds
+
+    return grid
```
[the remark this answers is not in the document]
jasonb5: This should be resolved in the latest.

xcdat/regridder/accessor.py
```diff
@@ -0,0 +1,284 @@
...
+    """
+    An accessor class that provides regridding attributes and methods on xarray
+    Datasets through the ``.regridder`` attribute.
+
+    Examples
+    --------
+
+    Import RegridderAccessor class:
+
+    >>> import xcdat  # or from xcdat.regridder import accessor
```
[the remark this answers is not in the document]
jasonb5: This should be resolved in the latest.

xcdat/regridder/grid.py
```diff
@@ -0,0 +1,530 @@
...
+        Bessel zero used to calculate the third legendre polynomial.
+    nlats : int
+        Number of lats.
+
+    Returns
+    -------
+    Tuple[float, float, float]
+        First, second and third legendre polynomial.
+    """
+    # TODO figure out why this is different compared to numpy.polynomial.legendre, is this wrong?
```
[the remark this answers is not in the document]
jasonb5: This is resolved in the latest.

xcdat/regridder/grid.py
```diff
@@ -0,0 +1,530 @@
...
+        weights[mid + 1] = 0.0
+
+        mid_weight = 2.0 / np.power(nlats, 2)
+
+        for x in range(2, nlats + 1, 2):
+            mid_weight = (mid_weight * np.power(nlats, 2)) / np.power(x - 1, 2)
+
+            weights[mid + 1] = mid_weight
+
+    # bounds = (180.0 / np.pi) * np.arcsin(bessel_zeros)
```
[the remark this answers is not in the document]
jasonb5: This is resolved in the latest.

xcdat/regridder/grid.py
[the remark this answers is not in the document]
jasonb5: This is resolved in the latest.

xcdat/regridder/grid.py
```diff
@@ -0,0 +1,530 @@
...
+    140.5871603528,
+    143.7287335737,
+    146.8703076258,
+    150.011882457,
+    153.1534580192,
+    156.2950342685,
+]
+
+EPS = 1e-14
+ESTIMATE_CONST = 0.25 * (1.0 - np.power(2.0 / np.pi, 2))
```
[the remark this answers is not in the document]
jasonb5: This is resolved in the latest.

xcdat/regridder/regrid2.py
```diff
@@ -0,0 +1,547 @@
...
+
+        output_da = xr.DataArray(
+            output_data,
+            dims=[axis_variable_name_map[x] for x in ordered_axis_names],
+            coords=coords,
+        )
+
+        data_vars[data_var] = output_da
+
+        return xr.Dataset(data_vars)
```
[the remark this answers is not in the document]
jasonb5: This is resolved in the latest.

xcdat/regridder/regrid2.py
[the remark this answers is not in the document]
jasonb5: This is resolved in the latest.

xcdat/regridder/regrid2.py
jasonb5: This is resolved in the latest.

xcdat/regridder/regrid2.py
jasonb5: This is resolved in the latest.

xcdat/regridder/regrid2.py
jasonb5: This is resolved in the latest.

conda-env/dev.yml
tomvothecoder: @jasonb5 Thanks, we can wait until we do the next release to add all of the new dependencies to the recipe (https://github.com/conda-forge/xcdat-feedstock/issues/2)

pochedls:
@chengzhuzhang asked me to share some code to loop over and test CMIP6 models to see whether they get the same global average values before and after regridding (I used `np.allclose()` as a test). 

I posted that code [here](https://gist.github.com/pochedls/9f60baea92822b1f0910cefc919cde13), which requires that you import [this script](https://github.com/xCDAT/xcdat_test/blob/main/scripts/cdw.py).

@jasonb5: I found one bug (after conservative regridding, the lat_bnds and lon_bnds are all NaN). 

The results of the script are:

> Results:                                                                                                                                               
>    xcdat was close in 51 of 62 samples
>    cdat was close in 53 of 62 samples
>    xcdat had better max differences in 32 of 62 samples
> 
> Models that did not match: AWI-ESM-1-1-LR
>                            CESM2-FV2
>                            CESM2-WACCM
>                            CESM2-WACCM-FV2
>                            FGOALS-f3-L
>                            IPSL-CM5A2-INCA
>                            IPSL-CM6A-LR
>                            IPSL-CM6A-LR-INCA
>                            MCM-UA-1-0
>                            MPI-ESM-1-2-HAM
>                            MPI-ESM1-2-LR

chengzhuzhang:
Thank you, Steve! @pochedls

taylor13:
I don't think horizontal interpolation needs to know anything about any bounds on the vertical coordinate.  The weights (at least in the case of any vertically independent mask applied) should not depend on the vertical coordinate; the same weights would be used for all levels in the vertical.  

It's possible that a set of 3-d weights could be produced (multiplying the horizontal weights by the vertical thickness of the layer).  There are a few ways one might do this, and I wouldn't attempt to generalize the weights-generator to handle these.  I think instead the horizontal regridder should only create a set of totally unmasked 2-d weights and _not_ account for layer thickness.  If there is masking (or fractional weighting or layer thickness weighting) that should be applied, it should be done in the matrix multiplication step (weights x source x fractional masking values = destination values).  This gives users flexible control on _how_ this is done.

What is the current approach if different from the above?

taylor13:
@pochedls, Regarding https://github.com/xCDAT/xcdat/pull/164#issuecomment-1149477948 , 

1. Do you suspect there is a problem with regrid2 or with the bounds passed to it?  
2. What is the difference between regridder.horizontal (xcdat_rmethod) and regrid_cdat (cdat_rmethod)?  Is either one based on regrid2?
3. Are you sure that in both cases the weights used by the "averager" are the same as the weights used by these two regridders?  
4. How close is "close" relative to say the max abs value of the field (or the mean of the field's absolute values)?  Can you be "not close" and still attribute difference to truncation errors?

Perhaps we could chat if that would be more efficient.

pochedls:
@taylor13 - to your [first comment](https://github.com/xCDAT/xcdat/pull/164#issuecomment-1150290388) - the horizontal regridder doesn't use vertical information. I believe this issue is arising in trying to make sure that the metadata stays with the dataset after it is returned from the regridding routine (or something like this). So I wouldn't worry that issue. 

Regarding the second comment

> @pochedls, Regarding [#164 (comment)](https://github.com/xCDAT/xcdat/pull/164#issuecomment-1149477948) ,
> 
>     1. Do you suspect there is a problem with regrid2 or with the bounds passed to it?
 
This is unlikely an issue with either (see above). 

>     2. What is the difference between regridder.horizontal (xcdat_rmethod) and regrid_cdat (cdat_rmethod)?  Is either one based on regrid2?

`regridder.horizontal` is the implementation of regridding in xcdat. `regrid_cdat` is a little Python function I made (it is at the top of the gist file) to wrap the CDAT regridding functionality. In both, you can specify regrid2 as a tool to do regridding. 

>     3. Are you sure that in both cases the weights used by the "averager" are the same as the weights used by these two regridders?

I didn't check this specifically, but I'm not comparing CDAT and xcdat. I'm checking to see if each yields the same global average values before and after conservative regridding. 

>     4. How close is "close" relative to say the max abs value of the field (or the mean of the field's absolute values)?  Can you be "not close" and still attribute difference to truncation errors?

I use [this routine](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html) to determine "closeness." I assume that the differences are **not** truncation errors, but I could be wrong about that. 

> Perhaps we could chat if that would be more efficient.

Happy to chat – just throw a WebEx on my calendar for a time that is convenient.

taylor13:
@pochedls , would it be difficult to post the table of individual model discrepancies printed by line 150 of your code?  Might give us so more hints.

Also, am I right that different versions of the ESMF regridder are being used (not regrid2) in your test?

taylor13:
From your summary, @pochedls, if I understand it correctly, it looks like on average across models, xcdat and cdat equally-well (or equally-poorly) preserve the global mean. (xcdat is better or equally good in half the cases: in 32 out of 64 cases.)  Right?  The reason I wanted to see the table was to check how big xcdat_max_abs_diff and cdat_max_abs_diff are in each case.

jasonb5:
@pochedls
> Also, I think we still need a try/except for regrid2. I still get this error: KeyError: 'Z bounds were not found, they must be added.'.

I've fixed this, it was specific to regrid2, forgot to remove some old code which masked the new code.
> I found one bug (after conservative regridding, the lat_bnds and lon_bnds are all NaN).

I couldn't reproduce this, was there a specific file it occurred with? If not it could have been fixed by my change to preserving bounds from the output grid.

jasonb5:
Analysis from meeting on 6/8

# From https://github.com/xCDAT/xcdat/pull/164#issuecomment-1131227290
Test 1:
> Model that disagreed was ICON-ESM-LR (see note on unstructured grid under take-aways)

xESMF doesn't support unstructured grids yet (https://github.com/pangeo-data/xESMF/issues/115.)

Test 2:
> For CNRM-CM6-1 (and likely the other CNRM models), it seems like there may be an issue at the poles (I suspect xcdat may not be correct)

"conservative_normed" method fixes issue. Still investigating difference between CDAT's and xESMF's use of ESMPy when method is conservative.

Test 3:
> Many models had errors in xcdat (e.g., ValueError: ESMC_FieldRegridStore failed with rc = 506. Please check the log files (named "*ESMF_LogFile").): 

This requires `ignore_degenerate=True`. This is now the default in xCDAT, same as CDAT.

Test 4:
> Same as `Test 3`.

# From https://github.com/xCDAT/xcdat/pull/164#issuecomment-1132027459
> TypeError: buffer is too small for requested array OR MemoryError: Unable to allocate 413. GiB for an array with shape (235403, 235403) and data type float64

Should open file chunked e.g. `xcdat.open_mfdataset(..., chunks={"time": 10})`.
> ValueError: Tool "Receive multiple variables for key 'longitude': {'lon', 'longitude'}. Expected only one. Please pass a list ['longitude'] instead to get all variables matching 'longitude'." does not exist, valid choices ['xesmf', 'regrid2']

Need to drop latitude/longitude, I've improved the error.
> ValueError: Tool 'lon_b' does not exist, valid choices ['xesmf', 'regrid2'] 

Not sure why xESMF can't infer `lat_b` and `lon_b` with these files. Still investigating and will open new issue.
> ValueError: ESMC_FieldRegridStore failed with rc = 506. Please check the log files (named "*ESMF_LogFile"). 

See `ignore_degenerate=True` explanation above.
> UFuncTypeError: ufunc 'subtract' cannot use operands with types dtype('O') and dtype('<m8[ns]')

Still investigating, will open issue.
> AttributeError: cf_xarray can't wrap attribute 'dims' because there are multiple values for 'X'. There is no unique mapping from 'X' to a value in 'dims'.

There are multiple coords that are possible 'X' dimension candidates, need to drop them.
> MergeError: conflicting values for variable 'vertices_longitude' on objects to be combined. You can skip this check by specifying compat='override'.

Conflicting variable value between multiple files, solved with `compat='override'`.

# From https://github.com/xCDAT/xcdat/pull/164#issuecomment-1132279922
Case 1:
Could not reproduce with latest.

Case 2:
I see the issue at the prime meridian with "conservative" but it goes away with "conservative_normed", still investigating.

pochedls:
@jasonb5 - both the regrid2 key error and the bound-NaN-values issues appear to be resolved on the latest commit.

taylor13:
I don't think the ESMF "conservative" option can generally preserve the _true_ global mean.  It preserves a global mean that it has been calculated using grid-cell areas for both source and target cells that ESMF has estimated (assuming cell corners are connected by great circles).  I think that for latxlon grids, the cdat averager constructs the cells assuming that the cell edges follow lines of longitude and latitude.  This true mean will be different from what ESMF thinks the mean is and will **_not_** be preserved. 
If I'm right, then the apparent differences in mean found in https://github.com/xCDAT/xcdat/pull/164#issuecomment-1149477948 should be expected, although it isn't clear why changing from cdat to xcdat should have an impact unless the ESMF regridder has been modified.

chengzhuzhang:
Hey @taylor13 Thank you for looking into this thread and the discussion. This PR discussion has became too long and hard to navigate (many conversations/comments are hidden). Based on the team meeting discussion, and comment from @jasonb5 https://github.com/xCDAT/xcdat/pull/164#issuecomment-1151693095, we will plan to merge this PR once the latest codes are tested, and to file separate PRs to  address individual problems. 

The global mean conservation test used both averager and regridder. It's hard to attribute the difference within this test. However, the test results of the averager has been done thoroughly and documented here: https://github.com/XCDAT/xcdat/issues/166#issuecomment-1082507179. One thing, we tried to find out is the use of normalization in CDAT's regridder (ESMF). We get closer results compare to CDAT if we use FRACAREA(Fraction area normalization) in ESMF than the DSTAREA(Destination area normalization). However, it is not clear to us if CDAT's use for ESMF. It seems like CDAT use DSTAREA as default, but re-normalize it afterwords. I'm wondering if you were involved in this development in CDAT and perhaps shed some lights? Thank you!

tomvothecoder:
Any updates on if this PR is ready to be merged @jasonb5?

tomvothecoder:
Great work @jasonb5!

pochedls:
Should this be deleted now?

taylor13:
I think the content of my https://github.com/xCDAT/xcdat/pull/164#issuecomment-1152464959 should be preserved and possibly included in documentation of the regridder.  Some work would first be required to check that the statements made there are in fact correct.  (I think they are, but I'm  not absolutely positive, since I have not seen results of a definitive test.)

pochedls:
@taylor13 - my mistake - this pull request (and your comment) is preserved, but is not showing up as active (I thought it was an active PR because of a silly mistake)...

PR #222, opened 2022-04-14
Remove `XCDATAccessor` class

### LINKED ISSUE #218: [FEATURE]: Consider removing the `XCDATAccessor` class to avoid redundancy

I am considering removing the `XCDATAccessor` because:

1. `XCDATAccessor` class is redundant and creates unnecessary coupling since it just copies and references methods from other accessor classes. 
2. It involves managing more code, writing tests, documentation, etc. We want this library to be lightweight to use and develop.
3. It is more modular to keep similar methods and functionality under their own accessor namespace, rather than adding them all under the same accessor class.
4. Providing too much flexibility with ways of using the library increases its complexity and might confuse users.
5. We don't expect to have _that_ many public APIs like CDAT, so the user doesn't really have to do much work in memorizing which accessor classes to use for their desired functions (especially with clear documentation).
   - API Documentation: https://xcdat.readthedocs.io/en/latest/api.html
6. `XCDATAccessor` method names must align with their source method. For example, `temporal.average()` is cleaner but becomes ambiguous as `xcdat.average()`, so we'd have to call it `xcdat.temporal_avg()`. However, `temporal.average()` is preferable.

- Remove `XCDATAccessor` class and its related tests and documentation.
- Rename `SpatialAccessor.spatial_avg()` to `SpatialAccessor.average()`.

### Additional context
There is less risk in making drastic, breaking API changes in the early stages of a library since the size of the user base is minimal.

https://github.com/XCDAT/xcdat/blob/c01b7d298a181e4385ec3aba3d265007a172cafe/xcdat/xcdat.py#L12-L95

tomvothecoder:
## Description

- Closes #218 

## Checklist

If applicable:


tomvothecoder:
Hey @pochedls, I rolled out this update pretty quickly. 

Not really much to review, just deleted some files, update the API docs main page, and renamed `.spatial_avg()` to `.spatial()`.  Good to see if you have any thoughts though.

pochedls (approved):

docs/api.rst
```diff
@@ -20,8 +22,20 @@ Top-level API
 .. currentmodule:: xarray
 
 
-Dataset
-------------------------------
+``xarray.Dataset`` Accessors
+----------------------------
+
+``xcdat`` accessors provide implicit namespaces for custom functionality that clearly identifies it as separate from built-in xarray methods.
+These accessors operate directly on  xarray objects, such as ``xarray.Dataset``.
```
tomvothecoder, line 29:
```suggestion
These accessors operate directly on ``xarray.Dataset`` objects.
```

Issue #166, opened 2021-11-19
Investigate CDAT versus xcdat spatial average differences


pochedls:
In order to continue to validate the spatial averaging code in `xcdat`, I looped over CMIP5/6 tas historical files (110 models) and 20+ domains [here](https://github.com/PCMDI/pcmdi_metrics/blob/27903488adefe1a2941100dfb56dceceb0321310/share/default_regions.py). 

227 (227/2486 = 9%) combinations had a CDAT-xcdat difference larger than 10<sup>-6</sup> [here](https://ghostbin.com/Ft18H/raw). 110 were specific to the NAFM region. This drops to 140 (and 110 for NAFM) if I look at differences larger than 10<sup>-3<sup>.

Two models produced errors for xcdat (but not CDAT): NorCPM1 and MCM-UA-1-0. In the case of NorCPM, it is related to Issue #162. In the case of MCM-UA-1-0, the problem is likely related to #163. There were no cases where CDAT throws an error and xcdat does not. Both packages produce an error for ICON-ESM-LR, because tas has a shape of [1980, 20480] (non-rectilinear grid).

This ticket is intended to document my findings in investigating these differences.

pochedls:
@lee1043 - In the metrics package for NAFM - do you load data differently when computing metrics for NAFM (e.g., tas = f('tas', longitude=(310, 309)) or something like this)? The longitude domain goes from 310<sup>o</sup>E to 60<sup>o</sup>E (wraps around the prime meridian). I suspect when I am calling CDAT it is averaging between 60<sup>o</sup>E and 310<sup>o</sup>E (rather than wrapping around the prime meridian.

lee1043:
@pochedls I think you've just found a bug in the PMP's Monsoon Wang metric's domain definition. I haven't had a chance to work with this specific metric so was not noticing the issue. In the PMP's driver for the Monsoon Wang metric, it just opens model file with regular cdms2 open, as [`f = cdms2.open(modelFile)`](https://github.com/PCMDI/pcmdi_metrics/blob/9d036a6b48296e27e08636cb918799f6fbe0cf06/pcmdi_metrics/monsoon_wang/monsoon_wang_driver.py#L189)

As you suspected, the extracted subdomain for NAFM (North African Monsoon) was for 60E to 310E with mirrored, instead of 310E to 60E. 

[...]

PR #226, opened 2022-04-25
Update spatial `axis` arg supported type and keys

### LINKED ISSUE #225: [FEATURE]: Update supported spatial `axis` arg keys  generic format

The valid `axis` arg values of `"lat"` and `"lon"` do not follow the axis naming convention that we adopted with our APIs.

https://github.com/XCDAT/xcdat/blob/79c488ea890febc422fe2e38b85c4e9dc7c72565/xcdat/spatial.py#L47

We implemented a mapping system for axis and coordinates names to their generic axis name.
https://github.com/XCDAT/xcdat/blob/79c488ea890febc422fe2e38b85c4e9dc7c72565/xcdat/axis.py#L14-L25

1. Update valid `axis` arg values from `"lat"` and `"lon"` to `"Y"`/`"y"` and `"X"`/`"x"` 
3. Convert the `axis` arg value to `.upper()` 
4. Map the `axis` arg to the dimension/coordinates in the `xr.Dataset`
5. Update `KeyError` if the CF `axis` attribute is not set for the X and Y axis coordinates
  * Related to https://github.com/XCDAT/xcdat/issues/166#issuecomment-1099382979
    >     We can probably improve the KeyError so that it says something like:
    >     KeyError: "Could not find an X and/or Y axis for spatial averaging. Make sure the data variable 'tas' has X and Y axis  coordinates and the 'axis' attribute is set for both."
  * A reusable convenience function/method might be useful for checking if the `axis` attribute is set for the desired axis in the Dataset. If it isn't, then raise a `KeyError`. `cf_xarray` already throws an error if we try something like `ds.cf["X"]` when the `"X"` axis attr is not set, but the error is ambiguous.

### Additional context

* `cdutil.averager()`'s `axis` arg accepts lowercase generic axis names (`"x"`, `"y"`, `"t"`, etc.)

tomvothecoder:
## Description

- Closes #225 

Summary of changes
- Update "lat" and "lon" references to "Y" and "X"
  - I decided to not allow "y" and "x" because they aren't CF-compliant (even if we could do `.upper()` to convert them).
  - https://github.com/XCDAT/xcdat/blob/79c488ea890febc422fe2e38b85c4e9dc7c72565/xcdat/axis.py#L9-L25
- Remove `_get_generic_axis_keys()` since it is no longer needed
- Add examples to `SpatialAccessor.average()` docstring
- Update `TemporalAccessor.__init__` check for 'T' axis 
- Add `TemporalAccessor.average()` docstring examples

## Checklist

If applicable:

### DIFF
--- a/xcdat/spatial.py
+++ b/xcdat/spatial.py
@@ -19,7 +19,6 @@ from dask.array.core import Array
 
 from xcdat.axis import (
     GENERIC_AXIS_MAP,
-    GenericAxis,
     _align_lon_bounds_to_360,
     _get_prime_meridian_index,
 )
@@ -28,7 +27,7 @@ from xcdat.dataset import get_data_var
 #: Type alias for a dictionary of axis keys mapped to their bounds.
 AxisWeights = Dict[Hashable, xr.DataArray]
 #: Type alias for supported spatial axis keys.
-SpatialAxis = Literal["lat", "lon"]
+SpatialAxis = Literal["X", "Y"]
 SPATIAL_AXES: Tuple[SpatialAxis, ...] = get_args(SpatialAxis)
 #: Type alias for a tuple of floats/ints for the regional selection bounds.
 RegionAxisBounds = Tuple[float, float]
@@ -44,7 +43,7 @@ class SpatialAccessor:
     def average(
         self,
         data_var: str,
-        axis: Union[List[SpatialAxis], SpatialAxis] = ["lat", "lon"],
+        axis: List[SpatialAxis] = ["X", "Y"],
         weights: Union[Literal["generate"], xr.DataArray] = "generate",
         lat_bounds: Optional[RegionAxisBounds] = None,
         lon_bounds: Optional[RegionAxisBounds] = None,
@@ -67,9 +66,9 @@ class SpatialAccessor:
         data_var: str
             The name of the data variable inside the dataset to spatially
             average.
-        axis : Union[List[SpatialAxis], SpatialAxis]
-            List of axis dimensions or single axis dimension to average over.
-            For example, ["lat", "lon"]  or "lat", by default ["lat", "lon"].
+        axis : List[SpatialAxis]
+            List of axis dimensions to average over, by default ["X", "Y"].
+            Valid axis strings include "X" and "Y".
         weights : Union[Literal["generate"], xr.DataArray], optional
             If "generate", then weights are generated. Otherwise, pass a
             DataArray containing the regional weights used for weighted
@@ -104,30 +103,36 @@ class SpatialAccessor:
 
         >>> import xcdat
 
-        Open a dataset and limit to a single variable:
-
-        >>> ds = xcdat.open_dataset("path/to/file.nc", var="tas")
-
         Call spatial averaging method:
 
-        >>> # First option
         >>> ds.spatial.average(...)
-        >>> # Second option
-        >>> ds.xcdat.average(...)
+
+        Check the `axis` attribute is set on the desired coordinates:
+
+        >>> ds.lat.attrs["axis"]
+        >>> Y
+        >>>
+        >>> ds.lon.attrs["axis"]
+        >>> X
+
+        Set the `axis` attribute on desired coordinates (if it isn't set):
+
+        >>> ds.lat.attrs["axis"] = "Y"
+        >>> ds.lon.attrs["axis"] = "X"
 
         Get global average time series:
 
-        >>> ts_global = ds.spatial.average("tas", axis=["lat", "lon"])["tas"]
+        >>> ts_global = ds.spatial.average("tas", axis=["X", "Y"])["tas"]
 
         Get time series in Nino 3.4 domain:
 
-        >>> ts_n34 = ds.spatial.average("ts", axis=["lat", "lon"],
+        >>> ts_n34 = ds.spatial.average("ts", axis=["X", "Y"],
         >>>     lat_bounds=(-5, 5),
         >>>     lon_bounds=(-170, -120))["ts"]
 
         Get zonal mean time series:
 
-        >>> ts_zonal = ds.spatial.average("tas", axis=['lon'])["tas"]
+        >>> ts_zonal = ds.spatial.average("tas", axis=["X"])["tas"]
 
         Using custom weights for averaging:
 
@@ -138,18 +143,18 @@ class SpatialAccessor:
         >>>     dims=["lat", "lon"],
         >>> )
         >>>
-        >>> ts_global = ds.spatial.average("tas", axis=["lat","lon"],
+        >>> ts_global = ds.spatial.average("tas", axis=["X", "Y"],
         >>>     weights=weights)["tas"]
         """
         dataset = self._dataset.copy()
         dv = get_data_var(dataset, data_var)
-        axis = self._validate_axis(dv, axis)
+        axis = self._validate_axis_arg(dv, axis)
 
         if isinstance(weights, str) and weights == "generate":
             if lat_bounds is not None:
-                self._validate_region_bounds("lat", lat_bounds)
+                self._validate_region_bounds("Y", lat_bounds)
             if lon_bounds is not None:
-                self._validate_region_bounds("lon", lon_bounds)
+                self._validate_region_bounds("X", lon_bounds)
             dv_weights = self._get_weights(axis, lat_bounds, lon_bounds)
         elif isinstance(weights, xr.DataArray):
             dv_weights = weights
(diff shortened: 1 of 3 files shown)


xcdat/spatial.py
```diff
@@ -104,30 +103,36 @@ def average(
...
+        >>> ds.lat.attrs["axis"]
+        >>> Y
+        >>>
+        >>> ds.lon.attrs["axis"]
+        >>> X
+
+        Set the `axis` attribute on desired coordinates (if it isn't set):
+
+        >>> ds.lat.attrs["axis"] = "Y"
+        >>> ds.lon.attrs["axis"] = "X"
```
tomvothecoder, line 121: These examples should help the user if they try to run spatial averaging on a Dataset with non-CF compliant axes and it breaks because the `axis` attributes aren't set.

xcdat/spatial.py
```diff
@@ -158,47 +163,50 @@ def average(
...
-                data_var.cf.axes[generic_axis_key]
+                data_var.cf.axes[generic_key]
             except KeyError:
                 raise KeyError(
-                    f"The data variable '{data_var.name}' is missing the '{axis}' "
-                    "dimension, which is required for spatial averaging."
+                    f"An '{generic_key}' axis dimension was not found in the data "
+                    f"variable. Make sure the data variable has '{generic_key}' axis "
+                    f"coordinates and its `axis` attribute is set to '{generic_key}'."
                 )
```
tomvothecoder, line 210: Related to https://github.com/XCDAT/xcdat/issues/166#issuecomment-1099382979

xcdat/spatial.py
```diff
@@ -655,14 +665,22 @@ def _validate_weights(
...
-            raise KeyError(f"Check weights DataArray includes {lon_key} dimension.")
+        # Check that the supplied weights include x and y dimensions.
+        x_key = data_var.cf.axes["X"][0]
+        y_key = data_var.cf.axes["Y"][0]
+
+        if "X" in axis and x_key not in weights.dims:
+            raise KeyError(
+                "The weights DataArray either does not include an 'X' axis, "
+                "or the axis's coordinates does not have the 'axis' attribute "
+                "set to 'X'"
```
tomvothecoder, line 676:
```suggestion
                "set to 'X'."
```

xcdat/spatial.py
```diff
@@ -655,14 +665,22 @@ def _validate_weights(
...
+            )
+        if "Y" in axis and y_key not in weights.dims:
+            raise KeyError(
+                "The weights DataArray either does not include an 'Y' axis, "
+                "or the axis's coordinates does not have the 'axis' attribute "
+                "set to 'Y'"
```
tomvothecoder, line 682:
```suggestion
                "set to 'Y'."
```

xcdat/spatial.py
```diff
@@ -655,14 +665,22 @@ def _validate_weights(
...
+        if "X" in axis and x_key not in weights.dims:
+            raise KeyError(
+                "The weights DataArray either does not include an 'X' axis, "
+                "or the axis's coordinates does not have the 'axis' attribute "
+                "set to 'X'"
+            )
+        if "Y" in axis and y_key not in weights.dims:
+            raise KeyError(
+                "The weights DataArray either does not include an 'Y' axis, "
+                "or the axis's coordinates does not have the 'axis' attribute "
```
tomvothecoder, line 681:
```suggestion
                "or the Y axis coordinates does not have the 'axis' attribute "
```

xcdat/spatial.py
```diff
@@ -655,14 +665,22 @@ def _validate_weights(
...
-        if "lon" in axis and lon_key not in weights.dims:
-            raise KeyError(f"Check weights DataArray includes {lon_key} dimension.")
+        # Check that the supplied weights include x and y dimensions.
+        x_key = data_var.cf.axes["X"][0]
+        y_key = data_var.cf.axes["Y"][0]
+
+        if "X" in axis and x_key not in weights.dims:
+            raise KeyError(
+                "The weights DataArray either does not include an 'X' axis, "
+                "or the axis's coordinates does not have the 'axis' attribute "
```
tomvothecoder, line 675:
```suggestion
                "or the X axis coordinates does not have the 'axis' attribute "
```

xcdat/spatial.py
```diff
@@ -655,14 +665,22 @@ def _validate_weights(
...
-            raise KeyError(f"Check weights DataArray includes {lat_key} dimension.")
-        if "lon" in axis and lon_key not in weights.dims:
-            raise KeyError(f"Check weights DataArray includes {lon_key} dimension.")
+        # Check that the supplied weights include x and y dimensions.
+        x_key = data_var.cf.axes["X"][0]
+        y_key = data_var.cf.axes["Y"][0]
+
+        if "X" in axis and x_key not in weights.dims:
+            raise KeyError(
+                "The weights DataArray either does not include an 'X' axis, "
```
tomvothecoder, line 674:
```suggestion
                "The weights DataArray either does not include an X axis, "
```

xcdat/spatial.py
```diff
@@ -655,14 +665,22 @@ def _validate_weights(
...
+                "or the axis's coordinates does not have the 'axis' attribute "
+                "set to 'X'"
+            )
+        if "Y" in axis and y_key not in weights.dims:
+            raise KeyError(
+                "The weights DataArray either does not include an 'Y' axis, "
```
tomvothecoder, line 680:
```suggestion
                "The weights DataArray either does not include an Y axis, "
```

xcdat/spatial.py
```diff
@@ -158,47 +163,50 @@ def average(
         dataset[dv.name] = self._averager(dv, axis, dv_weights)
         return dataset
 
-    def _validate_axis(
+    def _validate_axis_arg(
         self, data_var: xr.DataArray, axis: Union[List[SpatialAxis], SpatialAxis]
     ) -> List[SpatialAxis]:
-        """Validates if ``axis`` arg is supported and exists in the data var.
+        """Validates the ``axis`` dimension exists in the data variable.
```
pochedls, line 169: Should this be "Validation *that* the ``axis`` dimension exists..."?

xcdat/spatial.py
```diff
@@ -158,47 +163,50 @@ def average(
...
 
         Parameters
         ----------
         data_var : xr.DataArray
             The data variable.
         axis : Union[List[SpatialAxis], SpatialAxis]
-            List of axis dimensions or single axis dimension to average over.
+            List of axis dimensions or single axis dimensions.
```
pochedls, line 176: I like how you wrote this below: "List of axis dimensions or a single axis dimension"

pochedls (commented):
In CDAT, you specify the axes [as a single string](https://github.com/CDAT/genutil/blob/master/Lib/averager.py#L888), e.g., 'xy'. I think it would be worth including that as a  method for specifying the axes. I think this could be done simply by tweaking the `_validate_axis_arg`. Something like this (which could be combined with [this line](https://github.com/XCDAT/xcdat/pull/226/files#diff-bb3ba0d37e919e3f07121b387ed2fdcc08d25eaf68efb1b41f510bad44e72247R191)):

```
if isinstance(axis, str):
    # convert string axis argument to list of strings
    # representing the axes
    axis = [adim.upper() for adim in axis]
```

What do you think? Other than that potential change this looks great!

xcdat/spatial.py
tomvothecoder, line 176:
```suggestion
            List of axis dimensions or a single axis dimension.
```

tomvothecoder: I meant to use the bottom description, thanks for the catch

xcdat/spatial.py
tomvothecoder, line 169:
```suggestion
        """
        Validates that the ``axis`` dimension exists in the data variable.
```

tomvothecoder:
Thanks for the review!  

> In CDAT, you specify the axes [as a single string](https://github.com/CDAT/genutil/blob/master/Lib/averager.py#L888), e.g., 'xy'. I think it would be worth including that as a method for specifying the axes.

It is more Pythonic (and probably standard coding practice) for a function argument to accept a list of strings (`axis=["x", "y", "z", "t"]`) over a single string (`axis="xyzt"`), and to limit the accepted argument types as much as possible.

In this case, **each string should be treated as an independent instance**, rather than coupled as a single instance that has to be parsed. There is also less logic involved with accepting just a list of strings because type checking isn't necessary, the behavior of the API is more predictable, and it is more explicit and readable (in my opinion).

Posts I found on this subject:

----
https://stackoverflow.com/a/51737249
> **Parameters that can be either a thing or an iterable or things are a code smell**. It’s even worse when the thing is a string, because a string is an iterable, and even a sequence (so your test for isinstance(names, Iterable) would do the wrong thing).

This is what we're currently doing for a single axis (`axis="x"`).

> But if the best design really is a string or a (non-string) iterable of strings, or a string or a tuple of strings, the standard way to do that is type switching. It may look a bit ugly, but it calls attention to the fact that you’re doing exactly what you’re doing, and it does it in the most idiomatic way.
>
> def spam(names):
>     if isinstance(names, str):
>         names = [names]
>     dostuff(names)

I do agree with this person that it is ugly to have to perform this check for every function that has an argument that accepts a string or a list of strings.


https://stackoverflow.com/questions/998938/handle-either-a-list-or-single-integer-as-an-argument/998949#998949
>
> expect that the argument will always be a list - even if its just a list of one element.
> 
> > Remember:
> > 
> >     It is easier to ask for forgiveness than permission.
> > 

---

I suggest that we accept just a list of strings, but allow for upper and lower case axis names.

Valid `axis` arg values:
* `["x", "y", "t"]`
* `["X", "Y", "T"]`
* `["X"]`
* `["x"]`

Invalid `axis` arg values:
* `"xyt"`
* `"XYT"`
* `"X"`
* `"x"`

pochedls:
Thanks for the detailed reply.

xcdat/spatial.py
```diff
@@ -158,51 +166,39 @@ def average(
         dataset[dv.name] = self._averager(dv, axis, dv_weights)
         return dataset
 
-    def _validate_axis(
-        self, data_var: xr.DataArray, axis: Union[List[SpatialAxis], SpatialAxis]
-    ) -> List[SpatialAxis]:
-        """Validates if ``axis`` arg is supported and exists in the data var.
+    def _validate_axis_arg(self, axis: List[SpatialAxis]):
+        """
+        Validates that the ``axis`` dimension exists in the dataset.
```
tomvothecoder, line 171:
```suggestion
        Validates that the ``axis`` dimension(s) exists in the dataset.
```

PR #236, opened 2022-05-10
Add averages with time dimension removed

### LINKED ISSUE #201: [FEATURE]:  Add averages with time dimension removed
- It would be really handy, if this feature can include the capability to calculate weighted temporal average. Right now in `xarray`, one still has to generates the `weights` and then do something like `(d * weights).groupby("time.month").sum(dim="time")` I think `temporal_average` can cover this case straightforwardly? perhaps we can add a `mean` mode in addition or enable this if neither "climatology" nor "timeseries" is supplied?

_Originally posted by @chengzhuzhang in https://github.com/XCDAT/xcdat/issues/107#issuecomment-1069621197_

NA

### Additional context

NA

tomvothecoder:
## Description

- Closes #201 

This PR adds the ability to calculate averages weighted by time with the time dimension removed (via `ds.temporal.average()`). It simply wraps xarray's multi-line function calls into an API, with an example below for comparison.
```
# xarray
month_length = ds.time.dt.days_in_month
weights = (
    month_length.groupby("time.month") / month_length.groupby("time.month").sum()
)
pr_avg_xarray = ds.pr.weighted(weights).mean(dim="time")

#xcdat
pr_avg_xcdat = ds.temporal.average("pr",  freq="month")
```

Summary of Changes
- Add unit tests for `temporal.average()`
- Rename previous `.average()` to `.group_average()`
- Refactor many private methods in `class TemporalAccessor` 
  - Rename methods and variables, update docstrings
  - Refactor `._validate_weights()` for a significant performance increase of about 4-5x (for the monthly Dataset I was testing on)
- Remove redundant tests for private methods

Related API docs: https://xcdat.readthedocs.io/en/feature-201-temporal-mean/generated/xarray.Dataset.temporal.average.html#

Floating point closeness comparison with CDAT/cdutil: https://github.com/xCDAT/xcdat_test/blob/pr-236-validation/validation/v1.0.0/temporal_average/qa_cdat_closeness.ipynb

## Checklist

If applicable:


xcdat/temporal.py
```diff
@@ -224,7 +292,7 @@ def average(
         Returns
         -------
         xr.Dataset
-            Dataset containing the averaged data variable.
+            Dataset with the averaged data variable.
```
tomvothecoder, line 295:
```suggestion
            Dataset with the data variable averaged by time group.
```

xcdat/temporal.py
```diff
@@ -120,40 +133,95 @@ class TemporalAccessor:
...
         center_times: bool = False,
-        season_config: SeasonConfigInput = {
-            "dec_mode": "DJF",
-            "drop_incomplete_djf": False,
-            "custom_seasons": None,
-        },
+        season_config: SeasonConfigInput = DEFAULT_SEASON_CONFIG,
     ):
-        """Calculates the time series averages for a data variable.
+        """Returns group averages for a data variable.
```
tomvothecoder, line 224: Retu
```suggestion
        """Returns a Dataset with the data variable averaged by time group.
```

xcdat/temporal.py
```diff
@@ -120,40 +133,95 @@ class TemporalAccessor:
...
-        self._dataset: xr.Dataset = dataset
-
         # The weights for time coordinates, which are based on a chosen frequency.
         self._weights: Optional[xr.DataArray] = None
 
-    def average(
+    def average(self, data_var: str, freq: Frequency, center_times: bool = False):
+        """
+        Returns weighted average for a data variable with the time dimension
+        removed.
```
tomvothecoder, line 169:
```suggestion
        Returns a Dataset with weighted average applied to the data variable 
        with the time dimension removed.
```

xcdat/temporal.py
```diff
@@ -303,13 +358,9 @@ def climatology(
...
         center_times: bool = False,
-        season_config: SeasonConfigInput = {
-            "dec_mode": "DJF",
-            "drop_incomplete_djf": False,
-            "custom_seasons": None,
-        },
+        season_config: SeasonConfigInput = DEFAULT_SEASON_CONFIG,
     ):
-        """Calculates the climatology for a data variable.
+        """Returns the climatology for a data variable.
```
tomvothecoder, line 363:
```suggestion
        """Returns a Dataset with the climatology for the data variable.
```

xcdat/temporal.py
```diff
@@ -378,17 +429,14 @@ def climatology(
         Returns
         -------
         xr.Dataset
-            Dataset containing the averaged data variable.
+            Dataset with the climatology of a data variable.
```
tomvothecoder, line 432:
```suggestion
            Dataset with the climatology for a data variable.
```

xcdat/temporal.py
```diff
@@ -303,13 +358,9 @@ def climatology(
...
         center_times: bool = False,
-        season_config: SeasonConfigInput = {
-            "dec_mode": "DJF",
-            "drop_incomplete_djf": False,
-            "custom_seasons": None,
-        },
+        season_config: SeasonConfigInput = DEFAULT_SEASON_CONFIG,
     ):
-        """Calculates the climatology for a data variable.
+        """Returns the climatology for a data variable.
```
tomvothecoder, line 364:
```suggestion
    This method is similar to ``_group_average()``, but uses different time groups
    based on the frequency.

```

xcdat/temporal.py
```diff
@@ -695,15 +741,18 @@ def _temporal_avg(
...
-        # calculating the climatology of the data variable, so it is dropped
-        # and replaced.
-        ds = ds.drop_dims("time")
+        if self._mode == "average":
+            dv = self._average(dv)
+        else:
+            dv = self._group_average(dv)
+
+            # The original time dimension becomes obsolete after time averaging
+            # the data variable so it is dropped. Adding the time averaged
```
tomvothecoder, line 752:
```suggestion
            # The original time dimension is dropped from the Dataset because 
            # it becomes  after the data variable is averaged. A new time dimension
            # will be added to the Dataset when adding the averaged data variable.
```

xcdat/temporal.py
```diff
@@ -695,15 +741,18 @@ def _temporal_avg(
...
+            # data variable back to the dataset adds a new time dimension.
```
tomvothecoder, line 753:
```suggestion
```

xcdat/temporal.py
```diff
@@ -879,13 +928,36 @@ def _form_seasons(self, custom_seasons: List[List[str]]) -> Dict[str, List[str]]
 
         return c_seasons
 
-    def _averager(self, data_var: xr.DataArray) -> xr.DataArray:
-        """Averages a data variable by a grouping frequency.
+    def _average(self, data_var: xr.DataArray) -> xr.DataArray:
+        """Returns weighted averages for a data variable.
```
tomvothecoder, line 932:
```suggestion
    """"
    Returns the weighted averages for a data variable with the time dimension removed.
```

xcdat/temporal.py
```diff
@@ -879,13 +928,36 @@ def _form_seasons(self, custom_seasons: List[List[str]]) -> Dict[str, List[str]]
...
+
+        Parameters
+        ----------
+        data_var : xr.DataArray
+            The data variable.
+
+        Returns
+        -------
+        xr.DataArray
+            The averages of the data variable.
```
tomvothecoder, line 946:
```suggestion
        The weighted averages for a data variable with the time dimension removed.
```

xcdat/temporal.py
```diff
@@ -895,31 +967,32 @@ def _averager(self, data_var: xr.DataArray) -> xr.DataArray:
         Returns
         -------
         xr.DataArray
-            The averaged data variable.
+            The averages of the data variable by group.
```
tomvothecoder, line 970:
```suggestion
            The data variable averaged by time group.
```

xcdat/temporal.py
```diff
@@ -879,13 +928,36 @@ def _form_seasons(self, custom_seasons: List[List[str]]) -> Dict[str, List[str]]
...
+        dv = data_var.copy()
+
+        with xr.set_options(keep_attrs=True):
+            self._weights = self._get_weights()
+            dv = dv.weighted(self._weights).mean(dim=self._dim_name)
+
+        return dv
+
+    def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
+        """Returns group averages for a data variable.
```
tomvothecoder, line 957:
```suggestion
        """Returns the data variable averaged by time group.
```

xcdat/temporal.py
```diff
@@ -1047,6 +1120,15 @@ def _process_season_dataframe(self, df: pd.DataFrame) -> pd.DataFrame:
     def _convert_df_to_dt(self, df: pd.DataFrame) -> np.ndarray:
         """Converts a DataFrame of datetime components to datetime objects.
 
+        datetime objects require at least a year, month, and day value. However,
+        some modes and time frequencies don't require year, month, and/or day
+        for grouping. For these cases, use default values of 1 in order to
+        meet this requirement.
```
tomvothecoder, line 1126:
```suggestion
        meet this datetime requirement.
```

xcdat/temporal.py
```diff
@@ -1057,40 +1139,33 @@ def _convert_df_to_dt(self, df: pd.DataFrame) -> np.ndarray:
         np.ndarray
             A numpy ndarray of datetime.datetime or cftime.datetime objects.
 
-        Examples
+        References
         --------
```
tomvothecoder, line 1143:
```suggestion
        ----------
```

tomvothecoder (commented):
Initial review comments

chengzhuzhang:
Hey Tom, just to follow up the xcdat meeting that we have. Regarding to two points:
- new `average` method returns weighted means for a data variable with time dimension removed
-  potential API naming inconsistencies: differentiate ds.temporal.average() vs ds.temporal.average(): with the former for grouped average, later to collapse on time dimension. 

I'm thinking one possibility so resolve both: is to make the new method an option of average()/group_average(). 
i.e.:
 
if target frequency is not specified, the frequency should be the original data frequency (based on length of time interval), and returns means for the data variable with time dimension removed. if original data frequency is hourly or yearly, then the mean is weighted, otherwise, unweighted. 

example:
ds_month = ds.temporal.average("ts", freq=None, center_times=False) 
or 
ds_month = ds.temporal.average("ts", center_times=False)

tomvothecoder:
> I'm thinking one possibility to resolve both: is to make the new method an option of average()/group_average(). 
> i.e.:
>  
> Assume that `mean` only average based on original freqency: for instance, if target frequency is not specified, the frequency should be the original data frequency (based on length of time interval), and returns means for the data variable with time dimension removed. if original data frequency is monthly or yearly, then the mean is weighted, otherwise, unweighted. 

Hi Jill, thanks for the suggestion. On the surface it seems like a relatively simple solution. It would be nice to have a single API that can easily solve the naming inconsistency!

After investigating more closely with a prototype implementation, I unfortunately found that combining these two different averaging operations under a single API **makes the API less predictable because each operation produces a different output based on the `freq` argument.** 

For example, an end-user might expect a grouped average output, but they get a time collapsed average because they don't pass a `freq` argument (and vice versa). They might not understand why until they dig into the docs (which they should probably be reference first before using any API anyways haha). For the developer, we'd have to write more conditional logic based on the method arguments and more documentation.

---

My proposed (and current) solution is to have **two well-defined APIs that serve different purposes** because I think it is easier to use and can be cleanly implemented.

The name of the APIs/methods explains what they do, while their respective docstrings fills the gap in how to use them.

1. `.average()` -- returns averages for a data variable and collapses the time dimension
   -  Arguments include `data_var`, `weighted`, and `center_times`
   - I think we should include the `weighted` boolean arg so that the user explicitly declares whether or not output is weighted or not, rather than handling whether it is weighted internally based on the time frequency. It makes the output more predictable to the user because they know exactly what the output reflects. 
2. `.group_average()` -- returns grouped averages for a data variable
   - Arguments include `data_var`, `freq`, `weighted`, `center_times`, and `season_config`

---

As mentioned before, we will write an API translation table for CDAT users to help smooth the transition process to xCDAT.

Let me know what you think! Also tagging @lee1043 to see if he has any thoughts or ideas.

chengzhuzhang:
@tomvothecoder Thank you for the effort to explore the different options! Appreciate it. What you suggested about two APIs actually make sense to me. Tagging @lee1043 or @pochedls to see if they have some input on this implementation or perhaps better API names?

lee1043:
@tomvothecoder @chengzhuzhang thank you for the discussion. I was initially leaning toward what @chengzhuzhang suggested earlier -- single API on top of two capabilities -- but I am leaning back to the well-defined 2 separate APIs. I think the point @tomvothecoder raised is very important -- make API predictable, which I think is one of key to be a successful community tool. While we expect users do READ docs first, we know it often not the case. In my personal experiences, I often start directly from some sample of the code, and read docs when I get stumbled. When API's predictability is great and I know what exactly API does, I can trust the API more. I think `group_average` is fine, and also open to any other suggestions.

tests/test_temporal.py
```diff
@@ -1,8 +1,5 @@
-from datetime import datetime
```
tomvothecoder, line 1: I removed a lot of the unit tests for private methods. I learned that testing private methods is unnecessary and introduces coupling to the implementation details, which breaks encapsulation. It makes refactoring significantly more tedious because the tests for the private methods need to also be updated.

Instead, tests should revolve around public methods and their behaviors/outputs.

More info on why testing private methods is considered "bad" practice":
https://softwareengineering.stackexchange.com/a/380296
https://stackoverflow.com/questions/105007/should-i-test-private-methods-or-only-public-ones

tomvothecoder (commented):
Hi @lee1043 and @chengzhuzhang, this PR is ready for review.

The PR's description should contain all of the relevant information to make the review process smooth. Thanks!

tomvothecoder:
> Hi @lee1043 and @chengzhuzhang, this PR is ready for review.
> 
> The PR's description should contain all of the relevant information to make the review process smooth. Thanks!

@chengzhuzhang won't be able to review this PR this week. @lee1043 can you give this a shot whenever you get a chance? Thanks

lee1043:
I am testing this PR at https://github.com/xCDAT/xcdat_test/blob/test_temporal_avg/devel_test/temporal_average.ipynb. While I can give an overview at our regular meeting or we can have a separate chat, just sharing in advance for your interest. Basically what I am doing in the notebook is trying every option, even if that does not make sense, so I can learn how the function work and respond. 

While I think I am understanding `average()` which returns temporal averaged with time coordinate collapses to 1 time step, and I know the intention of having `freq` there which is for weighting adjustment, but I think there is a chance that user can confused it with group average be seeing the "freq" option. I wonder if the function automatically recognize temporal interval of the given dataset and its data array, and user only decide `weighted=True/False` without dealing with the `freq` parameter, and the function automatically apply proer `freq` if `weighted=True`.

In the notebook, I also tried some silly things, such as get temporal average of monthly time series with freq=day or hour (which does not make sense but just for curious what would return) and compare their differences in Section 1.6. theoretically they might have to return the same value as the original monthly data because they don't have higher frequency input, but It is interesting to see how different they are and there are some patterns in the differences. 

Regarding the group average, in the notebook I am using monthly time series as input (shape: (1872, Y, X)), so when I was doing `ds_gra_month = ds.temporal.group_average("ts", freq="month")` in Section 2.1, my logical sense was expecting a shape: (**12**, Y, X) (group same months together and get average of each group, like getting an annual cycle). 
For year and season, it seems right numbers were returned for the time dimension. 

So far I mostly checked number in time dimensions and yet to cross-validate returned values with CDAT. I will continue to work on that.

xcdat/temporal.py
```diff
@@ -120,40 +128,144 @@ class TemporalAccessor:
...
+
+        Get weighted averages for a monthly time series data variable:
+
+        >>> ds_month = ds.temporal.average("ts", freq="month", center_times=False)
+        >>> ds_month.ts
+        """
+        return self._averager(
+            data_var, "average", freq, True, center_times, season_config
+        )
+
```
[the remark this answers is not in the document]
lee1043: I wonder `freq` can be automatically assigned with a new parameter if `weighted=True`. Having `freq` seems can confuse with `group_average`.

tomvothecoder:
Thanks for reviewing @lee1043! You have a great validation process going on.

> While I think I am understanding average() which returns temporal averaged with time coordinate collapses to 1 time step, and I know the intention of having freq there which is for weighting adjustment, but I think there is a chance that user can confused it with group average be seeing the "freq" option. I wonder if the function automatically recognize temporal interval of the given dataset and its data array, and user only decide weighted=True/False without dealing with the freq parameter, and the function automatically apply proer freq if weighted=True.

I agree, and Jill mentioned a similar solution to what you outlined. 

I think it might confuse users if there is a `freq` argument for specifying the frequency used to calculate weights. Users might unintentionally pass a frequency that is not aligned with the actual frequency of the time intervals, resulting in incorrect weighting. 

**I will try to figure out how `cdutil` recognizes the time interval. If `cdutil`'s implementation isn't clear and/or robust, we can figure out our own implementation.**

> In the notebook, I also tried some silly things, such as get temporal average of monthly time series with freq=day or hour (which does not make sense but just for curious what would return) and compare their differences in Section 1.6. theoretically they might have to return the same value as the original monthly data because they don't have higher frequency input, but It is interesting to see how different they are and there are some patterns in the differences.

I noticed this too. This isn't necessarily an implementation bug, but if the user passes a frequency that isn't aligned with the actual frequency of the time intervals (e.g., `freq="day"` with monthly data), it can produce unintended odd behaviors.

For example, with the monthly time coordinates `['1850-01-16T12:00:00.000000000', '1850-02-15T00:00:00.000000000']`:
- If user passes `freq="day"`: calculate weights using groups day 16 and day 15
- If user passes `freq="hour"`: calculates weights using groups hour 12 and hour 0

This is possible for the user if they want to calculate weights like that, but it doesn't really make sense. **Removing the `freq` argument will eliminate the possibility of passing incorrect/unintended frequencies.**

> Regarding the group average, in the notebook I am using monthly time series as input (shape: (1872, Y, X)), so when I was doing ds_gra_month = ds.temporal.group_average("ts", freq="month") in Section 2.1, my logical sense was expecting a shape: (12, Y, X) (group same months together and get average of each group, like getting an annual cycle). However what returned was shape: (1872, Y, X), so same as the original one.

The grouping behaviors for group averages and climatologies are based on CDAT.

In the case of the monthly time series input, grouping on month will return the original data because `freq=month` groups on year and month. 

Here is a code snippet to validate this behavior against CDAT:
```
import cdms2
import cdutil
import numpy as np
import xarray as xr

import xcdat

filepath = "/home/vo13/XCDAT/xcdat/input/demo_data/CMIP5_demo_data/pr_Amon_ACCESS1-0_historical_r1i1p1_185001-200512.nc"

ds = xcdat.open_dataset(filepath)
ds_cdat = cdms2.openDataset(filepath)

pr_avg_cdat = cdutil.ANNUALCYCLE(ds_cdat["pr"])
# pr_avg_cdat.shape  == (1872, 145, 192)

pr_avg_xcdat = ds.temporal.group_average("pr", freq="month")["pr"]
# pr_avg_xcdat. shape == (1872, 145, 192)
```

The grouping behavior that produces (12, Y, X) is a result of **calculating the climatology, not the group average**:
```
pr_climo_cdat = cdutil.ANNUALCYCLE.climatology(ds_cdat["pr"])
pr_climo_cdat.shape == (12, 145, 192)

pr_climo_xcdat = ds.temporal.climatology(pr", freq="month")["pr"]
pr_climo_xcdat.shape == (12, 145, 192)
```

This is how grouping is done in xCDAT and CDAT. It was also validated in this notebook
https://github.com/xCDAT/xcdat_test/blob/master/validation/temporal_average/floating_point_comparisons.ipynb.
https://github.com/xCDAT/xcdat/blob/da2cf9e6ba32d6bc2e4ec81a69438e90b66b42db/xcdat/temporal.py#L33-L59

xcdat/temporal.py
```diff
@@ -120,40 +127,126 @@ class TemporalAccessor:
...
+        min_delta = pd.to_timedelta(np.diff(time_coords).min(), unit="ns")
+
+        if min_delta < pd.Timedelta(days=1):
+            return "hour"
+        elif min_delta >= pd.Timedelta(days=1) and min_delta < pd.Timedelta(days=28):
+            return "day"
+        elif min_delta >= pd.Timedelta(days=28) and min_delta < pd.Timedelta(days=365):
+            return "month"
+        else:
+            return "year"
```
tomvothecoder, line 239: @lee1043 I added this new method, `_infer_freq()` for inferring the `freq` of the coordinates. It will only be used if `weighed=True`.

Let me know how the algorithm looks, thanks.

xcdat/temporal.py
```diff
@@ -879,13 +939,37 @@ def _form_seasons(self, custom_seasons: List[List[str]]) -> Dict[str, List[str]]
...
+        with xr.set_options(keep_attrs=True):
+            if self._weighted:
+                self._weights = self._get_weights()
+                dv = dv.weighted(self._weights).mean(dim=self._dim_name)
+            else:
+                dv = dv.mean(dim=self._dim_name)
+
+        dv = self._add_operation_attrs(dv)
+
+        return dv
```
tomvothecoder, line 969: This is the main method that performs the work for `ds.temporal.average()`.

xcdat/temporal.py
lee1043: @tomvothecoder I think this is great and very straightforward, thank you for revising it. I am playing with the new function in the notebook now. 

P.S: (typo in above comment `weighed=True` : `weighed` --> `weighted`

lee1043:
@tomvothecoder thank you for the revision. I have tested the function in following notebooks using
- [monthly time series](https://github.com/xCDAT/xcdat_test/blob/36ca82fbf358655c0cedc15a137215de6c0df89b/devel_test/temporal_average_monthly.ipynb)
- [daily time series](https://github.com/xCDAT/xcdat_test/blob/36ca82fbf358655c0cedc15a137215de6c0df89b/devel_test/temporal_average_daily.ipynb)

And I think the function is working as expected!

tomvothecoder:
@lee1043 Great, thanks for adding those notebooks! I will do a final review and then merge.

chengzhuzhang:
@lee1043 thank you for reviewing and testing this PR, also great to get more validation notebooks for it! At some point, we can also do a similar test as Steve's, to loop over all CMIP models and compare results between cdat and xcdat.

Issue #242, opened 2022-05-26
[FEATURE]: Refactor test suite that covers private tests


tomvothecoder:
### Feature Request Criteria

- [X] Feature is not implemented by ``xarray``
- [X] Feature is not implemented by another actively developed xarray-based package
- [X] Feature is not limited to specific use cases (e.g., data quality issues)
- [X] Feature is generally reusable
- [X] Feature is relatively simple and lightweight to implement and use

### Is your feature request related to a problem?

The test suite includes tests for private methods and functions.

Per my comment here https://github.com/xCDAT/xcdat/pull/236#discussion_r875258760:
> I learned that testing private methods is unnecessary and introduces coupling to the implementation details, which breaks encapsulation. It makes refactoring significantly more tedious because the tests for the private methods need to also be updated (also risk of test duplication with public methods). Instead, tests should revolve around public methods and their behaviors/outputs.
> 
> More info on why testing private methods is considered "bad" practice:
> https://softwareengineering.stackexchange.com/a/380296
> https://stackoverflow.com/questions/105007/should-i-test-private-methods-or-only-public-ones
> 

### Describe the solution you'd like

Update tests for public APIs to cover tests for private APIs, and then remove the private API tests.

Issue #114, opened 2021-10-04
Returns weight information (axis or full matrix) after averaging operation


tomvothecoder:
### Is your feature request related to a problem? If yes, Please describe

Weights are not returned in the Dataset after spatial averaging.

### Describe your proposed solution

Attached the weights in the Dataset as the `var_wts` (e.g., `tas_wits`) after spatially averaging.

### Describe alternatives solutions you've considered

N/A

### Additional context

N/A

taylor13:
If I understand it correctly, I think this is a good idea.  The weights might be needed for subsequent operations.

durack1:
@tomvothecoder if this will be implemented through `xESMF` then there may be some other outputs that could also be grabbed, if I recall correctly, if `diags = {}` was defined as an empty dictionary and passed to the ESMPy call many diagnostics generated through the regridding process were returned, with the weights being one of numerous outputs (this is as close as I can find regarding docs for this in cdms2 https://cdms.readthedocs.io/en/latest/regrid2.html#regrid2.mvESMFRegrid.ESMFRegrid.fillInDiagnosticData)

pochedls:
@durack1 - we don't use xESMF for spatial averaging. 

I think this issue can probably be closed since the weights can be retrieved via `.get_weights()` [here](https://github.com/xCDAT/xcdat/blob/main/xcdat/spatial.py#L297). If a more complex implementation of this is needed in the future, that could probably be added as a feature request.

durack1:
@pochedls interesting, how consistent are the xESMF weights and the weights being generated through the separate spatial averaging routine?

pochedls:
> @pochedls interesting, how consistent are the xESMF weights and the weights being generated through the separate spatial averaging routine?

I have never used ESMF for spatial averaging – so I haven't attempted to compare the weights.

tomvothecoder:
> @durack1 - we don't use xESMF for spatial averaging.
> 
> I think this issue can probably be closed since the weights can be retrieved via .get_weights() [here](https://github.com/xCDAT/xcdat/blob/main/xcdat/spatial.py#L297). If a more complex implementation of this is needed in the future, that could probably be added as a feature request.

I had an inclination that the `._get_weights()` methods would be useful for users. I added a related issue to make those methods public (https://github.com/xCDAT/xcdat/issues/251).

Closing this based on the discussion.

PR #252, opened 2022-06-07
Update `_get_weights()` method for `SpatialAccessor` and `TemporalAccessor`

### LINKED ISSUE #251: [Feature]: Update `_get_weights()` method for `SpatialAccessor` and `TemporalAccessor`

These methods are useful to have publicly so the user can reuse weights for other operations, or to see what the generated values are for validation purposes.

Related to this comment:
>I think this issue can probably be closed since the weights can be retrieved via .get_weights() here. If a more complex implementation of this is needed in the future, that could probably be added as a feature request.
&mdash; @pochedls https://github.com/xCDAT/xcdat/issues/114#issuecomment-1146730677

- Rename `_get_weights()` to `get_weights()`
- Update references to the new name

_No response_

### Additional context

_No response_

tomvothecoder:
## Description

Summary of Changes
- Closes #251 
  - Related to #114
- Add default `None` value for `lat_bounds` and `lon_bounds` keyword args to `SpatialAccessor.get_weights()`
- Add kwarg `keep_weights` to both methods
- Make `_get_weights()` a public method for `SpatialAccessor`
- Delete redundant tests for private methods (related to #242)
- Delete faulty tests with domain bounds

## Checklist

If applicable:

### DIFF
--- a/tests/test_spatial.py
+++ b/tests/test_spatial.py
@@ -39,12 +39,184 @@ class TestAverage:
 
     def test_raises_error_if_data_var_not_in_dataset(self):
         with pytest.raises(KeyError):
-            self.ds.spatial.average(
-                "not_a_data_var",
-                axis=["Y", "incorrect_axis"],
-            )
+            self.ds.spatial.average("not_a_data_var", axis=["Y", "incorrect_axis"])
 
-    def test_spatial_average_for_lat_and_lon_region_using_custom_weights(self):
+    def test_raises_error_if_axis_list_contains_unsupported_axis(self):
+        with pytest.raises(ValueError):
+            self.ds.spatial.average("ts", axis=["Y", "incorrect_axis"])
+
+    def test_raises_error_if_lat_axis_does_not_exist(self):
+        ds = self.ds.copy()
+        ds.lat.attrs["axis"] = None
+        with pytest.raises(KeyError):
+            ds.spatial.average("ts", axis=["X", "Y"])
+
+    def test_raises_error_if_lon_axis_does_not_exist(self):
+        ds = self.ds.copy()
+        ds.lon.attrs["axis"] = None
+        with pytest.raises(KeyError):
+            ds.spatial.average("ts", axis=["X", "Y"])
+
+    def test_raises_error_if_bounds_type_is_not_a_tuple(self):
+        with pytest.raises(TypeError):
+            self.ds.spatial.average("ts", axis=["Y"], lat_bounds=[1, 1])
+
+        with pytest.raises(TypeError):
+            self.ds.spatial.average("ts", axis=["Y"], lat_bounds="str")
+
+    def test_raises_error_if_there_are_0_elements_in_the_bounds(self):
+        with pytest.raises(ValueError):
+            self.ds.spatial.average("ts", axis=["Y"], lat_bounds=())
+
+    def test_raises_error_if_there_are_more_than_two_elements_in_the_bounds(self):
+        with pytest.raises(ValueError):
+            self.ds.spatial.average("ts", axis=["Y"], lat_bounds=(1, 2, 3))
+
+    def test_raises_error_if_lower_bound_is_not_a_float_or_int(self):
+        with pytest.raises(TypeError):
+            self.ds.spatial.average("ts", axis=["Y"], lat_bounds=("invalid", 1))
+
+    def test_raises_error_if_upper_bound_is_not_a_float_or_int(self):
+        with pytest.raises(TypeError):
+            self.ds.spatial.average("ts", axis=["Y"], lat_bounds=(1, "invalid"))
+
+    def test_raises_error_if_lower_lat_bound_is_larger_than_upper(self):
+        with pytest.raises(ValueError):
+            self.ds.spatial.average("ts", axis=["Y"], lat_bounds=(2, 1))
+
+    def test_does_not_raise_error_if_lower_lon_bound_is_larger_than_upper(self):
+        self.ds.spatial.average("ts", axis=["X"], lon_bounds=(2, 1))
+
+    def test_raises_error_if_lat_axis_is_specified_but_lat_is_not_in_weights_dims(
+        self,
+    ):
+        weights = xr.DataArray(
+            data=np.ones(4), coords={"lon": self.ds.lon}, dims=["lon"]
+        )
+        with pytest.raises(KeyError):
+            self.ds.spatial.average("ts", axis=["X", "Y"], weights=weights)
+
+    def test_raises_error_if_lon_axis_is_specified_but_lon_is_not_in_weights_dims(
+        self,
+    ):
+        weights = xr.DataArray(
+            data=np.ones(4), coords={"lat": self.ds.lat}, dims=["lat"]
+        )
+        with pytest.raises(KeyError):
+            self.ds.spatial.average("ts", axis=["X", "Y"], weights=weights)
+
+    def test_raises_error_if_weights_lat_and_lon_dims_dont_align_with_data_var_dims(
+        self,
+    ):
+        # Get a slice of the dataset to reduce the size of the dimensions for
+        # simpler testing.
+        ds = self.ds.isel(lat=slice(0, 3), lon=slice(0, 3))
+        weights = xr.DataArray(
+            data=np.ones((3, 3)),
+            coords={"lat": ds.lat, "lon": ds.lon},
+            dims=["lat", "lon"],
+        )
+
+        with pytest.raises(ValueError):
+            self.ds.spatial.average("ts", axis=["X", "Y"], weights=weights)
+
+    def test_spatial_average_for_lat_region_and_keep_weights(self):
+        ds = self.ds.copy()
+
+        result = ds.spatial.average(
+            "ts", axis=["Y"], lat_bounds=(-5.0, 5), keep_weights=True
+        )
+
+        expected = self.ds.copy()
+        expected["ts"] = xr.DataArray(
+            data=np.array(
+                [[2.25, 2.25, 2.25, 2.25], [1.0, 1.0, 1.0, 1.0], [1.0, 1.0, 1.0, 1.0]]
+            ),
+            coords={"time": expected.time, "lon": expected.lon},
+            dims=["time", "lon"],
+        )
+        expected["lat_wts"] = xr.DataArray(
+            name="lat_wts",
+            data=np.array([0.0, 0.08715574, 0.08715574, 0.0]),
+            dims=["lat"],
+        )
+
+        xr.testing.assert_allclose(result, expected)
+
+    def test_spatial_average_for_lat_region(self):
+        ds = self.ds.copy()
+
+        # Specifying axis as a str instead of list of str.
+        result = ds.spatial.average("ts", axis=["Y"], lat_bounds=(-5.0, 5))
+
+        expected = self.ds.copy()
+        expected["ts"] = xr.DataArray(
+            data=np.array(
+                [[2.25, 2.25, 2.25, 2.25], [1.0, 1.0, 1.0, 1.0], [1.0, 1.0, 1.0, 1.0]]
+            ),
+            coords={"time": expected.time, "lon": expected.lon},
+            dims=["time", "lon"],
+        )
+
+        assert result.identical(expected)
+
+    @requires_dask
+    def test_spatial_average_for_lat_region_and_keep_weights_with_dask(self):
+        ds = self.ds.copy().chunk(2)
+
+        # Specifying axis as a str instead of list of str.
+        result = ds.spatial.average(
+            "ts", axis=["Y"], lat_bounds=(-5.0, 5), keep_weights=True
+        )
+
+        expected = self.ds.copy()
+        expected["ts"] = xr.DataArray(
+            data=np.array(
+                [[2.25, 2.25, 2.25, 2.25], [1.0, 1.0, 1.0, 1.0], [1.0, 1.0, 1.0, 1.0]]
+            ),
+            coords={"time": expected.time, "lon": expected.lon},
+            dims=["time", "lon"],
+        )
+        expected["lat_wts"] = xr.DataArray(
+            name="lat_wts",
+            data=np.array([0.0, 0.08715574, 0.08715574, 0.0]),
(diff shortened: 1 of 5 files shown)


xcdat/spatial.py
```diff
@@ -178,14 +178,98 @@ def average(
...
-            dv_weights = self._get_weights(axis, lat_bounds, lon_bounds)
+            dv_weights = self.get_weights(axis, lat_bounds, lon_bounds)
         elif isinstance(weights, xr.DataArray):
             dv_weights = weights
 
         self._validate_weights(dv, axis, dv_weights)
         dataset[dv.name] = self._averager(dv, axis, dv_weights)
         return dataset
 
+    def get_weights(
```
pochedls, line 189: I assume this is the only line that changed in this file? Github is highlighting this whole function...probably because it got moved?

tests/test_spatial.py
```diff
@@ -124,128 +210,171 @@ def test_chunked_spatial_average_for_lat_region(self):
...
-        with pytest.raises(TypeError):
-            self.ds.spatial._validate_region_bounds("X", (1, "invalid"))
+        xr.testing.assert_allclose(result, expected)
 
-    def test_raises_error_if_lower_lat_bound_is_bigger_than_upper(self):
-        with pytest.raises(ValueError):
-            self.ds.spatial._validate_region_bounds("Y", (2, 1))
+    def test_weights_for_region_in_lon_domain_with_domain_spanning_p_meridian(self):
+        # FIXME: ValueError when domain bounds contains lower bound larger than
+        # upper bound
```
pochedls, line 307: I think this should throw an error? I'm kind of confused why it wasn't before based on [this check](https://github.com/xCDAT/xcdat/blob/573c4ad361107f644184d2b9ae614d5b26d90d65/xcdat/spatial.py#L375). Should this test have region bounds such that `lon_bounds=np.array([350, 20])` or something like that? 

I think region bounds should be able to accept a larger right hand bound, but I don't think this should be true for domain bounds.

tests/test_spatial.py
```diff
@@ -766,115 +670,3 @@ def test_scales_lon_bounds_when_wrapping_around_prime_meridian(self):
...
-            coords={"lat": self.ds.lat, "lon": self.ds.lon},
-            dims=["lat", "lon"],
-        )
-
-        result = self.ds.spatial._averager(self.ds.ts, axis=["X", "Y"], weights=weights)
-        expected = xr.DataArray(
-            name="ts", data=np.ones(15), coords={"time": self.ds.time}, dims=["time"]
-        )
-
-        assert result.identical(expected)
```
pochedls, line 880: I kind of remember you commenting on this...is this all removed because they are private methods (and we don't need to test them)?

pochedls: Same for `TestGetWeights`?

pochedls (commented):
I have no objection to making `get_weights()` public, though I didn't fully understand the the changes to the unit tests. Should I view the changes in `tests/` as a reorganization or something that I should review more carefully? 

I think the error for `test_weights_for_region_in_lon_domain_with_both_spanning_p_meridian` is correct and we should represent the problematic line as `ds.lon_bnds.data[:] = np.array([[-1, 1], [1, 90], [90, 180], [180, 359]])`.

xcdat/spatial.py
tomvothecoder: Yup, the method was renamed, moved, and its references were updated. Nothing else was changed here.

tests/test_spatial.py
tomvothecoder: Exactly, I consolidated tests by moving some of the tests for private methods to their parent public method.

No need to review the tests (aside my two comments for the failing tests) because we still have 100% test cover which means the tests for public methods cover the old tests for private methods.

tests/test_spatial.py
tomvothecoder: Okay I understand now, domain bounds should not have a larger right hand value so it should throw an error.

The old test below was faulty in that it didn't throw an error for this case, since we skipped the `_validate_domain_bounds()` check by directly testing this case with `_get_longitude_bounds()` instead of `get_weights()`
https://github.com/xCDAT/xcdat/blob/223e73b60f19b2634a57fd0a5ce5105fda3fb1f6/tests/test_spatial.py#L494-L510

Basically, this test should be deleted.

tomvothecoder:
> I have no objection to making `get_weights()` public, though I didn't fully understand the the changes to the unit tests. Should I view the changes in `tests/` as a reorganization or something that I should review more carefully?

I addressed this comment in one of the PR review comments.

> 
> I think the error for `test_weights_for_region_in_lon_domain_with_both_spanning_p_meridian` is correct and we should represent the problematic line as `ds.lon_bnds.data[:] = np.array([[-1, 1], [1, 90], [90, 180], [180, 359]])`.

I ended up removing tests where the domain bounds has a larger left hand value than right hand value because **they are supposed to throw an error, rather than pass**.  There's already a test to check that an error is thrown for this case.

I kept `test_weights_for_region_in_lon_domain_with_region_spanning_p_meridian` since **regions can have a larger lower bound vs. upper bound**.

tests/test_spatial.py
```diff
@@ -124,128 +210,171 @@ def test_chunked_spatial_average_for_lat_region(self):
...
-    def setup(self):
-        self.ds = generate_dataset(cf_compliant=True, has_bounds=True)
-        self.weights = xr.DataArray(
-            data=np.ones((4, 4)),
-            coords={"lat": self.ds.lat, "lon": self.ds.lon},
-            dims=["lat", "lon"],
-        )
+    def test_weights_for_region_in_lon_domain_with_region_spanning_p_meridian(self):
+        # FIXME: ValueError when region bounds contains lower bound larger than
+        # upper bound
```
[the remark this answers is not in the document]
tomvothecoder: This should pass, so nothing to fix here

tests/test_spatial.py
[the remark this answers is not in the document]
tomvothecoder: This test is faulty, the domain should not have a larger lower bound than upper bound.  I ended up deleting this test.

tests/test_spatial.py
tomvothecoder: Correct, we should test the public method which will cover the private methods (implementation details). 

No need to review any of these changes here so I am just porting some tests over from the private methods to the public methods.

tests/test_spatial.py
tomvothecoder: I ended up deleting this test because it is faulty and we already have a test for raising a ValueError if the domain bounds has a larger lower bound vs. upper bound.

xcdat/spatial.py
tomvothecoder: Yes, this method was moved and renamed. Nothing else was changed in this file.

tomvothecoder:
Hi @pochedls, @lee1043, and @chengzhuzhang, instead of making `_get_weights()` public for temporal averaging so the user can get weights, I implemented a `keep_weights` keyword argument to the temporal averaging APIs. This gives users the choice with whether or not they want to couple `time_weights` with its related output.

| Option 	| Make `_get_weights()` public 	| Option to keep `time_weights` in the output `xr.Dataset` 	|
|---	|---	|---	|
| Pros 	| * User can calculate weights independently of temporal averaging, which makes it faster to get weights instead of running temporal averaging first (do scientists normally calculate weights independently? does CDAT allow have a method for this?) 	| * Makes sense to couple the weights to its related averaging output <br>* Implementation is straightforward, just add the `time_weights` DataArray to the final output Dataset                         <br>* User has the option to not include the weights in the output 	|
| Cons 	| * User needs to provide the same arguments as temporal averaging APIs (`data_var`, `mode`, `freq`, `season_config`) which makes the method less intuitive to implement and use <br>* Weights are decoupled from the output, which removes context on how they should be used for averaging<br>* We don't accept weights in our temporal averaging APIs right now (does it make sense to?) 	| * For group_average and climatology we must append `"_original"` to the name of `time_weights` time dimension/coordinates since datasets share dimensions and coordinates. Otherwise, it will use the reduced  time<br>dimension/coordinates (not correct, will result in  `np.nan` weights) 	|


Example for `ds.temporal.climatology()`:
https://github.com/xCDAT/xcdat/blob/44620c0953350fffa4df69c1ae2050088d31f058/xcdat/temporal.py#L332-L339
https://github.com/xCDAT/xcdat/blob/44620c0953350fffa4df69c1ae2050088d31f058/xcdat/temporal.py#L742-L745

Example for `ds.temporal.departures()`
https://github.com/xCDAT/xcdat/blob/44620c0953350fffa4df69c1ae2050088d31f058/xcdat/temporal.py#L624-L627
https://github.com/xCDAT/xcdat/blob/44620c0953350fffa4df69c1ae2050088d31f058/xcdat/temporal.py#L465-L472

`_keep_weights()` method:
https://github.com/xCDAT/xcdat/blob/44620c0953350fffa4df69c1ae2050088d31f058/xcdat/temporal.py#L1369-L1402


**If this looks good, we can implement `keep_weights` with the spatial averaging API as well.**

chengzhuzhang:
Hi @tomvothecoder Thank you for the discussion. I don't recall that I have a use case that requires to get/keep the weights to pass down to another operation. With the exception that, they are useful to be examed during xcdat development/validation. So I don't have strong opinion on the implementation, maybe keeping it private is fine. I will defer to Steve and Jiwoo.

pochedls:
I like `keep_weights`. I didn't totally understand the downside for keep_weights: why can't we just overwrite the original time_weights? Are both the full time series and the climatology time series saved in the dataset when calculating a climatology?

I think it would be good to keep `.get_weights` in spatial averaging, though I don't object to adding a `keep_weights` in that routine. One thing that might confuse people is that the weights we generate for spatial averaging are 2D (we never explicitly generate a time resolved weighting matrix).

tomvothecoder:
> I like `keep_weights`. I didn't totally understand the downside for keep_weights: why can't we just overwrite the original time_weights? 

There is only one `time_wts` generated using the `time_bounds`, which is saved in the Dataset when `keep_weights=True`. I'm not sure what you mean by overwriting the original `time_wts`. 

#### Keeping weights with Climatology

* `time_wts` uses the original time coordinates, so it CANNOT share the reduced time dimension and coordinates from the climatology data variable.
  *  `"time_original"` is the original time dimension (1919 coord pts)
  * `"time"` is the reduced time dimension (4 coord pts)

```
ds.temporal.climatology("tas", freq="season", weighted=True, keep_weights=True)

<xarray.Dataset>
Dimensions:        (lon: 320, nv: 2, lat: 160, bnds: 2, time: 4,
                    time_original: 1919)
Coordinates:
  * lon            (lon) float64 0.0 1.125 2.25 3.375 ... 356.6 357.8 358.9
  * lat            (lat) float64 -89.14 -88.03 -86.91 ... 86.91 88.03 89.14
  * time           (time) object 0001-01-01 00:00:00 ... 0001-10-01 00:00:00
  * time_original  (time_original) datetime64[ns] 1850-01-16T12:00:00 ... 200...
Dimensions without coordinates: nv, bnds
Data variables:
    lon_bounds     (lon, nv) float64 dask.array<chunksize=(320, 2), meta=np.ndarray>
    lat_bounds     (lat, nv) float64 dask.array<chunksize=(160, 2), meta=np.ndarray>
    lon_bnds       (lon, bnds) float64 -0.5625 0.5625 0.5625 ... 358.3 359.4
    lat_bnds       (lat, bnds) float64 -89.7 -88.59 -88.59 ... 88.59 88.59 89.7
    tas            (time, lat, lon) float64 dask.array<chunksize=(1, 160, 320), meta=np.ndarray>
    time_wts       (time_original) float64 0.002038 0.002038 ... 0.002083

```

#### Keeping weights with Departures

* `time_wts` uses the original "time" dimension and coordinates, which are maintained from the departures data variable so they can be shared.
* Notice there is no `"time_original"` dim or coords

```
ds.temporal.departures("tas", freq="season", weighted=True, keep_weights=True)

<xarray.Dataset>
Dimensions:     (lon: 320, nv: 2, lat: 160, time: 1919, bnds: 2)
Coordinates:
  * lon         (lon) float64 0.0 1.125 2.25 3.375 ... 355.5 356.6 357.8 358.9
  * lat         (lat) float64 -89.14 -88.03 -86.91 -85.79 ... 86.91 88.03 89.14
  * time        (time) datetime64[ns] 1850-01-16T12:00:00 ... 2009-11-16
Dimensions without coordinates: nv, bnds
Data variables:
    lon_bounds  (lon, nv) float64 dask.array<chunksize=(320, 2), meta=np.ndarray>
    lat_bounds  (lat, nv) float64 dask.array<chunksize=(160, 2), meta=np.ndarray>
    time_bnds   (time, bnds) datetime64[ns] 1850-01-01T18:00:00 ... 2009-12-0...
    tas         (time, lat, lon) float64 dask.array<chunksize=(2, 160, 320), meta=np.ndarray>
    lon_bnds    (lon, bnds) float64 -0.5625 0.5625 0.5625 ... 358.3 358.3 359.4
    lat_bnds    (lat, bnds) float64 -89.7 -88.59 -88.59 ... 88.59 88.59 89.7
    time_wts    (time) float64 0.002038 0.002038 0.002059 ... 0.002083 0.002083

```


> Are both the full time series and the climatology time series saved in the dataset when calculating a climatology?
Only the climatology time series is stored in the climatology output.

> I think it would be good to keep `.get_weights` in spatial averaging, though I don't object to adding a `keep_weights` in that routine. 

Sounds good, I added `keep_weights` as a keyword argument to `ds.spatial.average()`.

> One thing that might confuse people is that the weights we generate for spatial averaging are 2D (we never explicitly generate a time resolved weighting matrix).

Hmmm, we can address this if users open up an issue about it, unless you think it is worth addressing now.

pochedls:
Got it – thanks Tom.

xcdat/temporal.py
```diff
@@ -146,10 +146,9 @@ def __init__(self, dataset: xr.Dataset):
         self._dataset: xr.Dataset = dataset
         self._dim_name = _get_coord_var(self._dataset, "T").name
```
tomvothecoder, line 147:
```suggestion
        self._dataset: xr.Dataset = dataset
        
        self._dim_name = _get_coord_var(self._dataset, "T").name
```

PR #257, opened 2022-06-10
Update `swap_lon_axis` to ignore same systems

### LINKED ISSUE #250: [Bug]: Converting the longitude axis orientation to the same system results in odd behaviors
### What happened?

> As a side note, there is some weird behavior here: the longitude axis goes from size 360 to 361 (and one set of lon_bnds 
> goes from 0 to 0). I'm not sure if this is specific to converting from one longitude coordinate system to the same system 
> (something people wouldn't normally do) or a more generic issue. This doesn't happen when converting to (-180, 180).
> &mdash; @pochedls from https://github.com/xCDAT/xcdat/pull/239#issuecomment-1146235781

The coordinates and coordinate bounds should remain the same if attempting to convert to the same axis system.

Solution: Detect that the desired axis system is the same as the existing system, so use a `pass` statement to ignore.

_No response_

### Relevant log output

_No response_

_No response_

### Environment

`main` branch of xcdat

tomvothecoder:
## Description

- Closes #250 
- This fixes an issue when attempting to swap from (0, 360) to (0, 360) (same orientation), resulting in an additional coord point and lon bounds being added inadvertently. For example:
  1. We have bounds [0, 0.5625] for coordinate point 0
  2. It will convert to [359.4375, 0.5625], which is a prime meridian cell.
  3. Since a prime meridian cell exists, it will add another coordinate point for 360 and a lon bound of [359.4375, 360].
  4. This changes the shape of the lon coords and lon bounds by 1

## Checklist

If applicable:

### DIFF
--- a/xcdat/axis.py
+++ b/xcdat/axis.py
@@ -124,7 +124,7 @@ def swap_lon_axis(
                 "orientations."
             )
 
-    # If the current axis orientation is the same as the desired axis
+    # If the swapped axis orientation is the same as the existing axis
     # orientation, return the original Dataset.
     if new_lon.identical(lon):
         return dataset
--- a/xcdat/axis.py
+++ b/xcdat/axis.py
@@ -125,7 +125,7 @@ def swap_lon_axis(
             )
 
     # If the current axis orientation is the same as the desired axis
-    # orientation, `pass` and return the same dataset.
+    # orientation, return the original Dataset.
     if new_lon.identical(lon):
         return dataset
 
--- a/xcdat/axis.py
+++ b/xcdat/axis.py
@@ -107,7 +107,7 @@ def swap_lon_axis(
         if to == (-180, 180):
             new_lon = ((lon + 180) % 360) - 180
             new_lon_bounds = ((lon_bounds + 180) % 360) - 180
-            ds = _reassign_lon(ds, lon, lon_bounds)
+            ds = _reassign_lon(ds, new_lon, new_lon_bounds)
         elif to == (0, 360):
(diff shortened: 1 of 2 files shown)


xcdat/axis.py
```diff
@@ -124,6 +124,11 @@ def swap_lon_axis(
                 "orientations."
             )
 
+    # If the current axis orientation is the same as the desired axis
+    # orientation, `pass` and return the same dataset.
+    if new_lon.identical(lon):
+        return dataset
```
tomvothecoder, line 130: The fix was to simply compare the original longitude coords with the swapped coords. If they are identical, return the original dataset. Otherwise, return the dataset with the swapped coords.

xcdat/axis.py
```diff
@@ -124,8 +124,13 @@ def swap_lon_axis(
                 "orientations."
             )
 
+    # If the current axis orientation is the same as the desired axis
+    # orientation, `pass` and return the same dataset.
```
tomvothecoder, line 128:
```suggestion
    # If the current axis orientation is the same as the desired axis
    # orientation, return the original Dataset.
```

xcdat/axis.py
```diff
@@ -124,8 +124,13 @@ def swap_lon_axis(
                 "orientations."
             )
 
+    # If the current axis orientation is the same as the desired axis
+    # orientation, return the original Dataset.
```
tomvothecoder, line 128:
```suggestion
    # If the swapped axis orientation is the same as the existing axis
    # orientation, return the original Dataset.
```

PR #262, opened 2022-06-22
Update `get_axis_coord()` to interpret more keys

tomvothecoder:
## Description

Summary of Changes
- Closes #260 
- Closes #215 
- Update `get_axis_coord()` to interpret more keys
   - Rename `get_coord_var()` to `get_axis_coord()`
   - Update `add_missing_bounds()`, `add_bounds()`, and `get_bounds()` logic to use updated `get_axis_coord()`
   - Replace `GENERIC_AXIS_MAP` with `CF_NAME_MAP`
   - Add `CFAxisName`, `CFStandardName`, and `ShortName` type aliases
   - Update variable names in `_add_bounds()` for clarity
   - Move `add_bounds` kwarg in `open_dataset()` and `open_mfdataset()` to a higher position due to importance
- Update regridder APIs to use `get_axis_coord()`
   - Update references to `.cf.get_bounds()` to `.bounds.get_bounds()`
   - Update `grid.create_grid()` to set the `"bounds"` attr on the coord var to avoid `add_missing_bounds()` to attempt to add bounds for global mean grids
- Add `get_axis_dim()`
   - Update spatial averaging APIs to use this function
   - Update `_align_lon_bounds_to_360` to use this function
- Misc Refactor
   - Update name of attr `self._dim_name` to `self._dim` in `TemporalAccessor`
   - Update "is_generated" attr to "xcdat_bounds"
   - Add handles for lat ` "units" ` attr in `_add_bounds()`

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules

If applicable:

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass with my changes (locally and CI/CD build)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have noted that this is a breaking change for a major release (fix or feature that would cause existing functionality to not work as expected)

PR #263, opened 2022-06-23
better handling of non-cf-compliant time data

### LINKED ISSUE #261: [Bug]: Failure to decode non-CF compliant time
### What happened?

I tried opening a dataset with non-CF compliant time units. We don't currently have code to process this particular type of non-CF compliant time axis so I got an error. 

Ideally, we could change `open_dataset` (and `open_mfdataset`) to attempt to `decode_non_cf_time` and if it fails, simply do not decode the time axis, e.g., [here](https://github.com/xCDAT/xcdat/blob/main/xcdat/dataset.py#L87):

```
        if cf_compliant_time is False:
            # XCDAT handles decoding time values with non-CF units.
            ds = xr.open_dataset(path, decode_times=False, **kwargs)
            try:
                 ds = decode_non_cf_time(ds)
            except:
                 # some log message to note that decoding failed
```

[Note that there would also need to be some error handling in `_has_cf_compliant_time`]

```python
fn = '/p/user_pub/climate_work/pochedley1/surface/best.nc'
ds = xcdat.open_dataset(fn)
```

### Relevant log output

```python
~/code/xcdat/xcdat/dataset.py in _split_time_units_attr(units_attr)
    612         raise KeyError("No 'units' attribute found for the dataset's time coordinates.")
    613 
--> 614     units, reference_date = units_attr.split(" since ")
    615 
    616     return units, reference_date

ValueError: not enough values to unpack (expected 2, got 1)
```

This error results because we were expecting units of the form "X since Y", but the units are `year A.D.` (time axis has decimal years, i.e., 1850.041667, 1850.125, ...).

Should I attempt a PR?

### Environment

Recent-ish xcdat dev environment.

pochedls:
## Description

xcdat can decode some non-cf compliant data, but it cannot handle every case. This update makes it so that xcdat attempts to decode non-cf-compliant data and if it fails it returns a non-decoded dataset (instead of raising a ValueError).

- Closes #261 

## Checklist

If applicable:

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -52,7 +52,15 @@ time_non_cf = xr.DataArray(
         "standard_name": "time",
     },
 )
-
+time_non_cf_unsupported = xr.DataArray(
+    data=np.arange(1850 + 1 / 24.0, 1851 + 3 / 12.0, 1 / 12.0),
+    dims=["time"],
+    attrs={
+        "units": "year A.D.",
+        "long_name": "time",
+        "standard_name": "time",
+    },
+)
 time_bnds = xr.DataArray(
     name="time_bnds",
     data=np.array(
@@ -103,6 +111,16 @@ time_bnds_non_cf = xr.DataArray(
     coords={"time": time_non_cf},
     dims=["time", "bnds"],
     attrs={"xcdat_bounds": "True"},
+)
+tb = []
+for t in time_non_cf_unsupported:
+    tb.append([t - 1 / 24.0, t + 1 / 24.0])
+time_bnds_non_cf_unsupported = xr.DataArray(
+    name="time_bnds",
+    data=tb,
+    coords={"time": time_non_cf_unsupported},
+    dims=["time", "bnds"],
+    attrs={"is_generated": "True"},
 )
 
 # LATITUDE
@@ -159,7 +177,9 @@ ts_non_cf = xr.DataArray(
 )
 
 
-def generate_dataset(cf_compliant: bool, has_bounds: bool) -> xr.Dataset:
+def generate_dataset(
+    cf_compliant: bool, has_bounds: bool, unsupported=False
+) -> xr.Dataset:
     """Generates a dataset using coordinate and data variable fixtures.
 
     Parameters
@@ -189,8 +209,12 @@ def generate_dataset(cf_compliant: bool, has_bounds: bool) -> xr.Dataset:
             ds.coords["time"] = time_cf.copy()
             ds["time_bnds"] = time_bnds.copy()
         elif not cf_compliant:
-            ds.coords["time"] = time_non_cf.copy()
-            ds["time_bnds"] = time_bnds_non_cf.copy()
+            if unsupported:
+                ds.coords["time"] = time_non_cf_unsupported.copy()
+                ds["time_bnds"] = time_bnds_non_cf_unsupported.copy()
+            else:
+                ds.coords["time"] = time_non_cf.copy()
+                ds["time_bnds"] = time_bnds_non_cf.copy()
 
         # If the "bounds" attribute is included in an existing DataArray and
         # added to a new Dataset, it will get dropped. Therefore, it needs to be
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -178,7 +178,7 @@ ts_non_cf = xr.DataArray(
 
 
 def generate_dataset(
-    cf_compliant: bool, has_bounds: bool, unsupported=False
+    cf_compliant: bool, has_bounds: bool, unsupported: bool = False
 ) -> xr.Dataset:
     """Generates a dataset using coordinate and data variable fixtures.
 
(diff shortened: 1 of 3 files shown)


tests/fixtures.py
```diff
@@ -104,6 +112,16 @@
...
+tb = []
+for t in time_non_cf_unsupported:
+    tb.append([t - 1 / 24.0, t + 1 / 24.0])
+time_bnds_non_cf_unsupported = xr.DataArray(
+    name="time_bnds",
+    data=tb,
+    coords={"time": time_non_cf_unsupported},
+    dims=["time", "bnds"],
+    attrs={"is_generated": "True"},
+)
```
pochedls, line 124: I created test bounds for a dataset that is non-cf-compliant (but not handled by xcdat), but I didn't end up using these in the unit test I created (see below).

xcdat/dataset.py
```diff
@@ -304,7 +305,13 @@ def decode_non_cf_time(dataset: xr.Dataset) -> xr.Dataset:
...
     time_bounds = ds.get(time.attrs.get("bounds"), None)
     units_attr = time.attrs.get("units")
-    units, ref_date = _split_time_units_attr(units_attr)
+    # if the time axis cannot be split, we do not yet
+    # support time decoding and we return the original
+    # dataset
+    try:
+        units, ref_date = _split_time_units_attr(units_attr)
+    except ValueError:
+        return ds
```
pochedls, line 314: If there is no "since" in the time units, then the dataset is non-cf-compliant and unsupported, so instead of continuing this function (`decode_non_cf_time`) simply returns the original (non-decoded) dataset.

xcdat/dataset.py
```diff
@@ -403,7 +410,11 @@ def _has_cf_compliant_time(
         return None
 
     time = ds.cf["T"]
-    units = _split_time_units_attr(time.attrs.get("units"))[0]
+    # if the time units attr cannot be split it is not cf_compliant
+    try:
+        units = _split_time_units_attr(time.attrs.get("units"))[0]
+    except ValueError:
+        return False
```
pochedls, line 417: My read is that if there is no "since" the time axis is not cf-compliant [here](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch04s04.html).

durack1:
@pochedls if you want a quick check of CF validity, point your file at https://pumatest.nerc.ac.uk/cgi-bin/cf-checker.pl - catches all manner of things that aren't obvious through a quick `ncdump` scan

tests/test_dataset.py
```diff
@@ -43,9 +43,9 @@ def test_non_cf_compliant_and_unsupported_time_is_not_decoded(self):
 
         # even though decode_times=True, it should fail to decode unsupported time axis
         result = open_dataset(self.file_path, decode_times=True)
-        expected_times = np.arange(1850 + 1 / 24.0, 1851 + 3 / 12.0, 1 / 12.0)
+        expected = ds
 
-        assert np.all(expected_times == result.time.values)
+        assert result.identical(expected)
```
tomvothecoder, line 48: I updated the assertion statement since we expect the result to be the same as the original dataset.

xcdat/dataset.py
```diff
@@ -620,15 +623,17 @@ def _split_time_units_attr(units_attr: str) -> Tuple[str, str]:
...
     ------
+    KeyError
+        If the time units attribute was not found.
+
     ValueError
         If the time units attribute is not of the form `X since Y`.
     """
     if units_attr is None:
-        raise KeyError("No 'units' attribute found for the dataset's time coordinates.")
+        raise KeyError("The dataset's time coordinates does not have a 'units' attr.")
```
tomvothecoder, line 636: Updated a preexisting KeyError message

tomvothecoder (approved):
Overall, it looks great to me! 

After you take a look at some of my minor updates in this [commit](https://github.com/xCDAT/xcdat/pull/263/commits/a2f5fbd0483609a452c1cdccc3ffd511f5451409), feel free to merge.

tomvothecoder:
Oh yeah, I also rebased this branch on the latest `main` so there are no merge conflicts.

pochedls:
> I updated the assertion statement since we expect the result to be the same as the original dataset.

Thank you...this makes a lot of sense and I wish this is what I had done.

> Oh yeah, I also rebased this branch on the latest `main` so there are no merge conflicts.

You just needed to do this because you just merged #262, right? Or did I make a mistake?

I will squash and merge shortly. Thank you for the review, @tomvothecoder.

tomvothecoder:
> Just a note that I am getting some warnings, but I don't think I caused them:
> 
> > tests/test_regrid.py::TestRegrid2Regridder::test_output_bounds
> > tests/test_regrid.py::TestAccessor::test_convenience_methods
> > /home/pochedley1/code/xcdat/xcdat/regridder/regrid2.py:233: RuntimeWarning: invalid value encountered in true_divide
> > np.multiply(input_lon_segment, dot_weight).sum(
> > tests/test_regrid.py::TestXESMFRegridder::test_regrid
> > /home/pochedley1/bin/anaconda3/envs/xcdat_dev/lib/python3.9/site-packages/numba/np/arraymath.py:3845: DeprecationWarning: `np.MachAr` is deprecated (NumPy 1.22).
> > @overload(np.MachAr)

This is from `numba`, which was introduced in PR #44. I'll see if there's a way to fix it in a separate PR.

> > Oh yeah, I also rebased this branch on the latest `main` so there are no merge conflicts.
> 
> You just needed to do this because you just merged #262, right? Or did I make a mistake?
> 
> I will squash and merge shortly. Thank you for the review, @tomvothecoder.

No mistakes were made, #262 was merged so `main` was ahead of this branch. Just needed to rebase on top of the latest `main` to avoid conflicts.

Thanks for the great work!

PR #300, opened 2022-08-03
Bump to 0.3.1

tomvothecoder:
## Description

* Bumps the package to 0.3.1
* Updates the dependencies in the conda env yml files
* Updates the pre-commit hook versions
* Address `mypy` warnings
  * Remove `type: ignore` inline comments to silence mypy warnings related to xarray, can address in the future
  * Remove unused `type: ignore` based on latest version of mypy

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules

If applicable:

- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass with my changes (locally and CI/CD build)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have noted that this is a breaking change for a major release (fix or feature that would cause existing functionality to not work as expected)

Issue #285, opened 2022-07-28
[Bug]: Issues opening up datasets with multiple coord vars for the same axis mapped to single dim/independent dims (IPSL ESM files)


oliviermarti:
### What happened?

The IPSL Earth System Model output files has two time variables, and xcdat refuses to open them.

#### Developer Notes (@tomvothecoder, 8/30/22)
Dataset Info:
* Single time dimension called `time_counter`
* Two coord vars called `time_counter` and `time_variable` mapped to `time_counter` dimension

How xcdat handles coords:
* xcdat is designed to handle single coord for an axis mapped to a single dim (1:1 mapping)
* xcdat uses `cf_xarray.cf["T"]` to get the coord var, which breaks if there are multiple coord vars found
  * Have to use `cf_xarray.cf[["T"]]` to return a Dataset with all the coord vars mapped to an axis

### Minimal Complete Verifiable Example

```python
# Test1 - Original IPSL file (not CF compliant)

import numpy as np, xarray as xr, xcdat as xc
TS_File ='https://thredds-su.ipsl.fr/thredds/dodsC/ipsl_thredds/omamce/TestCases/XCDAT/thetao_noCF.nc'
dTS = xc.open_dataset (TS_File, add_bounds=False, use_cftime=True, decode_times=True)

# Test2 - Remove attributes referring to non-existing bounds
# This file is CF compliant (https://pumatest.nerc.ac.uk/cgi-bin/cf-checker.pl)

import numpy as np, xarray as xr, xcdat as xc
TS_File ='https://thredds-su.ipsl.fr/thredds/dodsC/ipsl_thredds/omamce/TestCases/XCDAT/thetao_CF.nc'
dTS = xc.open_dataset (TS_File, add_bounds=False, use_cftime=True, decode_times=True)

# Test3 - remove attr 'coordinates: "time_centered nav_lat nav_lon"' from variable 'thetao' attributes

import numpy as np, xarray as xr, xcdat as xc
TS_File ='https://thredds-su.ipsl.fr/thredds/dodsC/ipsl_thredds/omamce/TestCases/XCDAT/thetao_CF_1.nc'
dTS = xc.open_dataset (TS_File, add_bounds=False, use_cftime=True, decode_times=True)

tos_year = dTS.temporal.group_average ("tos", freq="year", weighted=True)
```

### Relevant log output

```python
# Test 1
[...]

PR #302, opened 2022-08-04
Use cftime for temporal averaging operations

### LINKED ISSUE #301: [Bug]: Out of bounds timestamp for temporal averaging
### What happened?

In trying to compute annual averages for data with non-cf-compliant time units I get an OutOfBoundsError. 

I think this functionality should work with non-cf-compliant time. 

```python
import xcdat

fn = '/p/user_pub/climate_work/pochedley1/cmip6_msu/ttt_Amon_ACCESS-ESM1-5_piControl_r1i1p1f1_gr_10101-100012.nc'
ds = xcdat.open_dataset(fn)
ds_yearly = ds.temporal.group_average('tf2', freq='year', weighted=True)
```

### Relevant log output

```python
.
.
.

File ~/code/xcdat/xcdat/temporal.py:1358, in TemporalAccessor._convert_df_to_dt(self, df)
   1351     dates = [
   1352         cftime.datetime(year, month, day, hour)
   1353         for year, month, day, hour in zip(
   1354             df_new.year, df_new.month, df_new.day, df_new.hour
   1355         )
   1356     ]
   1357 else:
-> 1358     dates = pd.to_datetime(df_new)
   1360 return np.array(dates)

.
.
.

ValueError: cannot assemble the datetimes: Out of bounds nanosecond timestamp: 1010-10-01 00:00:00
```

This issue does not occur with original data (with cf-compliant datetimes).

The condensed log above shows that this issue arises in the `_convert_df_to_dt` function when the dataframe is cast to datetime objects using pandas (because `year_is_unused = False`). `year_is_unused` is set to `False` because the mode is set to `average` and the `_freq` is not yearly. I don't totally understand why we use [`cftime`](https://github.com/xCDAT/xcdat/blob/main/xcdat/temporal.py#L1352) for this step for some conditions, but use [pandas](https://github.com/xCDAT/xcdat/blob/main/xcdat/temporal.py#L1358) for others.

Can we use `cftime` for all cases? I'm wondering if we started by using pandas and then used cftime for some conditions (e.g., [see the documentation](https://github.com/xCDAT/xcdat/blob/main/xcdat/temporal.py#L1316)), but in reality we may be able to simply use `cftime`. @tomvothecoder – do you remember these details well enough to know?

If I remove the `if year_is_unused` conditional and re-run the code I don't get an error. But I fail a bunch of unit tests – I'm not sure if this is a unit issue or an actual problem. 

Also note that it is possible that this issue affects other functions/functionality (but using `cftime` might resolve those potential issues). 

### Environment

main branch

pochedls:
## Description
This is an attempted fix for #301. The PR uses cftime.datetime objects in the `_convert_df_to_dt` function (instead of datetime / cfdatetime objects depending on the dataset attributes). 

- Closes #301

## Checklist

If applicable:

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -41,17 +41,26 @@ time_cf = xr.DataArray(
         "standard_name": "time",
     },
 )
+# NOTE: With `decode_times=True`, the "calendar" and "units" attributes are
+# stored in `.encoding`.
+time_cf.encoding["calendar"] = "standard"
+time_cf.encoding["units"] = "days since 2000-01-01"
+
+
+# NOTE: With `decode_times=False`, the "calendar" and "units" attributes are
+# stored in `.attrs`.
 time_non_cf = xr.DataArray(
     data=[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14],
     dims=["time"],
     attrs={
-        "units": "months since 2000-01-01",
-        "calendar": "standard",
         "axis": "T",
         "long_name": "time",
         "standard_name": "time",
+        "calendar": "standard",
+        "units": "months since 2000-01-01",
     },
 )
+
 time_non_cf_unsupported = xr.DataArray(
     data=np.arange(1850 + 1 / 24.0, 1851 + 3 / 12.0, 1 / 12.0),
     dims=["time"],
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -64,22 +64,53 @@ class TestOpenDataset:
             name="time",
             data=np.array(
                 [
-                    cftime.datetime(2000, 1, 1),
-                    cftime.datetime(2000, 2, 1),
-                    cftime.datetime(2000, 3, 1),
-                    cftime.datetime(2000, 4, 1),
-                    cftime.datetime(2000, 5, 1),
-                    cftime.datetime(2000, 6, 1),
-                    cftime.datetime(2000, 7, 1),
-                    cftime.datetime(2000, 8, 1),
-                    cftime.datetime(2000, 9, 1),
-                    cftime.datetime(2000, 10, 1),
-                    cftime.datetime(2000, 11, 1),
-                    cftime.datetime(2000, 12, 1),
-                    cftime.datetime(2001, 1, 1),
-                    cftime.datetime(2001, 2, 1),
-                    cftime.datetime(2001, 3, 1),
+                    cftime.DatetimeGregorian(
+                        2000, 1, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 2, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 3, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 4, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 5, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 6, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 7, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 8, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 9, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 10, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 11, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2000, 12, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2001, 1, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2001, 2, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
+                    cftime.DatetimeGregorian(
+                        2001, 3, 1, 0, 0, 0, 0, has_year_zero=False
+                    ),
                 ],
+                dtype="object",
             ),
             dims="time",
         )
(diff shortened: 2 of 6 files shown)


tomvothecoder:
Thanks for this PR. I already started taking a look, but will be able to review more closely next week.

We should have this merged before #300.

tests/fixtures.py
```diff
@@ -41,17 +41,26 @@
...
+time_cf.encoding["calendar"] = "standard"
+time_cf.encoding["units"] = "days since 2000-01-01"
+
+
+# NOTE: With `decode_times=False`, the "calendar" and "units" attributes are
+# stored in `.attrs`.
 time_non_cf = xr.DataArray(
     data=[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14],
     dims=["time"],
     attrs={
```
tomvothecoder, line 55: Just bringing this to attention.

tests/test_dataset.py
```diff
@@ -64,22 +64,53 @@ def test_non_cf_compliant_time_is_decoded(self):
...
-                    cftime.datetime(2000, 8, 1),
-                    cftime.datetime(2000, 9, 1),
-                    cftime.datetime(2000, 10, 1),
-                    cftime.datetime(2000, 11, 1),
-                    cftime.datetime(2000, 12, 1),
-                    cftime.datetime(2001, 1, 1),
-                    cftime.datetime(2001, 2, 1),
-                    cftime.datetime(2001, 3, 1),
+                    cftime.DatetimeGregorian(
+                        2000, 1, 1, 0, 0, 0, 0, has_year_zero=False
```
tomvothecoder, line 68: All of the `cftime.datetime` objects in our tests are now updated to `cftime.<OBJECT_TYPE>`.

tests/test_temporal.py
```diff
@@ -374,8 +388,9 @@ def setup(self):
             dims=["time"],
             attrs={"axis": "T", "long_name": "time", "standard_name": "time"},
         )
+        time.encoding = {"calendar": "standard"}
         time_bnds = xr.DataArray(
-            name="time_bnds",
+            name="time_bndsZ",
```
tomvothecoder, line 393: TODO: Remove accidental addition of "Z"

xcdat/dataset.py
```diff
@@ -321,34 +321,42 @@ def decode_non_cf_time(dataset: xr.Dataset) -> xr.Dataset:
...
+    time_attrs = time.attrs
+
+    # NOTE: When opening datasets with `decode_times=False`, the "calendar" and
+    # "units" attributes are stored in `.encoding` (unlike `decode_times=True`
+    # which stores them in `.attrs`). Since xCDAT manually decodes non-CF
+    # compliant time coordinates by first setting `decode_times=False`, the
+    # "calendar" and "units" attrs are popped from the `.attrs` dict and stored
+    # in the `.encoding` dict to mimic xarray's behavior.
+    calendar = time_attrs.pop("calendar", None)
+    units_attr = time_attrs.pop("units", None)
```
tomvothecoder, line 333: In `decode_non_cf_time()`, the "calendar" and "units" attributes were being saved in both the `.attrs` and `.encoding` dicts.

However, they should only be in the `.encoding` dict **after decoding time coordinates**. To achieve this behavior, we pop these attrs from the `.attrs` dict and later add them to `.encoding`.

xcdat/temporal.py
```diff
@@ -150,6 +150,17 @@ def __init__(self, dataset: xr.Dataset):
...
+        try:
+            self.calendar = self._dataset[self._dim].encoding["calendar"]
+            self.date_type = get_date_type(self.calendar)
+        except KeyError:
+            raise KeyError(
+                "This dataset's time coordinates do not have a 'calendar' encoding "
+                "attribute set, which might indicate that the time coordinates were not "
+                "decoded to datetime objects. Ensure that the time coordinates are "
+                "decoded before performing temporal averaging operations."
+            )
```
tomvothecoder, line 162: We now consider the "calendar" attribute and the associated `cftime` date type object for temporal averaging operations.

xcdat/temporal.py
```diff
@@ -1343,7 +1355,7 @@ def _convert_df_to_dt(self, df: pd.DataFrame) -> np.ndarray:
                 df_new[component] = default_val
 
         dates = [
-            cftime.datetime(year, month, day, hour)
+            self.date_type(year, month, day, hour)
```
tomvothecoder, line 1358: The `date_type` is used to enable xarray's indexing/subsetting with `cftime` objects.

tomvothecoder (commented):
Hi @pochedls, I made some updates for review. 

The diff compares your commit to mine. Let me know how it looks.

xcdat/dataset.py
```diff
@@ -321,34 +321,42 @@ def decode_non_cf_time(dataset: xr.Dataset) -> xr.Dataset:
...
-        calendar = time.attrs["calendar"]
-    except KeyError:
+    time_attrs = time.attrs
+
+    # NOTE: When opening datasets with `decode_times=False`, the "calendar" and
+    # "units" attributes are stored in `.encoding` (unlike `decode_times=True`
+    # which stores them in `.attrs`). Since xCDAT manually decodes non-CF
+    # compliant time coordinates by first setting `decode_times=False`, the
+    # "calendar" and "units" attrs are popped from the `.attrs` dict and stored
+    # in the `.encoding` dict to mimic xarray's behavior.
```
tomvothecoder, line 331:
```suggestion
    # NOTE: When opening datasets with `decode_times=False`, the "calendar" and
    # "units" attributes are stored in `.attrs` (unlike `decode_times=True`
    # which stores them in `.encoding`). Since xCDAT manually decodes non-CF
    # compliant time coordinates by first setting `decode_times=False`, the
    # "calendar" and "units" attrs are popped from the `.attrs` dict and stored
    # in the `.encoding` dict to mimic xarray's behavior.
```

xcdat/temporal.py
pochedls: Should "were not decoded to datetime objects" be "were not decoded to cftime.datetime objects"? Or are they sufficiently interchangeable?

tests/fixtures.py
pochedls: I think this makes sense. This is a separate issue when I open a dataset with `xarray.open_mfdataset` and `decode_times=True`, I don't see anything populated in `ds.time.encoding` (but I do see calendar and units information stored here when using `xarray.open_dataset`).

pochedls:
> 

@tomvothecoder - I had a couple minor questions, but this looks good to me.

xcdat/temporal.py
tomvothecoder: Temporal averaging accepts both `datetime.datetime` and `cftime.datetime` objects, but the end result will always be `cftime.datetime`.

I think it is sufficiently interchangeable to reference both as just `datetime`.

tests/fixtures.py
tomvothecoder: Nice catch! I found an open xarray issue related to `xr.open_mfdataset()` dropping encoding attributes.

https://github.com/pydata/xarray/issues/2436
https://github.com/pydata/xarray/issues/2436#issuecomment-449737841

\

Somebody mentions this about the issue here https://github.com/pydata/xarray/issues/2436#issuecomment-888204512:

> However, there is another problem related to xarray.open_mfdataset:
The encoding dictionary gets lost somewhere during the merging operation of the datasets of the respective files (https://github.com/pydata/xarray/issues/2436).
>
> This leads to problems for example with cf-xarray when trying to detect coordinates or bounds, but also leads to problems related to the time axis encoding apparently (as seen in the linked issue). I managed at least to avoid the problems for cf-xarray bounds and coordinates detection by using the decode functionality of xarray only after the datasets have been read in (leaving however the unnecessary time dimension in place ...):
> 
> ds = xarray.open_mfdataset("/path/to/files/*.nc")
> ds = xarray.decode_cf(ds, decode_coords="all")

---

**I think we might need to use the `preprocess` function to save the `.encoding` dict.**

tomvothecoder: We can open up a separate issue for this. I'm glad you're finding all of these nonobvious behaviors/bugs in xarray and xCDAT.

tomvothecoder (approved):
This PR should be ready to merge if you are satisfied with it.

I will continue work with #308 and #285.

pochedls:
> This PR should be ready to merge if you are satisfied with it.
> 
> I will continue work with #308 and #285.

Thanks for the help and getting this done so quickly.

PR #309, opened 2022-08-11
Fix `open_mfdataset()` dropping time encoding attrs

### LINKED ISSUE #308: [Bug]: `xarray.open_mfdataset` drops `.encoding` attributes, need to handle with `xcdat.open_mfdataset()`
### What happened?

Related to https://github.com/xCDAT/xcdat/pull/302#discussion_r942905147.

> Nice catch! I found an open xarray issue related to xr.open_mfdataset() dropping encoding attributes.
>
> - https://github.com/pydata/xarray/issues/2436
> - https://github.com/pydata/xarray/issues/2436#issuecomment-449737841
> 
> Somebody mentions this about the issue here https://github.com/pydata/xarray/issues/2436#issuecomment-888204512:
> 
>
>>     However, there is another problem related to xarray.open_mfdataset:
>>     The encoding dictionary gets lost somewhere during the merging operation of the datasets of the respective files (https://github.com/pydata/xarray/issues/2436).
>> 
>>     This leads to problems for example with cf-xarray when trying to detect coordinates or bounds, but also leads to problems related to the time axis encoding apparently (as seen in the linked issue). I managed at least to avoid the problems for cf-xarray bounds and coordinates detection by using the decode functionality of xarray only after the datasets have been read in (leaving however the unnecessary time dimension in place ...):
>> 
>>     ds = xarray.open_mfdataset("/path/to/files/*.nc")
>>     ds = xarray.decode_cf(ds, decode_coords="all")
>
> 
> I think we might need to use the preprocess function to save the .encoding dict, or some other workaround like from one of the comments.
> 

The temporal averaging APIs require decoded time coordinates and the `ds.time.encoding` dict to be populated with the `"calendar"` attr.

The `.encoding` attributes should remain persistent after running `xcdat.open_mfdataset()`. 

This does not affect `xcdat.open_dataset()`.

```python
>> import xcdat

>> ds = xcdat.open_mfdataset(
    "/p/css03/cmip5_css02/data/cmip5/output1/ICHEC/EC-EARTH/historical/mon/atmos/Amon/r1i1p1/tas/1/*.nc")

>> ds.time.encoding

{}
```

### Relevant log output

_No response_

Possible Solutions
1. Store .encoding and copy it back to the final merged dataset (maybe with `preprocess kwarg`)
  - https://github.com/pydata/xarray/issues/2436#issuecomment-610008589

Related Issues

### Environment

xcdat 0.3.0

INSTALLED VERSIONS
------------------
commit: None
python: 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:56:21) 
[GCC 10.3.0]
python-bits: 64
OS: Linux
OS-release: 3.10.0-1160.45.1.el7.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: ('en_US', 'UTF-8')
libhdf5: 1.12.1
libnetcdf: 4.8.1

xarray: 2022.3.0
pandas: 1.4.3
numpy: 1.22.4
scipy: 1.9.0
netCDF4: 1.5.8
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: 1.6.0
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: None
dask: 2022.6.1
distributed: 2022.6.1
matplotlib: 3.5.2
cartopy: None
seaborn: None
numbagg: None
fsspec: 2022.7.1
cupy: None
pint: None
sparse: 0.13.0
setuptools: 63.3.0
pip: 22.1.2
conda: None
pytest: 7.1.2
IPython: 8.4.0
sphinx: 4.5.0

tomvothecoder:
## Description

- Closes #308 

## Checklist

If applicable:

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -341,8 +341,11 @@ class TestOpenMfDataset:
             "units": "days since 2000-01-01",
             "calendar": "standard",
         }
-
         assert result.identical(expected)
+
+        # Delete '_FillValue' dict entry because it is `np.nan`, and asserting
+        # equality between two `np.nan` will result in False.
+        del result.time.encoding["_FillValue"]
         assert result.time.encoding == expected.time.encoding
 
     def test_non_cf_compliant_time_is_not_decoded(self):
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -305,6 +305,46 @@ class TestOpenMfDataset:
         self.file_path1 = f"{dir}/file1.nc"
         self.file_path2 = f"{dir}/file2.nc"
 
+    def test_mfdataset_keeps_time_encoding_dict(self):
+        # FIXME: This test always passes because `xr.open_mfdatset()` always
+        # keeps the time encoding attrs, which isn't the expected behavior.
+        # Based on this test, if datasets are generated in xarray and written
+        # out with `to_netcdf()` and then opened and merged using
+        # `xr.open_mfdataset()`, the time encoding attributes are not dropped.
+        # On the other hand, if multiple real world datasets that did not
+        # originate from xarray (written out with `.to_netcdf()`) are opened
+        # using `xr.open_mfdataset()`, the time encoding attrs are dropped.
+        # (Refer to https://github.com/pydata/xarray/issues/2436). My theory is
+        # that xarray maintains the time encoding attrs if datasets are written
+        # out with `.to_netcdf()`, and drops it for other cases such
+        # as opening multiple datasets from other sources.
+        ds1 = generate_dataset(cf_compliant=True, has_bounds=True)
+        ds1.to_netcdf(self.file_path1)
+        ds2 = generate_dataset(cf_compliant=True, has_bounds=True)
+        ds2 = ds2.rename_vars({"ts": "tas"})
+        ds2.to_netcdf(self.file_path2)
+
+        result = open_mfdataset([self.file_path1, self.file_path2], decode_times=True)
+
+        expected = ds1.copy().merge(ds2.copy())
+        expected.time.encoding = {
+            "zlib": False,
+            "shuffle": False,
+            "complevel": 0,
+            "fletcher32": False,
+            "contiguous": True,
+            "chunksizes": None,
+            # Set source as result source because it changes every test run.
+            "source": result.time.encoding["source"],
+            "original_shape": (15,),
+            "dtype": np.dtype(np.int64),
+            "units": "hours since 2000-01-16 12:00:00",
+            "calendar": "proleptic_gregorian",
+        }
+
+        assert result.identical(expected)
+        assert result.time.encoding == expected.time.encoding
+
     def test_non_cf_compliant_time_is_not_decoded(self):
         ds1 = generate_dataset(cf_compliant=False, has_bounds=True)
         ds1.to_netcdf(self.file_path1)
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -306,34 +306,34 @@ class TestOpenMfDataset:
         self.file_path2 = f"{dir}/file2.nc"
 
     def test_mfdataset_keeps_time_encoding_dict(self):
-        # FIXME: This test always passes because `xr.open_mfdatset()` always
-        # keeps the time encoding attrs, which isn't the expected behavior.
-        # Based on this test, if datasets are generated in xarray and written
-        # out with `to_netcdf()` and then opened and merged using
-        # `xr.open_mfdataset()`, the time encoding attributes are not dropped.
-        # On the other hand, if multiple real world datasets that did not
-        # originate from xarray (written out with `.to_netcdf()`) are opened
-        # using `xr.open_mfdataset()`, the time encoding attrs are dropped.
-        # (Refer to https://github.com/pydata/xarray/issues/2436). My theory is
-        # that xarray maintains the time encoding attrs if datasets are written
-        # out with `.to_netcdf()`, and drops it for other cases such
-        # as opening multiple datasets from other sources.
         ds1 = generate_dataset(cf_compliant=True, has_bounds=True)
         ds1.to_netcdf(self.file_path1)
+
+        # Create another dataset that extends the time coordinates by 1 value,
+        # to mimic a multifile dataset.
         ds2 = generate_dataset(cf_compliant=True, has_bounds=True)
-        ds2 = ds2.rename_vars({"ts": "tas"})
+        ds2 = ds2.isel(dict(time=slice(0, 1)))
+        ds2["time"].values[:] = np.array(
+            ["2002-01-16T12:00:00.000000000"],
+            dtype="datetime64[ns]",
+        )
         ds2.to_netcdf(self.file_path2)
 
         result = open_mfdataset([self.file_path1, self.file_path2], decode_times=True)
-        expected = ds1.copy().merge(ds2.copy())
+        expected = ds1.merge(ds2)
 
         assert result.identical(expected)
 
-        # We mainly care for calendar and units attributes. We don't perform
-        # equality assertion on the entire time `.encoding` dict because there
-        # might be different encoding attributes added or removed between xarray
-        # versions (e.g., "bzip2", "ztsd", "blosc", and "szip" are added in
-        # v2022.06.0), which makes that assertion fragile.
+        # We mainly care for the "source" and "original_shape" attrs (updated
+        # internally by xCDAT), and the "calendar" and "units" attrs. We don't
+        # perform equality assertion on the entire time `.encoding` dict because
+        # there might be different encoding attributes added or removed between
+        # xarray versions (e.g., "bzip2", "ztsd", "blosc", and "szip" are added
+        # in v2022.06.0), which makes that assertion fragile.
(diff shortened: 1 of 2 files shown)


xcdat/dataset.py
```diff
@@ -393,16 +406,41 @@ def decode_non_cf_time(dataset: xr.Dataset) -> xr.Dataset:
...
+
+    # xcdat.open_dataset() is called instead of xr.open_dataset() because
+    # we want to handle decoding non-CF compliant as well.
+    # FIXME: Remove `type: ignore` comment after properly handling the type
+    # annotations in `_get_first_path()`.
+    ds = open_dataset(first_path, decode_times=True, add_bounds=False)  # type: ignore
+
+    time_coord = get_axis_coord(ds, "T")
+
+    return time_coord.encoding
```
tomvothecoder, line 440: The new `_keep_time_encoding()` function.

xcdat/dataset.py
```diff
@@ -462,12 +489,45 @@ def _has_cf_compliant_time(
...
+        first_file = glob(path)[0]
+    elif isinstance(path, str) or isinstance(path, pathlib.Path):
+        first_file = path
+    elif isinstance(path, list):
+        if any(isinstance(sublist, list) for sublist in path):
+            first_file = path[0][0]  # type: ignore
+        else:
+            first_file = path[0]  # type: ignore
+
+    return first_file
```
tomvothecoder, line 521: Extracted `_get_first_path()` to be reused in other functions (`_keep_time_encoding()` and `_has_cf_compliant_time()`).

xcdat/dataset.py
```diff
@@ -526,6 +590,10 @@ def _postprocess_dataset(
                 "This dataset does not have longitude coordinates to reorient."
             )
 
+    if time_encoding is not None:
+        time_dim = get_axis_dim(dataset, "T")
+        dataset[time_dim].encoding = time_encoding
+
```
tomvothecoder, line 596: There were to options for where to include the logic for adding `time_encoding` attrs back to the final dataset:

1. In `_postprocess_dataset()`, which is shown here
   - On the other hand, this function is used by both `open_dataset()` and `open_mfdataset()`, but this logic is only needed for `open_mfdataset()`. 
2. Keep this logic stored in `open_mfdataset()`, based on above reasoning.

I chose 1 since it is a "postprocessing" operation, although I am open to 2 since the logic is specific to `open_mfdataset()`.

I will think about this more.

tests/test_dataset.py
```diff
@@ -194,6 +194,46 @@ def setUp(self, tmp_path):
...
+        # Based on this test, if datasets are generated in xarray and written
+        # out with `to_netcdf()` and then opened and merged using
+        # `xr.open_mfdataset()`, the time encoding attributes are not dropped.
+        # On the other hand, if multiple real world datasets that did not
+        # originate from xarray (written out with `.to_netcdf()`) are opened
+        # using `xr.open_mfdataset()`, the time encoding attrs are dropped.
+        # (Refer to https://github.com/pydata/xarray/issues/2436). My theory is
+        # that xarray maintains the time encoding attrs if datasets are written
+        # out with `.to_netcdf()`, and drops it for other cases such
+        # as opening multiple datasets from other sources.
```
tomvothecoder, line 209: A giant FIXME: comment for why this test always passes.

One workaround might be to include a subset of a real world dataset in the test suite, although I prefer to not do this to keep the test suite lightweight and reproducible without the need to reference a specific file. We can address this in the future.

xcdat/dataset.py
```diff
@@ -14,14 +14,24 @@
...
 
+# Type annotation for the `paths` arg.
+Paths = Union[
+    str,
+    pathlib.Path,
+    List[str],
+    List[pathlib.Path],
+    List[List[str]],
+    List[List[pathlib.Path]],
+]
```
tomvothecoder, line 33: Extracted `Paths` type annotation since it was reused in multiple functions and making the function param definitions long. 

I noticed xarray refactored the `paths` type annotation here: https://github.com/pydata/xarray/blob/f8fee902360f2330ab8c002d54480d357365c172/xarray/backends/api.py#L734

`    paths: str | NestedSequence[str | os.PathLike],`

xcdat/dataset.py
```diff
@@ -462,12 +489,45 @@ def _has_cf_compliant_time(
...
+
+    Returns
+    -------
+    str
+        Returns the first path from a list of paths.
+    """
+    # FIXME: This function should throw an exception if the first file
+    # is not a supported type.
+    # FIXME: The `type: ignore` comments should be removed after properly
+    # handling the types.
```
tomvothecoder, line 508: I added a few FIXME to `_get_first_path()`. The first FIXME is a bit tricky based on the nesting of elements, so I deferred it to later.

tomvothecoder (commented):
My first attempt at fixing #309.

I will open up a separate issue to ensure we address the FIXME comments and avoid tech debt.

tests/test_dataset.py
tomvothecoder: I'm taking a closer look at this since the test is failing in the build, and the branch wasn't rebased on the latest `main` with #302.

tomvothecoder: After fixing the `expected` dataset, it still always passes even without `_keep_time_encoding()` called. 

The FIXME: comment is still valid.

pochedls (commented):
@tomvothecoder - I reviewed this PR and the fix for the mfdataset encoding works great. 

Since the fix is only a couple lines, I wonder if it makes sense to go with [Option 2](https://github.com/xCDAT/xcdat/pull/309/files#r943756817) and just fix this in the `xcdat.open_mfdataset` function? It might make sense to keep it in `_postprocess_dataset()` if there were instances where the encoding wasn't handled correctly during an `xcdat.open_dataset`, but I don't think this should happen. 

I don't view this as an urgent PR – so it might also make sense to fix the broken test (which looks like a surprising amount of work?).

tomvothecoder:
> Since the fix is only a couple lines, I wonder if it makes sense to go with [Option 2](https://github.com/xCDAT/xcdat/pull/309/files#r943756817) and just fix this in the xcdat.open_mfdataset function? It might make sense to keep it in _postprocess_dataset() if there were instances where the encoding wasn't handled correctly during an xcdat.open_dataset, but I don't think this should 

Thanks for the input. I agree, `_postprocess_dataset()` is a reusable function so the logic for time encoding should be coupled to `open_mfdataset()` where it is needed. I just pushed a commit to follow option 2.

> I don't view this as an urgent PR – so it might also make sense to fix the broken test (which > looks like a surprising amount of work?).

I marked this as critical because it `TemporalAccessor` requires `.encoding` attrs (specifically "calendar"), which is not properly set with `xr.open_mfdataset()`. 

A possible quick fix to the test is to have two minimal, dummy `.nc` dataset with just a time axis and coordinates. It probably needs to be generated outside of xarray. I'll figure this out, and then merge this PR.

tests/test_dataset.py
```diff
@@ -306,36 +306,32 @@ def setUp(self, tmp_path):
...
+
+        # Create another dataset that extends the time coordinates by 1 value,
+        # to mimic a multifile dataset.
         ds2 = generate_dataset(cf_compliant=True, has_bounds=True)
-        ds2 = ds2.rename_vars({"ts": "tas"})
+        ds2 = ds2.isel(dict(time=slice(0, 1)))
+        ds2["time"].values[:] = np.array(
+            ["2002-01-16T12:00:00.000000000"],
+            dtype="datetime64[ns]",
+        )
```
tomvothecoder: My theory about xarray generated datasets keeping time encoding attributes was wrong. Xarray drops time encoding attributes if _coordinates need to merged_. If datasets share the same coordinates, no merging needs to be performed so the time encoding attributes are maintained, which was happening in the old version of this test).

I fixed this test by attempting to open two datasets, with the second being an extension of the first by 1 time coordinate.

xcdat/dataset.py
```diff
@@ -232,6 +232,8 @@ def open_mfdataset(
     if time_encoding is not None:
         time_dim = get_axis_dim(ds, "T")
         ds[time_dim].encoding = time_encoding
+        # Update "original_shape" to reflect the final time coordinates shape.
+        ds[time_dim].encoding["original_shape"] = ds[time_dim].shape
```
tomvothecoder, line 236: Use the "original_shape" of the final merged time coordinates.

xcdat/dataset.py
```diff
@@ -448,9 +450,11 @@ def _keep_time_encoding(paths: Paths) -> Dict[Hashable, Any]:
     # FIXME: Remove `type: ignore` comment after properly handling the type
     # annotations in `_get_first_path()`.
     ds = open_dataset(first_path, decode_times=True, add_bounds=False)  # type: ignore
-
     time_coord = get_axis_coord(ds, "T")
 
+    time_encoding = time_coord.encoding
+    time_encoding["source"] = paths
```
tomvothecoder, line 456: Set "source" to the `paths` arg.

tomvothecoder (commented):
@pochedls, I managed to fix the test and added additional improvements. 

I will merge this PR now.

PR #320, opened 2022-08-18
fix incorrect results for group averaging with missing data

### LINKED ISSUE #319: [Bug]: incorrect group averaging with missing data
### What happened?

I took the annual average of some data. In one example grid cell, the values were:

> [  nan -0.02 -0.02   nan   nan   nan   nan   nan   nan   nan   nan   nan] 

Since two months have the same value and data is missing for all other months, the weighted average should be `-0.02`. Instead, I get `-0.0032786885`; `group_average` seems to be incorrectly excluding the missing data.

I expected to get `-0.02`.

Load some data for an example grid cell:

```python
import xcdat
# open dataset
fn = '/p/user_pub/climate_work/pochedley1/surface/gistemp1200_GHCNv4_ERSSTv5.nc' 
ds = xcdat.open_dataset(fn)
# fix missing calendar
time_encoding = ds.time.encoding
time_encoding['calendar'] = 'standard'
ds.time.encoding = time_encoding
# select grid cell in subsetted dataset
dss = ds.isel(lat=[11], lon=[0])
```

Note that first year of data has two months (of twelve) with data values

```
print(dss.tempanomaly[0:12, 0, 0].values)
```

> [  nan -0.02 -0.02   nan   nan   nan   nan   nan   nan   nan   nan   nan] 

Taking the annual average I get an unexpected result:

```

ds_yearly = dss.temporal.group_average('tempanomaly', freq='year', weighted=True)

print(ds_yearly.tempanomaly[0, 0, 0].values)
```

> -0.0032786885

This is not the right value. Since there are two values in the first year (and they are the same), then the average should simply be `-0.02`. 

I think I see what is happening. Each month is assigned some weight in the first year (proportional to the number of days in each month): 

```
w = dss.temporal._get_weights()
```

If we consider just the first year, then:

```
w1 = w[0:12]
t1 = dss.tempanomaly[0:12, 0, 0]
print(w1.values, t1.values)
```

> [0.08469945 0.07923497 0.08469945 0.08196721 0.08469945 0.08196721                                                                             
>  0.08469945 0.08469945 0.08196721 0.08469945 0.08196721 0.08469945]
> 
> [  nan -0.02 -0.02   nan   nan   nan   nan   nan   nan   nan   nan   nan]

A simple weighted average is `WA = sum(T*W)/sum(W)`

```
print((np.sum(w1*t1)/np.sum(w1)).values)
```

> -0.0032786884513057645 

The problem is that there should be no weight assigned if there is no data for a given month/index. The weights should be corrected to reflect this: 

> [0.0 0.07923497 0.08469945 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0]

This weighting matrix would yield `-0.02` (which is the correct answer).

### Relevant log output

_No response_

I think this is probably handled correctly in [`_average`](https://github.com/xCDAT/xcdat/blob/main/xcdat/temporal.py#L897) (via xarray `.mean`), but [`_group_average`](https://github.com/xCDAT/xcdat/blob/main/xcdat/temporal.py#L923) is calling `.sum` instead of `.mean`. 

### Environment

main branch

pochedls:
## Description

This PR fixes an issue that leads to incorrect results for group averaging with missing data; the PR ensures that these data points receive no weight when averaging.

- Closes #319

## Checklist

If applicable:

### DIFF
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -418,6 +418,7 @@ class TestGroupAverage:
             ),
             coords={"time": self.ds.time, "lat": self.ds.lat, "lon": self.ds.lon},
             dims=["time", "lat", "lon"],
+            attrs={"test_attr": "test"},
         )
 
     def test_weighted_annual_averages(self):
@@ -458,6 +459,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "year",
@@ -467,6 +469,7 @@ class TestGroupAverage:
 
         xr.testing.assert_allclose(result, expected)
         assert result.ts.attrs == expected.ts.attrs
+        assert result.time.attrs == expected.time.attrs
 
     @requires_dask
     def test_weighted_annual_averages_with_chunking(self):
@@ -507,6 +510,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "year",
@@ -516,6 +520,7 @@ class TestGroupAverage:
 
         xr.testing.assert_allclose(result, expected)
         assert result.ts.attrs == expected.ts.attrs
+        assert result.time.attrs == expected.time.attrs
 
     def test_weighted_seasonal_averages_with_DJF_and_drop_incomplete_seasons(self):
         ds = self.ds.copy()
@@ -555,6 +560,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "season",
@@ -605,6 +611,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "season",
@@ -664,6 +671,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "season",
@@ -716,6 +724,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "season",
@@ -807,6 +816,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "month",
@@ -824,6 +834,7 @@ class TestGroupAverage:
             ),
             coords={"time": self.ds.time, "lat": self.ds.lat, "lon": self.ds.lon},
             dims=["time", "lat", "lon"],
+            attrs={"test_attr": "test"},
         )
 
         result = ds.temporal.group_average("ts", "month")
@@ -856,6 +867,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "month",
@@ -898,6 +910,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "day",
@@ -941,6 +954,7 @@ class TestGroupAverage:
             },
             dims=["time", "lat", "lon"],
             attrs={
+                "test_attr": "test",
                 "operation": "temporal_avg",
                 "mode": "group_average",
                 "freq": "hour",
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -831,7 +831,7 @@ class TestGroupAverage:
         expected = expected.drop_dims("time")
         expected["ts"] = xr.DataArray(
             name="ts",
-            data=np.array([[[2.0]], [[0.0]], [[1.0]], [[1.0]], [[2.0]]]),
+            data=np.array([[[2.0]], [[np.nan]], [[1.0]], [[1.0]], [[2.0]]]),
             coords={
                 "lat": expected.lat,
                 "lon": expected.lon,
--- a/xcdat/temporal.py
+++ b/xcdat/temporal.py
@@ -919,10 +919,28 @@ class TemporalAccessor:
 
         if self._weighted:
             self._weights = self._get_weights()
+            # grab metadata
+            dv_attrs = dv.attrs
+            dv_name = dv.name
+            # grab metadata (populated after calling ._group_data())
+            lt_attrs = self._labeled_time.attrs
+            lt_encoding = self._labeled_time.encoding
+            # weight the data variable
             dv *= self._weights
-            dv = self._group_data(dv).sum()
+            # cast the weights to match the dv.shape / dims
+            weights, x = xr.broadcast(self._weights, dv)
+            # ensure missing data receives no weight
+            weights = xr.where(np.isnan(dv), 0.0, weights)
+            # perform weighted average
+            dv = self._group_data(dv).sum() / self._group_data(weights).sum()
+            # add dv attributes
+            dv.attrs = dv_attrs
+            dv.name = dv_name
         else:
             dv = self._group_data(dv).mean()
+            # grab metadata (populated after calling ._group_data())
+            lt_attrs = self._labeled_time.attrs
+            lt_encoding = self._labeled_time.encoding
 
         # After grouping and aggregating the data variable values, the
         # original time dimension is replaced with the grouped time dimension.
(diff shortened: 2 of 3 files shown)


tests/test_temporal.py
```diff
@@ -831,7 +831,7 @@ def test_weighted_monthly_averages_with_masked_data(self):
         expected = expected.drop_dims("time")
         expected["ts"] = xr.DataArray(
             name="ts",
-            data=np.array([[[2.0]], [[0.0]], [[1.0]], [[1.0]], [[2.0]]]),
+            data=np.array([[[2.0]], [[np.nan]], [[1.0]], [[1.0]], [[2.0]]]),
```
pochedls, line 834: The value that was `0.0` is taking a monthly average for a month with one data point, which is `np.nan`. The average of one `np.nan` value should be `np.nan`.

xcdat/temporal.py
```diff
@@ -919,10 +919,28 @@ def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
 
         if self._weighted:
             self._weights = self._get_weights()
+            # grab metadata
+            dv_attrs = dv.attrs
+            dv_name = dv.name
```
pochedls, line 924: When I updated `dv = self._group_data(dv).sum()` to `dv = self._group_data(dv).sum() / self._group_data(weights).sum()` these attributes were dropped, so I held onto them to put them back on.

xcdat/temporal.py
```diff
@@ -919,10 +919,28 @@ def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
...
+            # grab metadata (populated after calling ._group_data())
+            lt_attrs = self._labeled_time.attrs
+            lt_encoding = self._labeled_time.encoding
+            # weight the data variable
             dv *= self._weights
-            dv = self._group_data(dv).sum()
+            # cast the weights to match the dv.shape / dims
+            weights, x = xr.broadcast(self._weights, dv)
```
pochedls, line 931: `weights` is one-dimensional (with a temporal dimension), but I want to make sure that grid cells that are `np.nan` in `dv` receive zero weight. Since `dv` is multidimensional, I need my `weights` array to have the same shape in order to reflect this.

xcdat/temporal.py
```diff
@@ -919,10 +919,28 @@ def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
...
+            # ensure missing data receives no weight
+            weights = xr.where(np.isnan(dv), 0.0, weights)
```
pochedls, line 933: Zero out the weights that correspond to `dv` cells with `np.nan` values.

xcdat/temporal.py
```diff
@@ -919,10 +919,28 @@ def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
...
+            # perform weighted average
+            dv = self._group_data(dv).sum() / self._group_data(weights).sum()
```
pochedls, line 935: New weighted average corresponding to `WA = sum(data*weights) / sum(weights)`. The denominator wasn't included before because it was always one. This is not true when there is missing data, so we need to include the denominator.

xcdat/temporal.py
```diff
@@ -936,8 +954,8 @@ def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
         # attributes are removed. Xarray's `keep_attrs=True` option only keeps
         # attributes for data variables and not their coordinates, so the
         # coordinate attributes have to be restored manually.
-        dv[self._dim].attrs = self._labeled_time.attrs
-        dv[self._dim].encoding = self._labeled_time.encoding
+        dv[self._dim].attrs = lt_attrs
+        dv[self._dim].encoding = lt_encoding
```
pochedls, line 958: These lines used to be 

```
dv[self._dim].attrs = self._labeled_time.attrs
dv[self._dim].encoding = self._labeled_time.encoding
```

Somehow this information was getting dropped in the new weighted average call (`dv = self._group_data(dv).sum() / self._group_data(weights).sum()`). 

I'm not sure if this can be improved...I store `lt_attrs` and `lt_encoding` in two places: under `if self._weighted:` and `else` because these values are only populated after `._group_data()` (in the `else` statement these values are populated after calling `_get_weights()` because `_get_weights()` calls `_group_data()`.

tomvothecoder:
Thanks for opening this PR. FYI, I realized I won't have time to review this until next week.

xcdat/temporal.py
```diff
@@ -1037,7 +1034,12 @@ def _group_data(self, data_var: xr.DataArray) -> DataArrayGroupBy:
         if self._mode == "average":
             dv_gb = dv.groupby(f"{self._dim}.{self._freq}")
         else:
-            self._labeled_time = self._label_time_coords(dv[self._dim])
+            # Reuse the labeled time coordinates if they are already generated.
+            # This specifically applies when generating weights, which calls
+            # this method again after it has already been called to group
+            # the data variable with the same time coordinates.
+            if not hasattr(self, "_labeled_time"):
+                self._labeled_time = self._label_time_coords(dv[self._dim])
```
tomvothecoder, line 1042: In one of your previous comments, you mentioned that the `self._labeled_time` attrs were being dropped.

This was because we call `_group_data()` twice, one for `dv` and one for `weights`. 
`weights` doesn't preserve the time dimension attrs because it is unnecessary. However, this caused a problem where the second call to `_group_data()` with `weights` overwrites `self._labeled_time` (resulting in attrs being dropped).

This new conditional fixes resolves this problem.

xcdat/temporal.py
```diff
@@ -919,43 +919,40 @@ def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
...
         dv = dv.rename({self._labeled_time.name: self._dim})
 
         # After grouping and aggregating, the grouped time dimension's
         # attributes are removed. Xarray's `keep_attrs=True` option only keeps
         # attributes for data variables and not their coordinates, so the
         # coordinate attributes have to be restored manually.
-        dv[self._dim].attrs = lt_attrs
-        dv[self._dim].encoding = lt_encoding
+        dv[self._dim].attrs = self._labeled_time.attrs
+        dv[self._dim].encoding = self._labeled_time.encoding
```
tomvothecoder, line 955: We can restore these lines of code since `self._labeled_time` attrs are preserved with the new conditional in `_group_data()`.

xcdat/temporal.py
```diff
@@ -919,43 +919,40 @@ def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
...
-            dv = self._group_data(dv).sum() / self._group_data(weights).sum()
-            # add dv attributes
-            dv.attrs = dv_attrs
-            dv.name = dv_name
+
+            # Perform weighted average using the formula
+            # WA = sum(data*weights) / sum(weights). The denominator must be
+            # included to take into account zero weight for missing data.
+            with xr.set_options(keep_attrs=True):
+                dv = self._group_data(dv).sum() / self._group_data(weights).sum()
```
tomvothecoder, line 936: `xr.set_options(keep_attrs=True)` will preserve the attrs for the `dv`. It does not restore the attrs for the time dimension, so we manually restore those in lines of code below.

xcdat/temporal.py
```diff
@@ -919,43 +919,40 @@ def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
...
-            lt_attrs = self._labeled_time.attrs
-            lt_encoding = self._labeled_time.encoding
 
         # After grouping and aggregating the data variable values, the
         # original time dimension is replaced with the grouped time dimension.
         # For example, grouping on "year_season" replaces the time dimension
         # with "year_season". This dimension needs to be renamed back to
         # the original time dimension name before the data variable is added
-        # back to the dataset so that the CF compliant name is maintained.
+        # back to the dataset so that the original name is preserved.
```
tomvothecoder, line 947: Updated this comment since the name of the time dimension isn't always necessarily CF compliant.

tests/test_temporal.py
```diff
@@ -418,6 +418,7 @@ def setup(self):
             ),
             coords={"time": self.ds.time, "lat": self.ds.lat, "lon": self.ds.lon},
             dims=["time", "lat", "lon"],
+            attrs={"test_attr": "test"},
```
tomvothecoder, line 421: Now test that the attrs for `ts` is preserved.

tests/test_temporal.py
```diff
@@ -467,6 +469,7 @@ def test_weighted_annual_averages(self):
 
         xr.testing.assert_allclose(result, expected)
         assert result.ts.attrs == expected.ts.attrs
+        assert result.time.attrs == expected.time.attrs
```
tomvothecoder, line 472: Also check the time attrs are preserved.

xcdat/temporal.py
```diff
@@ -919,43 +919,40 @@ def _group_average(self, data_var: xr.DataArray) -> xr.DataArray:
...
             dv *= self._weights
-            # cast the weights to match the dv.shape / dims
-            weights, x = xr.broadcast(self._weights, dv)
-            # ensure missing data receives no weight
+
+            # Ensure missing data (`np.nan`) receives no weight (zero). To
+            # achieve this, first broadcast the one-dimensional (temporal
+            # dimension) shape of the `weights` DataArray to the
+            # multi-dimensional shape of its corresponding data variable.
+            weights, _ = xr.broadcast(self._weights, dv)
```
tomvothecoder, line 929: Replace the unused `x` variable with `_`.

tomvothecoder (approved):
Hi Steve, I did some minor refactoring and updated the code comments based on your PR comments.

The tests pass, so you can merge if you are in agreement with the changes.

Thanks!

xcdat/temporal.py
tomvothecoder: Refactored this in my recent review

xcdat/temporal.py
tomvothecoder: Incorporated this PR comment in the code comment,

xcdat/temporal.py
tomvothecoder: Incorporated this PR comment in the code comment.

xcdat/temporal.py
tomvothecoder: Incorporated this PR comment in the code comment.

Issue #338, opened 2022-09-08
[Feature]: Handle "out of order" bounds


pochedls:
### Is your feature request related to a problem?

# Example

```python
# get the file
wget https://www.metoffice.gov.uk/hadobs/hadisst/data/HadISST_sst.nc.gz
gunzip HadISST_sst.nc.gz
ipython

fn = 'HadISST_sst.nc'
ds = xc.open_dataset(fn)
ds.spatial.average('sst')
```

> ValueError: The bounds have unexpected ordering. A lower bound has a greater value than an upper bound.

In general it seems that the first bound should be less than the second bound `latitude_bnds[:, 0] < latitude_bnds[:, 1]` (see image below), but this may only be for [bounds that are increasing](https://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries):

> For a coordinate variable such as lat(lat) with associated boundary variable latbnd(x,2), the interval endpoints must be ordered consistently with the associated coordinate, e.g., for an increasing coordinate, lat(1) > lat(0) implies latbnd(i,1) >= latbnd(i,0) for all i ([Figure 7.1](https://cfconventions.org/cf-conventions/cf-conventions.html#img-bnd_1d_coords)).

![CF Convention Cell Boundary Schemaric](https://cfconventions.org/cf-conventions/images/order_horizontal_bounds__1D_coord_variables.png)

We should therefore try to support bounds in which `latitude_bnds[:, 0] > latitude_bnds[:, 1]`.

### Describe the solution you'd like

I think we could likely create a copy of the domain bounds and simply reverse the order for the purpose of calculating weights, though some analysis would need to be done to determine which function should implement this. 

- [x] Analyze existing spatial averaging to determine best place to implement "switch"
- [x] Ensure this does not have adverse effects by: checking logic (e.g., `_scale_domain_to_region`)
- [x] Ensure this does not have adverse effects by: implementing new unit tests 
- [x] Ensure this does not have adverse effects by: performing spatial averages with real data

tomvothecoder:
I noticed this didn't close with #340 because it was marked with "Closes #324, #338" instead of two separate Closes statement. A quirk of GitHub.

tomvothecoder:
I created https://github.com/xCDAT/xcdat/issues/743 since this issue is related to spatial averaging.

PR #340, opened 2022-09-10
Fix spatial bugs: handle datasets with domain bounds out of order and zonal averaging

### LINKED ISSUE #324: [Bug]: averaging a lat-plev grid fails
### What happened?

I was looking at a `[time, plev, lat]` data and wanted to create a global average time series (averaging over `lat`). Spatial averaging fails because the dataset is missing the `X` coordinate.

The spatial average should be computed – the `X` dimension was not specified (so it should not be needed). 

```python
import xcdat as xc

fn = '/p/css03/esgf_publish/CMIP6/CMIP/CNRM-CERFACS/CNRM-CM6-1/historical/r1i1p1f2/AERmonZ/ta/grz/v20180917/ta_AERmonZ_CNRM-CM6-1_historical_r1i1p1f2_grz_185001-201412.nc'
ds = xc.open_dataset(fn)
ds.spatial.average('ta', axis=['Y'])
```


### Relevant log output

```python
KeyError                                  Traceback (most recent call last)                                                                                                                        
Input In [4], in <cell line: 1>()                                                                                                                                                                  
----> 1 ds.spatial.average('ta', axis=['Y'])                                                                                                                                                       
                                                                                                                                                                                                   
File ~/code/xcdat/xcdat/spatial.py:190, in SpatialAccessor.average(self, data_var, axis, weights, keep_weights, lat_bounds, lon_bounds)                                                            
    188     if lon_bounds is not None:                                                                                                                                                             
    189         self._validate_region_bounds("X", lon_bounds)                                                                                                                                      
--> 190     self._weights = self.get_weights(axis, lat_bounds, lon_bounds)                                                                                                                         
    191 elif isinstance(weights, xr.DataArray):                                                                                                                                                    
    192     self._weights = weights                                                                                                                                                                
                                                                                                                                                                                                   
File ~/code/xcdat/xcdat/spatial.py:252, in SpatialAccessor.get_weights(self, axis, lat_bounds, lon_bounds)                                                                                         
    208 """                                                                                                                                                                                        
    209 Get area weights for specified axis keys and an optional target domain.                                                                                                                    
    210                                                                                                                                                                                            
   (...)                                                                                                                                                                                           
    244 and pressure).
    245 """
    246 Bounds = TypedDict(
    247     "Bounds",
    248     {"domain": xr.DataArray, "region": Optional[RegionAxisBounds]},
    249 )
    250 axis_bounds: Dict[SpatialAxis, Bounds] = {
    251     "X": {
--> 252         "domain": self._dataset.bounds.get_bounds("X").copy(),
    253         "region": lon_bounds,
    254     },
    255     "Y": {
    256         "domain": self._dataset.bounds.get_bounds("Y").copy(),
    257         "region": lat_bounds,
    258     },
    259 }
    261 axis_weights: AxisWeights = {}
    263 for key in axis:

File ~/code/xcdat/xcdat/bounds.py:195, in BoundsAccessor.get_bounds(self, axis)
    171 def get_bounds(self, axis: CFAxisName) -> xr.DataArray:
    172     """Get bounds for axis coordinates if both exist.
    173 
    174     Parameters
   (...)
    193         If the coordinate bounds were not found for the ``axis``.
    194     """
--> 195     self._validate_axis_arg(axis)
    196     coord_var = get_axis_coord(self._dataset, axis)
    198     try:

File ~/code/xcdat/xcdat/bounds.py:385, in BoundsAccessor._validate_axis_arg(self, axis)
    380     keys = ", ".join(f"'{key}'" for key in cf_axis_names)
    381     raise ValueError(
    382         f"Incorrect 'axis' argument value. Supported values include {keys}."
    383     )
--> 385 get_axis_coord(self._dataset, axis)

File ~/code/xcdat/xcdat/axis.py:81, in get_axis_coord(obj, axis)
     78         pass
     80 if coord_var is None:
---> 81     raise KeyError(
     82         f"A coordinate variable for the {axis} axis was not found. Make sure "
     83         "the coordinate variable exists and either the (1) 'axis' attr or (2) "
     84         "'standard_name' attr is set, or (3) the dimension name follows the "
     85         "short-hand convention (e.g.,'lat')."
     86     )
     87 return coord_var

KeyError: "A coordinate variable for the X axis was not found. Make sure the coordinate variable exists and either the (1) 'axis' attr or (2) 'standard_name' attr is set, or (3) the dimension name follows the short-hand convention (e.g.,'lat')."
```


We probably just need to modify the portion of `get_weights` that [gets the bounds](https://github.com/xCDAT/xcdat/blob/main/xcdat/spatial.py#L250) for `X` and `Y` to only get those bounds if they are needed. Or maybe this dictionary isn't needed and the [logic can be put in this loop](https://github.com/xCDAT/xcdat/blob/main/xcdat/spatial.py#L263) over the user-specified axes.

### Environment

main branch

pochedls:
## Description
This PR is intended to fix #324 and #338. It allows users to perform spatial averaging on datasets in which `domain_bounds[:, 0] > domain_bounds[:, 1]` by re-ordering the domain bounds for the purposes of calculating spatial weights. It also fixes an issue that prevented spatial averaging on lat-plev grids. A section of the code tried to load longitude bounds even when that axis was not specified for averaging; this section of code is updated to only load bounds information if the axis is specified for averaging. 

- Closes #324, #338

## Checklist

If applicable:

### DIFF
--- a/tests/test_spatial.py
+++ b/tests/test_spatial.py
@@ -254,12 +254,26 @@ class TestGetWeights:
     def setup(self):
         self.ds = generate_dataset(cf_compliant=True, has_bounds=True)
 
-    def test_raises_error_if_domain_lower_bound_exceeds_upper_bound(self):
-        ds = self.ds.copy()
-        ds.lon_bnds.data[:] = np.array([[1, 0], [1, 2], [2, 3], [3, 4]])
+    def test_bounds_reordered_when_upper_indexed_first(self):
+        domain_bounds = xr.DataArray(
+            name="lon_bnds",
+            data=np.array(
+                [[-89.375, -90], [0.0, -89.375], [0.0, 89.375], [89.375, 90]]
+            ),
+            coords={"lat": self.ds.lat},
+            dims=["lat", "bnds"],
+        )
+        result = self.ds.spatial._force_domain_order_low_to_high(domain_bounds)
 
-        with pytest.raises(ValueError):
-            ds.spatial.get_weights(axis=["X"])
+        expected_domain_bounds = xr.DataArray(
+            name="lon_bnds",
+            data=np.array(
+                [[-90, -89.375], [-89.375, 0.0], [0.0, 89.375], [89.375, 90]]
+            ),
+            coords={"lat": self.ds.lat},
+            dims=["lat", "bnds"],
+        )
+        assert result.identical(expected_domain_bounds)
 
     def test_weights_for_region_in_lat_and_lon_domains(self):
         result = self.ds.spatial.get_weights(
--- a/xcdat/spatial.py
+++ b/xcdat/spatial.py
@@ -1,6 +1,7 @@
 """Module containing geospatial averaging functions."""
 from functools import reduce
 from typing import (
+    Callable,
     Dict,
     Hashable,
     List,
@@ -244,42 +245,36 @@ class SpatialAccessor:
         and pressure).
         """
         Bounds = TypedDict(
-            "Bounds",
-            {"domain": xr.DataArray, "region": Optional[RegionAxisBounds]},
+            "Bounds", {"weights_method": Callable, "region": Optional[np.ndarray]}
         )
-        axis_bounds: Dict[SpatialAxis, Bounds] = {}
-        if "X" in axis:
-            axis_bounds["X"] = {
-                "domain": self._dataset.bounds.get_bounds("X").copy(),
-                "region": lon_bounds,
-            }
-        if "Y" in axis:
-            axis_bounds["Y"] = {
-                "domain": self._dataset.bounds.get_bounds("Y").copy(),
-                "region": lat_bounds,
-            }
+
+        axis_bounds: Dict[SpatialAxis, Bounds] = {
+            "X": {
+                "weights_method": self._get_longitude_weights,
+                "region": np.array(lon_bounds, dtype="float")
+                if lon_bounds is not None
+                else None,
+            },
+            "Y": {
+                "weights_method": self._get_latitude_weights,
+                "region": np.array(lat_bounds, dtype="float")
+                if lat_bounds is not None
+                else None,
+            },
+        }
 
         axis_weights: AxisWeights = {}
-
         for key in axis:
-            d_bounds = axis_bounds[key]["domain"]
-            # the logic for generating longitude weights depends on the
-            # bounds being ordered such that d_bounds[:, 0] < d_bounds[:, 1]
-            # they are re-ordered (if need be) for the purpose of creating
-            # weights
+            d_bounds = self._dataset.bounds.get_bounds(key).copy()
+            # The logic for generating longitude weights depends on the
+            # bounds being ordered such that d_bounds[:, 0] < d_bounds[:, 1].
+            # They are re-ordered (if need be) for the purpose of creating
+            # weights.
             d_bounds = self._force_domain_order_low_to_high(d_bounds)
 
-            r_bounds: Union[Optional[RegionAxisBounds], np.ndarray] = axis_bounds[key][
-                "region"
-            ]
-            if r_bounds is not None:
-                r_bounds = np.array(r_bounds, dtype="float")
+            r_bounds = axis_bounds[key]["region"]
 
-            if key == "X":
-                weights = self._get_longitude_weights(d_bounds, r_bounds)
-            elif key == "Y":
-                weights = self._get_latitude_weights(d_bounds, r_bounds)
-
+            weights = axis_bounds[key]["weights_method"](d_bounds, r_bounds)
             weights.attrs = d_bounds.attrs
             axis_weights[key] = weights
 
@@ -315,22 +310,28 @@ class SpatialAccessor:
             get_axis_coord(self._dataset, key)
 
     def _force_domain_order_low_to_high(self, domain_bounds: xr.DataArray):
-        """Re-orders the ``domain_bounds`` to be ordered low-to-high such that
-        domain_bounds[:, 0] < domain_bounds[:, 1].
+        """Reorders the ``domain_bounds`` low-to-high.
+
+        This method ensures all lower bound values are less than the upper bound
+        values (``domain_bounds[:, 1] < domain_bounds[:, 1]``).
+
         Parameters
         ----------
         domain_bounds: xr.DataArray
             The bounds of an axis.
+
         Returns
         ------
         xr.DataArray
             The bounds of an axis (re-ordered if applicable).
         """
         index_bad_cells = np.where(domain_bounds[:, 1] - domain_bounds[:, 0] < 0)[0]
+
         if len(index_bad_cells) > 0:
             new_domain_bounds = domain_bounds.copy()
             new_domain_bounds[index_bad_cells, 0] = domain_bounds[index_bad_cells, 1]
             new_domain_bounds[index_bad_cells, 1] = domain_bounds[index_bad_cells, 0]
+
             return new_domain_bounds
         else:
             return domain_bounds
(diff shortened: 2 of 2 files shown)


xcdat/spatial.py
```diff
@@ -247,22 +247,23 @@ def get_weights(
...
-            },
-            "Y": {
+            }
+        if "Y" in axis:
+            axis_bounds["Y"] = {
                 "domain": self._dataset.bounds.get_bounds("Y").copy(),
                 "region": lat_bounds,
-            },
-        }
+            }
```
pochedls, line 260: Previously, the `axis_bounds` dict tried to populate `X` and `Y` every time, but we do not need to populate the bounds information if we are not performing spatial averaging over the `X` or `Y` axis. This fixes #324. I think this was a bug and I didn't see a need to modify the unit tests in this case.

tests/test_spatial.py
```diff
@@ -254,12 +254,26 @@ class TestGetWeights:
...
-            ds.spatial.get_weights(axis=["X"])
+        expected_domain_bounds = xr.DataArray(
+            name="lon_bnds",
+            data=np.array(
+                [[-90, -89.375], [-89.375, 0.0], [0.0, 89.375], [89.375, 90]]
+            ),
+            coords={"lat": self.ds.lat},
+            dims=["lat", "bnds"],
+        )
+        assert result.identical(expected_domain_bounds)
```
pochedls, line 276: Previously, we threw an error if the lower domain bound was larger than the upper domain bound. Now we re-order them. This unit test checks to make sure that is done in `_force_domain_order_low_to_high`.

xcdat/spatial.py
```diff
@@ -309,25 +310,26 @@ def _validate_axis_arg(self, axis: List[SpatialAxis]):
...
-            raise ValueError(
-                "The bounds have unexpected ordering. A lower bound has a "
-                "greater value than an upper bound."
-            )
+            new_domain_bounds = domain_bounds.copy()
+            new_domain_bounds[index_bad_cells, 0] = domain_bounds[index_bad_cells, 1]
+            new_domain_bounds[index_bad_cells, 1] = domain_bounds[index_bad_cells, 0]
+            return new_domain_bounds
+        else:
+            return domain_bounds
```
pochedls, line 332: Previously, we threw an error if the lower domain bound was larger than the upper domain bound (in `_validate_domain_bounds`) This function is replaced with `_force_domain_order_low_to_high` which re-orders the bounds such that `domain_bound[:, 0] < domain_bounds[:, 1]`.

Note I had originally replaced [these lines](https://github.com/xCDAT/xcdat/blob/main/xcdat/spatial.py#L601-L602) with logic to simply determine, which of the bounds is higher versus lower (e.g., `domain_lowers = np.min(d_bounds, axis=1)`), but the order matters further down ([here](https://github.com/xCDAT/xcdat/blob/main/xcdat/spatial.py#L619)).

xcdat/spatial.py
```diff
@@ -359,7 +361,7 @@ def _validate_region_bounds(self, axis: SpatialAxis, bounds: RegionAxisBounds):
                 "upper bounds, (lower, upper)."
             )
 
-        if len(bounds) <= 0 or len(bounds) > 2:
+        if len(bounds) != 2:
```
pochedls, line 364: I randomly caught this... I think `<=0` doesn't really make any sense, right?

pochedls:
I think this is in good shape and can be reviewed. I do plan on re-running it over a bunch of cases before merging.

pochedls:
I re-did comparisons with CDAT - we still agree with CDAT in all cases where both can load the dataset (loop over 115 model files and 21 domains).

xcdat/spatial.py
tomvothecoder: I agree with this change, and `<=0` does not make sense (arrays can't have < 0 elements)

tests/test_spatial.py
tomvothecoder: Looks good

xcdat/spatial.py
tomvothecoder: Logic looks good. I did some minor refactoring and will push the code for your review.

xcdat/spatial.py
tomvothecoder: Looks good

tomvothecoder (commented):
Looks good overall, with some minor refactoring changes in my upcoming commit.

xcdat/spatial.py
```diff
@@ -244,42 +245,36 @@ def get_weights(
...
-
-            if key == "X":
-                weights = self._get_longitude_weights(d_bounds, r_bounds)
-            elif key == "Y":
-                weights = self._get_latitude_weights(d_bounds, r_bounds)
+            r_bounds = axis_bounds[key]["region"]
 
+            weights = axis_bounds[key]["weights_method"](d_bounds, r_bounds)
             weights.attrs = d_bounds.attrs
             axis_weights[key] = weights
```
tomvothecoder, line 279: I refactored a bit here to reduce `if else` statements and avoids fetching `domain_bounds` for unspecified keys in `axis`.

xcdat/spatial.py
```diff
@@ -315,22 +310,28 @@ def _validate_axis_arg(self, axis: List[SpatialAxis]):
...
         index_bad_cells = np.where(domain_bounds[:, 1] - domain_bounds[:, 0] < 0)[0]
+
         if len(index_bad_cells) > 0:
             new_domain_bounds = domain_bounds.copy()
             new_domain_bounds[index_bad_cells, 0] = domain_bounds[index_bad_cells, 1]
             new_domain_bounds[index_bad_cells, 1] = domain_bounds[index_bad_cells, 0]
+
             return new_domain_bounds
         else:
             return domain_bounds
```
tomvothecoder, line 337: Updated docstring and code spacing for consistency with code style

tomvothecoder (approved):
Hi @pochedls, please review my refactoring changes. The unit tests still pass.

If you agree with them, merge when ready.

xcdat/spatial.py
tomvothecoder: This commented was intended to highlight the entire method, not just this line.

xcdat/spatial.py
```diff
@@ -244,37 +245,36 @@ def get_weights(
...
+                if lon_bounds is not None
+                else None,
             },
             "Y": {
-                "domain": self._dataset.bounds.get_bounds("Y").copy(),
-                "region": lat_bounds,
+                "weights_method": self._get_latitude_weights,
+                "region": np.array(lat_bounds, dtype="float")
+                if lat_bounds is not None
+                else None,
```
pochedls, line 262: You learn something new every day.

tomvothecoder: Yup, they are called ternary operators (aka inline conditionals).
https://datagy.io/inline-if-in-python/

I don't use them too often for variable assignment, but it can be useful where it makes sense.

xcdat/spatial.py
```diff
@@ -309,25 +309,32 @@ def _validate_axis_arg(self, axis: List[SpatialAxis]):
...
-                "The bounds have unexpected ordering. A lower bound has a "
-                "greater value than an upper bound."
-            )
+            new_domain_bounds = domain_bounds.copy()
+            new_domain_bounds[index_bad_cells, 0] = domain_bounds[index_bad_cells, 1]
+            new_domain_bounds[index_bad_cells, 1] = domain_bounds[index_bad_cells, 0]
+
+            return new_domain_bounds
+        else:
+            return domain_bounds
```
tomvothecoder, line 337:
```suggestion

        return domain_bounds
```

Just saw that this can be simplified further (nitpick)

PR #344, opened 2022-09-13
modify preserve bounds logic

### LINKED ISSUE #316: [Bug]: regridder storing NaN for lat bounds
### What happened?

It seems like (in some instances) the horizontal regridder is saving latitude bounds with `NaN` values.

Since we are passing in a grid with valid `lat_bnds`, these should be saved to the regridded dataset. 

```python
# imports
from xcdat.regridder import grid
import xcdat as xc
import numpy as np

# example file
fn = '/p/user_pub/climate_work/pochedley1/cesm2le/tas/tas_Amon_CESM2_historical-ssp370_1251-005_gr_185001-210012.nc'

# create target grid
ngrid = grid.create_grid(np.arange(-88.75, 90, 2.5), np.arange(1.25, 360., 2.5))

# open file
ds = xc.open_dataset(fn)
dsr = ds.regridder.horizontal('tas', ngrid, tool='xesmf', method='bilinear', periodic=True)

# notice that spatial averaging produces NaNs
print(dsr.spatial.average('tas').tas.values)                                                                                          
```

> [nan nan nan ... nan nan nan]

```
# this likely results because bounds_lat has NaN values
print(dsr.bounds_lat)
```

> <xarray.DataArray 'bounds_lat' (lat: 72, bound: 2)>
> array([[nan, nan],
>        [nan, nan],
>        [nan, nan],
>        [nan, nan],
>        ...

### Relevant log output

_No response_

A separate `lat_bnds` variable is saved in the regridded dataset (it was not in the original dataset) with the correct bounds, but the `ds.lat.bounds` attribute is `bounds_lat`. 

This may be as simple as using that attribute to choose the name of the new bounds (if it exists) when saving the new, regridded dataset. 

For some reason this doesn't seem to affect longitude (in this example). `ds.bounds_lon` has the right shape and does not have NaN values. 

### Environment

main branch

pochedls:
## Description
This resolves an issue in which the bounds are populated by NaN values after regridding. The issue was related to functionality to preserve the source grid bounds in the target grid. Sometimes this functionality would result in multiple bounds (with different names) being saved for a single spatial axis (though I'm not sure why the values were NaN). 

- Closes #316
## Checklist

If applicable:

### DIFF
--- a/tests/test_regrid.py
+++ b/tests/test_regrid.py
@@ -641,16 +641,12 @@ class TestBase:
 
         output_ds = base.preserve_bounds(ds_with_bounds, ds_without_bounds, target)
 
-        assert "lat_bnds" in output_ds
-        assert "lon_bnds" in output_ds
         assert "time_bnds" in output_ds
 
         target = xr.Dataset()
 
         output_ds = base.preserve_bounds(ds_without_bounds, ds_without_bounds, target)
 
-        assert "lat_bnds" not in output_ds
-        assert "lon_bnds" not in output_ds
         assert "time_bnds" in output_ds
 
     def test_regridder_implementation(self):
--- a/tests/test_regrid.py
+++ b/tests/test_regrid.py
@@ -641,12 +641,16 @@ class TestBase:
 
         output_ds = base.preserve_bounds(ds_with_bounds, ds_without_bounds, target)
 
+        assert "lat_bnds" not in output_ds
+        assert "lon_bnds" not in output_ds
         assert "time_bnds" in output_ds
 
         target = xr.Dataset()
 
         output_ds = base.preserve_bounds(ds_without_bounds, ds_without_bounds, target)
 
+        assert "lat_bnds" not in output_ds
+        assert "lon_bnds" not in output_ds
         assert "time_bnds" in output_ds
 
     def test_regridder_implementation(self):
--- a/xcdat/regridder/base.py
+++ b/xcdat/regridder/base.py
@@ -10,11 +10,11 @@ logger = setup_custom_logger(__name__)
 
 
 def preserve_bounds(
-    source: xr.Dataset, source_grid: xr.Dataset, target: xr.Dataset
+    source: xr.Dataset, target_grid: xr.Dataset, target: xr.Dataset
 ) -> xr.Dataset:
     """Preserves bounds from sources to target.
 
-    Preserve the lat/lon bounds from `source_grid` to `target`.
+    Ensure the lat/lon bounds from `target_grid` are included in the `target` dataset.
 
     Preserve any additional bounds e.g. time, vertical from `source` to `target`.
 
@@ -22,8 +22,8 @@ def preserve_bounds(
     ----------
     source : xr.Dataset
         Source Dataset.
-    source_grid : xr.Dataset
-        Source grid Dataset.
+    target_grid : xr.Dataset
+        Target grid Dataset.
     target : xr.Dataset
         Target Dataset to preserve bounds to.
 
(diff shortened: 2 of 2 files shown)


xcdat/regridder/base.py
```diff
@@ -10,20 +10,20 @@
...
     Preserve any additional bounds e.g. time, vertical from `source` to `target`.
 
     Parameters
     ----------
     source : xr.Dataset
         Source Dataset.
-    source_grid : xr.Dataset
-        Source grid Dataset.
+    target_grid : xr.Dataset
+        Target grid Dataset.
```
pochedls, line 26: Did I rename things correctly here? I think we are passing in the target grid here (even though it is initially being used to "source" some bounds information).

xcdat/regridder/base.py
```diff
@@ -54,6 +54,9 @@ def preserve_bounds(
         else:
             if source_bnds.name in target:
                 logger.debug(f"Bounds {source_bnds.name!r} already present")
+            elif dim_name in ["X", "Y"]:
+                # the X / Y bounds are copied from the target grid above
+                continue
```
pochedls, line 59: I skip copying the `X` and `Y` bounds information, because I think they should always be populated in the logic above from the target grid. Is this an okay assumption? Regardless, I think the source grid horizontal bounds are no longer applicable?

tests/test_regrid.py
```diff
@@ -641,16 +641,12 @@ def test_preserve_bounds(self):
 
         output_ds = base.preserve_bounds(ds_with_bounds, ds_without_bounds, target)
 
-        assert "lat_bnds" in output_ds
-        assert "lon_bnds" in output_ds
```
pochedls, line 645: If my logic makes sense, then we should not expect horizontal grid bounds to be preserved.

pochedls (commented):
This is my quick assessment / fix, but this deserves review. I can also try to test it more broadly before we merge.

xcdat/regridder/base.py
pochedls: I guess one question here: can you regrid just `X` or `Y` (e.g., if I have [`time`, `lat`] datasets)? I think in this scenario you could still skip copying the source dataset `X` and `Y` bounds.

pochedls:
FYI – I did loop over a bunch of `tas` files and CDAT/xcdat agreed in every instance.

tomvothecoder:
@jasonb5 @pochedls, do you both think this will be merged today or tomorrow?
Just checking if it can make 0.3.2, which I am targeting by tomorrow.

jasonb5 (changes requested):

tests/test_regrid.py
pochedls: For some reason it says that @jasonb5 requested changes (I think because of this comment), but I think @jasonb5 made these changes in his last commit.

pochedls:
I think this can be merged upon approval from @jasonb5 and @tomvothecoder

tests/test_regrid.py
tomvothecoder: Jason marked requested changes in the review before making the commit, so it will still show up as requested changes until he marks approved.

Example of how to request changes from a review:

image

tomvothecoder (approved):
Looks good to me. Thanks for this PR and @jasonb5 for reviewing.

You can go ahead and merge.

jasonb5 (approved):

tests/test_regrid.py
jasonb5: @tomvothecoder @pochedls Yea I forgot to approve the changes after I made them. I'll get this merged in.

PR #385, opened 2022-11-09
Wrap open_dataset and open_mfdataset to flexibly open datasets

### LINKED ISSUE #128: Add `xcdat.open()` to wrap `xcdat.open_dataset()` and `xcdat.open_mfdataset()`
### Is your feature request related to a problem? If yes, Please describe

Suggesting a thin wrapper for opening files. Because cdscan-generated xml is still a major part of the PCMDI’s CMIP database, having the capability of open xmls in addition would help buying some time when transitioning PMP to use xcdat. 

### Describe your proposed solution

xcdat.open(input)
If input is string and .nc file, use xcdat.open_dataset
If input is list of .nc files, use xcdat.open_mfdataset
If input is string and .xml file, read xml (which is ascii file) and find where linked .nc files are, and depending if linked data file is single .nc file or multiple .nc files, use xcdat_opendataset or xcdat.open_mfdataset

### Additional context

@pochedls's comment:
"To query for CMIP paths, you need to import xrw (here) and call xrw.get_cmip_paths(). Alternatively, I think you could open an xml file following something like [this](https://gist.github.com/pochedls/5c331279f65e0383bfb506cc33e60c87). This is all “scratch” code…"

pochedls:
## Description

This PR is intended to wrap `xc.open_dataset()` and `xc.open_mfdataset()` such that xcdat can flexibly open the dataset path passed in (whether it is a directory, wildcard search, xml file, single file, etc.). 

- Closes #128

## Checklist

If applicable:

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -5,6 +5,7 @@ import cftime
 import numpy as np
 import pytest
 import xarray as xr
+from lxml import etree
 
 from tests.fixtures import generate_dataset
 from xcdat.dataset import (
@@ -585,10 +586,10 @@ class TestOpenMfDataset:
     @pytest.fixture(autouse=True)
     def setUp(self, tmp_path):
         # Create temporary directory to save files.
-        dir = tmp_path / "input_data"
-        dir.mkdir()
-        self.file_path1 = f"{dir}/file1.nc"
-        self.file_path2 = f"{dir}/file2.nc"
+        self.dir = tmp_path / "input_data"
+        self.dir.mkdir()
+        self.file_path1 = f"{self.dir}/file1.nc"
+        self.file_path2 = f"{self.dir}/file2.nc"
 
     def test_raises_warning_if_decode_times_but_no_time_coords_found(self, caplog):
         ds = generate_dataset(decode_times=False, cf_compliant=True, has_bounds=True)
@@ -624,6 +625,61 @@ class TestOpenMfDataset:
 
         expected = ds1.merge(ds2)
         assert result.identical(expected)
+
+    def test_raises_error_if_xml_does_not_have_root_direcory(self):
+        """
+        The structure of CDAT xml files (climate data markup language) is a dialect of xml files that have a defined set of attributes.
+        We only care that the xml file has a directory attribute – we can probably just return an error saying "the xml file provided does not have a root directory attribute" or something like that
+        XML probably isn't used that widely by the broad climate community, but it is used pretty widely by former CDAT users
+        """
+        ds1 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
+        ds1.to_netcdf(self.file_path1)
+        ds2 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
+        ds2 = ds2.rename_vars({"ts": "tas"})
+        ds2.to_netcdf(self.file_path2)
+
+        # Create the XML file
+        xml_path = f"{self.dir}/datasets.xml"
+        page = etree.Element("dataset")
+        doc = etree.ElementTree(page)
+        doc.write(xml_path, xml_declaration=True, encoding="utf-16")
+
+        with pytest.raises(KeyError):
+            open_mfdataset(xml_path, decode_times=True)
+
+    def test_opens_datasets_from_xml(self):
+        ds1 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
+        ds1.to_netcdf(self.file_path1)
+        ds2 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
+        ds2 = ds2.rename_vars({"ts": "tas"})
+        ds2.to_netcdf(self.file_path2)
+
+        # Create the XML file
+        xml_path = f"{self.dir}/datasets.xml"
+        page = etree.Element("dataset", directory=str(self.dir))
+        doc = etree.ElementTree(page)
+        doc.write(xml_path, xml_declaration=True, encoding="utf-16")
+
+        result = open_mfdataset(xml_path, decode_times=True)
+        expected = ds1.merge(ds2)
+
+        result.identical(expected)
+
+    def test_raises_error_if_directory_has_no_netcdf_files(self):
+        with pytest.raises(ValueError):
+            open_mfdataset(str(self.dir), decode_times=True)
+
+    def test_opens_netcdf_files_in_a_directory(self):
+        ds1 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
+        ds1.to_netcdf(self.file_path1)
+        ds2 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
+        ds2 = ds2.rename_vars({"ts": "tas"})
+        ds2.to_netcdf(self.file_path2)
+
+        result = open_mfdataset(str(self.dir), decode_times=True)
+        expected = ds1.merge(ds2)
+
+        result.identical(expected)
 
     def test_user_specified_callable_results_in_subsetting_dataset_on_time_slice(self):
         def callable(ds):
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -642,7 +642,7 @@ class TestOpenMfDataset:
         with pytest.raises(KeyError):
             open_mfdataset(xml_path, decode_times=True)
 
-    def test_opens_datasets_from_xml(self):
+    def test_opens_datasets_from_xml_using_str_path(self):
         ds1 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
         ds1.to_netcdf(self.file_path1)
         ds2 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
@@ -651,6 +651,24 @@ class TestOpenMfDataset:
 
         # Create the XML file
         xml_path = f"{self.dir}/datasets.xml"
+        page = etree.Element("dataset", directory=str(self.dir))
+        doc = etree.ElementTree(page)
+        doc.write(xml_path, xml_declaration=True, encoding="utf-16")
+
+        result = open_mfdataset(xml_path, decode_times=True)
+        expected = ds1.merge(ds2)
+
+        result.identical(expected)
+
+    def test_opens_datasets_from_xml_using_pathlib_path(self):
+        ds1 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
+        ds1.to_netcdf(self.file_path1)
+        ds2 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
+        ds2 = ds2.rename_vars({"ts": "tas"})
+        ds2.to_netcdf(self.file_path2)
+
+        # Create the XML file
+        xml_path = self.dir / "datasets.xml"
         page = etree.Element("dataset", directory=str(self.dir))
         doc = etree.ElementTree(page)
         doc.write(xml_path, xml_declaration=True, encoding="utf-16")
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -626,7 +626,7 @@ class TestOpenMfDataset:
         expected = ds1.merge(ds2)
         assert result.identical(expected)
 
-    def test_raises_error_if_xml_does_not_have_root_direcory(self):
+    def test_raises_error_if_xml_does_not_have_root_directory_attr(self):
         ds1 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
         ds1.to_netcdf(self.file_path1)
         ds2 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -627,11 +627,6 @@ class TestOpenMfDataset:
         assert result.identical(expected)
 
     def test_raises_error_if_xml_does_not_have_root_direcory(self):
-        """
-        The structure of CDAT xml files (climate data markup language) is a dialect of xml files that have a defined set of attributes.
-        We only care that the xml file has a directory attribute – we can probably just return an error saying "the xml file provided does not have a root directory attribute" or something like that
-        XML probably isn't used that widely by the broad climate community, but it is used pretty widely by former CDAT users
-        """
         ds1 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
         ds1.to_netcdf(self.file_path1)
         ds2 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
--- a/xcdat/dataset.py
+++ b/xcdat/dataset.py
@@ -1,7 +1,6 @@
 """Dataset module for functions related to an xarray.Dataset."""
 import os
 import pathlib
-import xml.etree.ElementTree as ET
 from datetime import datetime
 from functools import partial
 from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Union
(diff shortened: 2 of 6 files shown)


pochedls:
I created this *draft* PR in part to fine-tune the proposed API and so others could identify (and help fix?) issues with the approach in this PR. 

I'm wondering if we should just make `xcdat.open_dataset()` open directories, xml files, individual files, lists, etc. (rather than `xcdat.open()`, which is what is in the initial PR and what is outlined in Issue [#128](https://github.com/xCDAT/xcdat/issues/128). Or maybe it's better to distinguish the proposed functionality from `open_dataset` and `open_mfdataset` (that way you could always fall back to these foundational functions). 

There's not a lot of error handling in the inital PR and I assume there are gaps in the PR logic (and other cases to be considered aside from xml/directory/list/etc). Help identifying these would be useful. 

A big assumption in the current PR is that when the user passes in a directory, xcdat should try to open all files in the directory as a multi-file dataset. This gets around having to assume they are all `.nc` files, but if a directory has mixed files, this could be a problem.

lee1043:
@pochedls I like the idea of having `open` wrapper that can also open xml. I have drafted the similar [ad-hoc function in the PMP](https://github.com/PCMDI/pcmdi_metrics/blob/709ffeeb1a9d13f11c8b443a9495a6a70f0fcb15/pcmdi_metrics/io/xcdat_openxml.py#L9) because the current PMP workflow rely on xml. I was considering my function would be used for interim until the PMP workflow to adapt your `xsearch`. I like the way you coded that is looking great and solid. I'd happy to give a test when you feel this PR is ready.

lee1043:
@pochedls I confirm `open` warpper is working as expected. I tried with giving xml (str), directory path for netcdf files (str), list of netcdf files (list), and single netcdf file (str) to the `xcdat.open`. All work as expected! Thank you very much for adding this convenient wrapper. I am happy with the addition.

lee1043 (approved):

tomvothecoder:
Thanks for this PR. I'll comment once I take a closer look.

tomvothecoder:
Notes from meetings:
- Consider other file types
- HDFs might be tricky – embedded datasets into one another?

Tom will review PR 
- Naming of function (open vs. open_dataset/open_mfdataset())
- Potential inconsistency with xarray usage
- Make explicit that open wraps open_dataset()/open_mfdataset()
- Handling mixed file types, file types outside supported xarray

lee1043:
> Naming of function (open vs. open_dataset/open_mfdataset())

I think we can consider `open` as an addition to existing open_dataset/open_mfdataset, so "plus" rather than "vs".

xcdat/dataset.py
```diff
@@ -35,13 +37,97 @@
...
+    """
+
+    # inspect path attributes and use either open_dataset
+    # or open_mfdataset as appropriate
+    #   * list (open list of files with open_mfdataset)
+    #   * pathlib.Path (open pathlib.Path with open_mfdataset)
+    #   * deprecated CDAT xml file type
+    #   * directory (opens with open_dataset or open_mfdataset
+    #                depending in number of files)
+    #   * wildcard input (opens data with open_mfdataset)
```
tomvothecoder, line 61: From my understanding, the motivation behind adding a new `open` function is to provide an abstracted API that determines whether to use `open_dataset()`/`open_mfdataset()` based on the `path` argument to make opening datasets hopefully more intuitive. Additionally, `open` would include support for the deprecated CDAT XML format.

Let me know if this _isn't_ correct.

xcdat/dataset.py
```diff
@@ -35,13 +37,97 @@
...
+    """
+    Documentation to be written when this PR is
+    further along...
+    """
+
+    # inspect path attributes and use either open_dataset
+    # or open_mfdataset as appropriate
+    #   * list (open list of files with open_mfdataset)
+    #   * pathlib.Path (open pathlib.Path with open_mfdataset)
+    #   * deprecated CDAT xml file type
```
tomvothecoder, line 58: A few questions I had about supporting XML:
- I noticed in the docstring it says "deprecated CDAT xml file type". Is the structure of the XML formatted specifically by CDAT, and are there plans to eventually move away from it?
  - I think if users pass XML files that don't align with a specific format, then `open` will break. We'd have to include documentation that the XML file must follow a specific format (not sure if this is appropriate to support though since it isn't generalized like a list of paths).
- I'm not really familiar with XML usage. How common is the XML format used for reading datasets in the community?

xcdat/dataset.py
```diff
@@ -35,13 +37,97 @@
...
+    else:
+        ds = open_dataset(
+            path,
+            data_var=data_var,
+            add_bounds=add_bounds,
+            decode_times=decode_times,
+            center_times=center_times,
+            lon_orient=lon_orient,
+            **kwargs,
+        )
```
tomvothecoder, line 119: I noticed this logic for determining whether to use `open_dataset()` or `open_mfdataset()`.

**`open_mfdataset()` supports opening a single dataset file and multiple dataset files.** The main difference (that I'm aware of) with opening a single dataset with `open_dataset()` vs. `open_mfdataset()` is that data variables are automatically Dask arrays with `open_mfdataset()`. 

I'm not sure of pros/cons, implications (e.g. performance), etc. of using `open_dataset()` vs. `open_mfdataset()` for a single dataset file, but **I'd probably lean towards always using `open_mfdataset()` if you're looping over models or lots of directories.**

xcdat/dataset.py
pochedls: * The structure of CDAT xml files (climate data markup language) is a [dialect of xml files](https://cdms.readthedocs.io/en/latest/manual/cdms_6.html) that have a defined set of attributes. 
* We only care that the xml file has a `directory` attribute – we can probably just return an error saying "the xml file provided does not have a root directory attribute" or something like that
* XML probably isn't used that widely, but it is used pretty widely by former CDAT users

xcdat/dataset.py
pochedls: When I loop over models I currently use `open_mfdataset` universally: `xc.open_mfdataset(dpath + '*.nc')`. `xc.open_dataset()` used to be faster than `xc.open_mfdataset()`, though this may have been addressed a few months ago (I remember discussing this, perhaps as part of the large PR that changes the logic in dealing with coords/bounds and I think it may have been addressed). 

I just checked for one dataset and they were close (24 versus 21 ms), though I should look to see if both functions are close in speed across a broader range of datasets.

xcdat/dataset.py
tomvothecoder: Gotcha, this makes sense to me. I think being explicit about these details as you describe is a good idea.

xcdat/dataset.py
pochedls: I'd be fine with this, but ideally `open_mfdataset()` could also handle a string [`directory`](https://github.com/xCDAT/xcdat/blob/3913ea2cba91a24d099686072832cebf62ac4911/xcdat/dataset.py#L76). I think this would handle all cases in this PR. 

I think I personally would then just use `open_mfdataset()`. 

The case where `open()` would still be useful is if `open_dataset()` is faster than `open_mfdataset()`; `open()` would be faster  / better by intelligently using `open_mfdataset()` only when necessary.

tomvothecoder: > The case where `open()` would still be useful is if `open_dataset()` is faster than `open_mfdataset()`; `open()` would be faster / better by intelligently using `open_mfdataset()` only when necessary.

I found some xarray GH issues related to the slow performance of `open_mfdataset()`, which are **mostly be related to [coordinate compatibility checks when concatenating files](https://github.com/pydata/xarray/issues/1385#issuecomment-561920115) and [decoding time coordinates](https://github.com/pydata/xarray/issues/1385#issuecomment-370064483) with more than 1 dataset**. There are [workarounds](https://docs.xarray.dev/en/stable/user-guide/io.html#reading-multi-file-datasets) for users if they find `open_mfdataset()` too slow.

**For our specific case involving opening only 1 dataset** with `open_dataset()` vs. `open_mfdataset()`, I don't think the performance difference is that significant because coordinate concatenation isn't necessary like it is with multiple datasets. **I seems safe to extend `open_mfdataset()`, but some performance metrics would be nice to confirm this.** 

> I'd be fine with this, but ideally `open_mfdataset()` could also handle a string [`directory`](https://github.com/xCDAT/xcdat/blob/3913ea2cba91a24d099686072832cebf62ac4911/xcdat/dataset.py#L76). I think this would handle all cases in this PR.

I'll take a look at your implementation and suggest some changes to extend `open_mfdataset()`.

xcdat/dataset.py
```diff
@@ -200,9 +203,14 @@ def open_mfdataset(
...
     References
     ----------
-
     .. [2] https://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html
     """
+    if isinstance(paths, str):
+        if paths.split(".")[-1] == "xml":
+            paths = _parse_xml_for_nc_glob(paths)
+        elif os.path.isdir(paths):
+            paths = _parse_dir_for_nc_glob(paths)
```
tomvothecoder: The main logic in `open_mfdataset()` for parsing XML or a directory path.

xcdat/dataset.py
```diff
@@ -200,9 +203,14 @@ def open_mfdataset(
...
     References
     ----------
-
     .. [2] https://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html
     """
+    if not isinstance(paths, list):
+        if _is_paths_to_xml(paths):
+            paths = _parse_xml_for_nc_glob(paths)  # type: ignore
+        elif os.path.isdir(paths):
+            paths = _parse_dir_for_nc_glob(paths)  # type: ignore
```
tomvothecoder, line 212: The main logic addition in `open_mfdataset()` for the `paths` argument being an XML or directory.

xcdat/dataset.py
```diff
@@ -359,6 +367,95 @@ def decode_time(dataset: xr.Dataset) -> xr.Dataset:
...
+        A string or pathlib.Path represnting a file path.
+
+    Returns
+    -------
+    bool
+    """
+    if isinstance(paths, str):
+        return paths.split(".")[-1] == "xml"
+    elif isinstance(paths, pathlib.Path):
+        return paths.parts[-1].endswith("xml")
```
tomvothecoder, line 385: We also needed to support `pathlib.Path` alongside `str`.

xcdat/dataset.py
```diff
@@ -359,6 +367,95 @@ def decode_time(dataset: xr.Dataset) -> xr.Dataset:
...
+    dir_attr = root.attrib.get("directory")
+    if dir_attr is None:
+        raise KeyError(
+            f"The XML file ({xml_path}) does not have a 'directory' attribute "
+            "that points to a directory of `.nc` dataset files."
+        )
+
+    glob_path = dir_attr + "/*.nc"
+
+    return glob_path
```
tomvothecoder, line 431: New function `_parse_xml_for_nc_glob()` for parsing CDML XML files.

xcdat/dataset.py
```diff
@@ -359,6 +367,95 @@ def decode_time(dataset: xr.Dataset) -> xr.Dataset:
...
+    file_list = os.listdir(dir_path)
+
+    if len(file_list) == 0:
+        raise ValueError(
+            f"The directory {dir_path} has no netcdf (`.nc`) files to open."
+        )
+
+    glob_path = dir_path + "/*.nc"
+
+    return glob_path
```
tomvothecoder, line 456: New function `_parse_dir_for_nc_glob` for parsing a directory

xcdat/dataset.py
```diff
@@ -359,6 +367,95 @@ def decode_time(dataset: xr.Dataset) -> xr.Dataset:
...
+    -------
+    bool
+    """
+    if isinstance(paths, str):
+        return paths.split(".")[-1] == "xml"
+    elif isinstance(paths, pathlib.Path):
+        return paths.parts[-1].endswith("xml")
+
+
+def _parse_xml_for_nc_glob(xml_path: str) -> str:
```
tomvothecoder, line 388:
```suggestion
def _parse_xml_for_nc_glob(xml_path: Union[str, pathlib.Path]) -> str:
```

xcdat/dataset.py
```diff
@@ -359,6 +367,95 @@ def decode_time(dataset: xr.Dataset) -> xr.Dataset:
...
+    """Parses a CDAT XML file for a glob of `*.nc` paths.
+
+    The CDAT "Climate Data Markup Language" (CDML) is a dialect of XML with a
+    defined set of attributes. The "directory" attribute pointing to a directory
+    of `.nc` files is extracted from the specified XML file. Refer to [6]_ for
+    more information on CDML.
+
+    Parameters
+    ----------
+    xml_path : str
```
tomvothecoder, line 398:
```suggestion
    xml_path : Union[str, pathlib.Path]
```

xcdat/dataset.py
```diff
@@ -359,6 +367,95 @@ def decode_time(dataset: xr.Dataset) -> xr.Dataset:
...
+    References
+    ----------
+    .. [6] https://cdms.readthedocs.io/en/latest/manual/cdms_6.html
+    """
+    # `resolve_entities=False` and `no_network=True` guards against XXE attacks.
+    # Source: https://rules.sonarsource.com/python/RSPEC-2755
+    parser = etree.XMLParser(resolve_entities=False, no_network=True)
+    tree = etree.parse(xml_path, parser)
+    root = tree.getroot()
+
```
tomvothecoder, line 421: I'm using `lxml` to with `resolve_entities=False` and `no_network=True` to avoid XXE attacks.
These settings aren't easily configurable with `ElementTree` from the Python standard lib.

More info on XXE attacks:
- https://rules.sonarsource.com/python/RSPEC-2755
- https://codeql.github.com/codeql-query-help/python/py-xxe/
- [Python docs](https://docs.python.org/3/library/xml.etree.elementtree.html) warning that `ElementTree` is not secure

tests/test_dataset.py
```diff
@@ -625,6 +626,74 @@ def test_skip_decoding_times_explicitly(self):
...
+        ds1.to_netcdf(self.file_path1)
+        ds2 = generate_dataset(decode_times=False, cf_compliant=False, has_bounds=True)
+        ds2 = ds2.rename_vars({"ts": "tas"})
+        ds2.to_netcdf(self.file_path2)
+
+        result = open_mfdataset(str(self.dir), decode_times=True)
+        expected = ds1.merge(ds2)
+
+        result.identical(expected)
+
```
tomvothecoder, line 696: New unit tests covering XML and directory `paths` argument

tomvothecoder (commented):
Hey @pochedls, this PR is ready for review.

It was pretty easy to adapt your code into `open_mfdataset()`. The hardest part was figuring out how to generate a dummy XML file for the unit tests, but I figured it out lol.

tests/test_dataset.py
```diff
@@ -625,6 +626,74 @@ def test_skip_decoding_times_explicitly(self):
         expected = ds1.merge(ds2)
         assert result.identical(expected)
 
+    def test_raises_error_if_xml_does_not_have_root_direcory(self):
```
tomvothecoder, line 629:
```suggestion
    def test_raises_error_if_xml_does_not_have_root_directory_attr(self):
```

pochedls (commented):
This looks good to me. I think we could be more sophisticated in globbing directories. We could see if all files are either `.nc`, `.hdf`, `.grib`, etc. and, if so, open them as a multi-file dataset. If the files are mixed, we can throw an error. This sort of functionality can be left to future development (and worked on when someone identifies this need).

tomvothecoder:
> This looks good to me. I think we could be more sophisticated in globbing directories. We could see if all files are either `.nc`, `.hdf`, `.grib`, etc. and, if so, open them as a multi-file dataset. If the files are mixed, we can throw an error. This sort of functionality can be left to future development (and worked on when someone identifies this need).

Great idea, I agree with you. I'll go ahead and merge this PR!

PR #388, opened 2022-11-11
Add API extending xgcm vertical regridding

### LINKED ISSUE #45: [Feature]: Add vertical regridding
Blocked by #71 

Vertical regridder that can remap hybrid vertical coordinates to standard pressure levels

Acceptance criteria:

- [ ] Remapping to z-levels
- [ ] Add more....

### Additional context

[cdms](https://github.com/CDAT/cdutil/blob/master/cdutil/vertical.py#L8) has reconstructPressureFromHybrid, logLinearInterpolation, and linearInterpolation

jasonb5:
## Description

- Closes #45

Adds vertical regridding via [xgcm](https://xgcm.readthedocs.io/en/latest/transform.html#).

## Checklist

If applicable:


jasonb5:
Tasks left:
- [ ] Add documentation
- [ ] Add tests
- [ ] Add example notebook
- [ ] Determine method to derive grid coord, do we have a reliable way to tell position of data point e.g. left, center, right, inner, outer, maybe via cf_xarray
- [ ] Add method to derive `target_data` if `formula_terms` is present in vertical coordinate and variables are present in dataset otherwise rely on user to pass variable.

tomvothecoder:
Hey @jasonb5, feel free to tag people you think should review this PR. We can split testers up by API usability and code review like we do with other big PRs.

jasonb5:
@tomvothecoder @pochedls @lee1043 @chengzhuzhang Feel free to begin reviewing the code and let me know what you think about the API usability. There shouldn't be any big changes coming from my end.

chengzhuzhang:
The codes are nicely structured and easy to review, I'm just logging some initial observation after a first review and my plan next:

- Unlike `xesmf`, `xgcm` is made to be a mandatory dependency. I understand that it is our only backend for vertical regriding and it doesn't have issue that `xesmf` currently have (not supported on certain platforms). 
- An ocean data analysis notebook is provided, I will plan to update with comments and supply a common case used in atmosphere data. 
- I need to review the comment https://github.com/xCDAT/xcdat/pull/388#issuecomment-1412428664, and summarize  what has been done and create issue or discussion item for future enhancement. 
- Look at API usability with more types of data.

chengzhuzhang:
@jasonb5 Hey, Jason, I tried to push commits to this PR with `git push jasonb5 HEAD:jasonb5/adds_vertical_regrid`
But got `! [remote rejected] HEAD -> jasonb5/adds_vertical_regrid (permission denied)`
I think as fork owner, you could allow changes following[ this post](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). 
I figured it might be easier to commit changes to this PR created from your fork, but let me know if I should do it other ways. Thanks!

jasonb5:
> * An ocean data analysis notebook is provided, I will plan to update with more comments and supply a common case used in atmosphere data.

An atmosphere example would be great. Thanks!

xcdat/regridder/xgcm.py
```diff
@@ -0,0 +1,238 @@
...
+logger = setup_custom_logger(__name__)
+
+
+class XGCMRegridder(BaseRegridder):
+    def __init__(
+        self,
+        input_grid: xr.Dataset,
+        output_grid: xr.Dataset,
+        method: XGCMVerticalMethods = "linear",
+        target_data: Optional[Union[str, xr.DataArray, None]] = None,
```
tomvothecoder, line 20:
```suggestion
        target_data: Optional[Union[str, xr.DataArray]] = None,
```

Remove `None` in type annotation since `Optional` implies it

docs/gallery.rst
```diff
@@ -14,4 +14,5 @@ Please checkout the :doc:`contributing` guide.
     examples/spatial-average.ipynb
     examples/temporal-average.ipynb
     examples/climatology-and-departures.ipynb
-    examples/horizontal-regridding.ipynb
+    examples/regridding-horizontal.ipynb
+    examples/regridding-vertical.ipynb
```
tomvothecoder, line 18: Added these notebooks to the docs

xcdat/regridder/xgcm.py
```diff
@@ -0,0 +1,238 @@
...
+        grid_positions: Optional[Dict[str, str]] = None,
+        periodic: Optional[bool] = False,
```
tomvothecoder, line 22:
```suggestion
        periodic: bool = False,
```

`Optional` is used if the default value is `None`

tomvothecoder (changes requested):
My initial code review. I'll do another one once this PR seems ready to merge.

It looks good overall!

pochedls:
Hi @jasonb5 – this looks great! [Moving this chain to the PR rather than the original issue]. 

Adding a few comments based on a minimal atmospheric example (code below for reference):

* I notice that the `target_lev` dataset that I create needs to include the coordinate `lev`, though standard pressure level data is typically named `plev`. Would it be hard to give user control over the name of the level name (perhaps later as an improvement)? Or should the user just rename this after regridding?
* Note the `FutureWarning` – should the syntax `Grid.transform` be implemented now (I didn't actually look at the code to see if this comment makes sense or not – this is based only off the warming)
* Is it easy to get rid of the logger warning about missing latitude units (by solving the underlying issue)? Why is this happening for latitude and not longitude? 
* Can the method options be added to the docstring or would this be problematic (maybe this is actually a xgcm option)? If this is easy to add and isn't bad practice, it would be helpful to add it. 
* Does the xgcm method have a docstring that you can link to from `ds.regridder.vertical`? You do this for `ds.regridder.horizontal` (pointing to `xcdat.regridder.xesmf.XESMFRegridder`, for example). I'm curious what `theta` is in the docstring example (maybe this addresses my first comment?). 
* The output units on the regridded dataset are incorrect. The minimal example below doesn't provide enough information to add the units, so this makes sense, but it seems like the output units / attributes for `lev` should be pulled from the target dataset rather than copied from the source dataset. It could be good to address this.
* Suggestion for an API Change (for consideration / comment):
    * This comment is based on the behavior from CDAT, which [accepts a list of target levels](https://github.com/CDAT/cdutil/blob/master/cdutil/vertical.py#L52). Could the `target_grid` (called `output_grid` in the docstring) be either a List, Dataset, or String? I'm envisioning the following behavior: 
        * **If a dataset** – keep the core logic the same (except copy the level attributes to the output dataset); 
        * **If a List** – generate a target dataset, which can then be used for vertical regridding (output level attributes would be largely un-populated since they are not specified); 
        * **If a String** (either `standard_atmosphere` or `standard_ocean`): create a standard target dataset (with appropriate attributes, e.g., axis, units, name, etc.), which is then used for vertical regridding. 

### Simple example code

```python
import xcdat as xc
import numpy as np

fn = '/p/css03/esgf_publish/CMIP6/CMIP/MPI-M/MPI-ESM1-2-HR/historical/r8i1p1f1/CFmon/ta/gn/v20190710/*.nc'
ds = xc.open_mfdataset(fn)
p = ds.ap + ds.b*ds.ps
target_lev = xc.create_grid(lev=np.linspace(100000, 1, 13))

ds_regrid = ds.regridder.vertical("ta", target_lev, method="log", target_data=p)
```

> /../xgcm/grid.py:987: FutureWarning: From version 0.8.0 the Axis computation methods will be removed, in favour of using the Grid computation methods instead. i.e. use `Grid.transform` instead of `Axis.transform`
>   warnings.warn(
> 2023-04-18 14:09:52,753 [WARNING]: bounds.py(_create_bounds:398) >> The 'lat' coordinate variable is missing a 'units' attribute. Assuming 'units' is 'degrees_north'.

### Example output

```
ds_regrid.lev
```
> <xarray.DataArray 'lev' (lev: 13)>
> array([1.000000e+05, 9.166675e+04, 8.333350e+04, 7.500025e+04, 6.666700e+04,
>        5.833375e+04, 5.000050e+04, 4.166725e+04, 3.333400e+04, 2.500075e+04,
>        1.666750e+04, 8.334250e+03, 1.000000e+00])
> Coordinates:
>   * lev      (lev) float64 1e+05 9.167e+04 8.333e+04 ... 1.667e+04 8.334e+03 1.0
> Attributes:
>     bounds:         lev_bnds
>     units:          1
>     axis:           Z
>     positive:       down
>     long_name:      hybrid sigma pressure coordinate
>     standard_name:  atmosphere_hybrid_sigma_pressure_coordinate
>     formula:        p = ap + b*ps
>     formula_terms:  ap: ap b: b ps: ps

Tagging @mzelinka / @chengzhuzhang – to see if they have comments on my suggestions / comments.

pochedls (commented):
@jasonb5 – I did not review the code (Tom did this) or extensively look across CMIP (Mark did this). I think this looks good. I have made some comments – some of which may be better addressed in the future.

chengzhuzhang:
@durack1 just to pump this thread up. By any chance you have reviewed the ocean example for usability issues? Here is [the link](https://github.com/xCDAT/xcdat/blob/9e2800ccb77f6f118f74429ab40052019206e5d6/docs/examples/regridding-vertical.ipynb). Mark, Steve and I have looked at using vertical remapping for atmospheric variables. It would be very helpful to get your insights from ocean side.

xcdat/regridder/accessor.py
```diff
@@ -306,13 +314,71 @@ def horizontal(
...
+    def vertical(
+        self,
+        data_var: str,
+        output_grid: xr.Dataset,
+        tool: VerticalRegridTools = "xgcm",
+        **options: Any,
+    ) -> xr.Dataset:
+        """
+        Apply vertical regridding to ``data_var`` of the current
+        ``xr.Dataset`` to ``output_grid``.
```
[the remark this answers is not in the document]
jasonb5: Resolved with 7420928bfa6cccc66c9db25d43a6a21a4814e31f.

xcdat/regridder/accessor.py
```diff
@@ -306,13 +314,71 @@ def horizontal(
...
+
+        >>> ds.regridder.vertical("T", output_grid, method="linear", theta="pressure")
+        """
+        try:
+            regrid_tool = VERTICAL_REGRID_TOOLS[tool]
+        except KeyError as e:
+            raise ValueError(
+                f"Tool {e!s} does not exist, valid choices {list(VERTICAL_REGRID_TOOLS)}"
+            )
+        regridder = regrid_tool(self._ds, output_grid, **options)
```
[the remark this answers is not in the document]
jasonb5: Resolved with 7420928bfa6cccc66c9db25d43a6a21a4814e31f.

xcdat/regridder/grid.py
```diff
@@ -425,94 +429,109 @@ def create_zonal_grid(grid: xr.Dataset) -> xr.Dataset:
...
-) -> xr.Dataset:
+COORD_DEFAULT_ATTRS: Dict[str, Dict[str, Any]] = {
+    "X": dict(units="degrees_east", **CF_ATTR_MAP["X"]),
+    "Y": dict(units="degrees_north", **CF_ATTR_MAP["Y"]),
+    "Z": CF_ATTR_MAP["Z"],
+}
+
+
+def create_grid(**kwargs: CoordOptionalBnds) -> xr.Dataset:
     """Creates a grid for a give latitude and longitude array.
```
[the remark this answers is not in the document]
jasonb5: Resolved with 7420928bfa6cccc66c9db25d43a6a21a4814e31f.

xcdat/regridder/grid.py
```diff
@@ -425,94 +429,109 @@ def create_zonal_grid(grid: xr.Dataset) -> xr.Dataset:
...
+    >>> lat_bnds = np.vstack((lat - (2.5 / 2), lat + (2.5 / 2))).T
+    >>> xcdat.create_grid(lat=(lat, lat_bnds), lon=lon)
+
+    Create vertical grid:
+
+    >>> xcdat.create_grid(lev=np.linspace(1000, 1, 20))
     """
-    data_vars = {}
+    if len(kwargs) == 0:
+        raise ValueError("Must pass atleast 1 coordinate.")
```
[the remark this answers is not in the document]
jasonb5: Resolved with 7420928bfa6cccc66c9db25d43a6a21a4814e31f.

xcdat/regridder/xgcm.py
```diff
@@ -0,0 +1,238 @@
...
+        Regrid data:
+
+        >>> data_new_grid = regridder.vertical("T", ds)
+        """
+        super().__init__(input_grid, output_grid)
+
+        if method not in get_args(XGCMVerticalMethods):
+            raise ValueError(
+                f"{method!r} is invalid, possible choices: {', '.join(get_args(XGCMVerticalMethods))}"
+            )
```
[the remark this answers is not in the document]
jasonb5: Resolved with 7420928bfa6cccc66c9db25d43a6a21a4814e31f.

xcdat/regridder/xgcm.py
```diff
@@ -0,0 +1,238 @@
...
+
+        output_ds = output_ds.bounds.add_missing_bounds()
+
+        return output_ds
+
+    def _get_grid_coords(self) -> Dict[str, Dict[str, str]]:
+        if self._method == "conservative":
+            raise RuntimeError(
+                "Conservative regridding requires a second point position, pass these manually"
+            )
```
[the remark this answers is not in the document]
jasonb5: Resolved with 7420928bfa6cccc66c9db25d43a6a21a4814e31f.

xcdat/regridder/xgcm.py
```diff
@@ -0,0 +1,238 @@
...
+        ):
+            grid_positions = {"center": coord_z.name}
+        elif coord_z[0] == bounds_z[0][0]:
+            grid_positions = {"left": coord_z.name}
+        elif coord_z[0] == bounds_z[0][1]:
+            grid_positions = {"right": coord_z.name}
+        else:
+            raise RuntimeError(
+                "Could not determine the grid point positions, pass these manually using the `grid_positions` argument"
+            )
```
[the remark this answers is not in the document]
jasonb5: Resolved with 7420928bfa6cccc66c9db25d43a6a21a4814e31f.

xcdat/regridder/xgcm.py
jasonb5: Resolved with 7420928bfa6cccc66c9db25d43a6a21a4814e31f.

xcdat/regridder/xgcm.py
jasonb5: Resolved with 7420928bfa6cccc66c9db25d43a6a21a4814e31f.

xcdat/regridder/grid.py
```diff
@@ -1,9 +1,13 @@
...
 
 import numpy as np
 import xarray as xr
 
-from xcdat.axis import CFAxisKey, get_dim_coords
+from xcdat.axis import CF_ATTR_MAP, VAR_NAME_MAP, CFAxisKey, get_dim_coords
+from xcdat.regridder.base import CoordOptionalBnds
+
+_VAR_NAME_MAP = VAR_NAME_MAP.copy()
+_VAR_NAME_MAP["Z"] = ["vertical", "height", "pressure", "lev"]
```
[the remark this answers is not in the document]
jasonb5: I did this original as I was unsure why we were using `coordinate_criteria["Z"]["standard_name"]` which wasn't giving the correct names. I intended to update `axis.py` but missed it during the initial commit, I'll fix it in this PR.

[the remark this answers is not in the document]
jasonb5: Resolved with 3c32cec4884825befb49f0f87a7dbe02d502ee32.

xcdat/regridder/grid.py
```diff
@@ -425,94 +429,109 @@ def create_zonal_grid(grid: xr.Dataset) -> xr.Dataset:
...
-    lat: Union[np.ndarray, xr.DataArray],
-    lon: Union[np.ndarray, xr.DataArray],
-    lat_bnds: Optional[Union[np.ndarray, xr.DataArray]] = None,
-    lon_bnds: Optional[Union[np.ndarray, xr.DataArray]] = None,
-) -> xr.Dataset:
+COORD_DEFAULT_ATTRS: Dict[str, Dict[str, Any]] = {
+    "X": dict(units="degrees_east", **CF_ATTR_MAP["X"]),
+    "Y": dict(units="degrees_north", **CF_ATTR_MAP["Y"]),
+    "Z": CF_ATTR_MAP["Z"],
+}
```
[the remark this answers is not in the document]
jasonb5: Resolved with 0b36eb938e42f3a40b508456fd8c3f22e2ed4d30. Should we update the other files in this PR or create a new one?

xcdat/regridder/grid.py
```diff
@@ -425,94 +429,109 @@ def create_zonal_grid(grid: xr.Dataset) -> xr.Dataset:
...
-    lat : Union[np.ndarray, xr.DataArray]
-        Array of latitude values.
-    lon : Union[np.ndarray, xr.DataArray]
-        Array of longitude values.
-    lat_bnds : Optional[Union[np.ndarray, xr.DataArray]]
-        Array of bounds for latitude values.
-    lon_bnds : Optional[Union[np.ndarray, xr.DataArray]]
-        Array of bounds for longitude values.
+    **kwargs : CoordOptionalBnds
+        Mapping of coordinate name and data with optional bounds.
```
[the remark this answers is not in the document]
jasonb5: Resolved with 08c206e644351bdab2f7891041e7a8aeb3f11e4b.

xcdat/regridder/xgcm.py
```diff
@@ -0,0 +1,238 @@
...
+            **self._extra_options,
+        )
+
+        input_z_name = ds.cf["Z"].name
+
+        # xgcm always outputs vertical coordinate as 'lev'
+        if input_z_name != "lev":
+            output_da = output_da.rename({"lev": input_z_name})
+
+        output_da = output_da.transpose(*ds[data_var].dims)
```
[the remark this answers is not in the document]
jasonb5: Resolved with e3691fff5f2415b98d639def9a034566a120a6aa.

xcdat/regridder/xgcm.py
```diff
@@ -0,0 +1,238 @@
...
+
+        output_ds = xr.Dataset({data_var: output_da}, attrs=ds.attrs)
+
+        output_ds = preserve_bounds(ds, self._output_grid, output_ds)
+
+        output_ds = output_ds.bounds.add_missing_bounds()
```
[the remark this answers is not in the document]
jasonb5: Resolved with e3691fff5f2415b98d639def9a034566a120a6aa.

jasonb5:
@pochedls 
>     * I notice that the `target_lev` dataset that I create needs to include the coordinate `lev`, though standard pressure level data is typically named `plev`. Would it be hard to give user control over the name of the level name (perhaps later as an improvement)? Or should the user just rename this after regridding?

This is fixed you can use any valid name found in xcdat.axis.VAR_NAME_MAP["Z"] e.g., height, lev, plev, etc.

Would you expected the output vertical coordinate name to match the input or the `output_grid`'s vertical coordinate name? Would providing `target_data` change this?
>     * Note the `FutureWarning` – should the syntax `Grid.transform` be implemented now (I didn't actually look at the code to see if this comment makes sense or not – this is based only off the warming)

This is something `xgcm` needs to update in their code, I can silence the error in ours.
>     * Is it easy to get rid of the logger warning about missing latitude units (by solving the underlying issue)? Why is this happening for latitude and not longitude?
> 
>     * Can the method options be added to the docstring or would this be problematic (maybe this is actually a xgcm option)? If this is easy to add and isn't bad practice, it would be helpful to add it.

I can add them to xcdat.Dataset.regridder.vertical like I did in [xcdat.Dataset.regridder.horizontal](https://xcdat.readthedocs.io/en/stable/generated/xarray.Dataset.regridder.horizontal.html).
>     * Does the xgcm method have a docstring that you can link to from `ds.regridder.vertical`? You do this for `ds.regridder.horizontal` (pointing to `xcdat.regridder.xesmf.XESMFRegridder`, for example). I'm curious what `theta` is in the docstring example (maybe this addresses my first comment?).

I'll add a link to `xcdat.regridder.xesmf.XESMFRegridder` in `ds.regridder.vertical`.
>     * The output units on the regridded dataset are incorrect. The minimal example below doesn't provide enough information to add the units, so this makes sense, but it seems like the output units / attributes for `lev` should be pulled from the target dataset rather than copied from the source dataset. It could be good to address this.

Should the attributes on the output's vertical coordinate come from the `output_grid`'s vertical coordinate? If `target_data` is present, should they still come from the `output_grid`'s vertical coordinate?
>     * Suggestion for an API Change (for consideration / comment):
>       
>       * This comment is based on the behavior from CDAT, which [accepts a list of target levels](https://github.com/CDAT/cdutil/blob/master/cdutil/vertical.py#L52). Could the `target_grid` (called `output_grid` in the docstring) be either a List, Dataset, or String? I'm envisioning the following behavior:
>         
>         * **If a dataset** – keep the core logic the same (except copy the level attributes to the output dataset);
>         * **If a List** – generate a target dataset, which can then be used for vertical regridding (output level attributes would be largely un-populated since they are not specified);
>         * **If a String** (either `standard_atmosphere` or `standard_ocean`): create a standard target dataset (with appropriate attributes, e.g., axis, units, name, etc.), which is then used for vertical regridding.

I'm fine doing this as long as their are no other objections.

chengzhuzhang:
@pochedls Thank you for carefully review this feature. Great catch on the incorrect units of regridded data. We should correct this, but I don't know what is the best way to address this:

For a simple interpolation (without needing target_data), it is fine to default units of lev to original datasets. I don't know if we should allow units change in this case, i.e. original data have plev.units = Pa, while, output.lev.units = hPa or mb

More complicated, the units of output_grid.lev needs to be consistent with the units of target_data if target_data is an input.  In this case, I think users should be responsible to provide `units` for output_grid.lev or target_data, and that  the units of the output dataset should base on either, if units unchanged, xcdat can use original lev units and print a warning, or flag error if there is conflict between `units` for output_grid.lev and target_data?

pochedls:
> @pochedls Thank you for carefully review this feature. Great catch on the incorrect units of regridded data. We should correct this, but I don't know what is the best way to address this:
> 
> For a simple interpolation (without needing target_data), it is fine to default units of lev to original datasets. I don't know if we should allow units change in this case, i.e. original data have plev.units = Pa, while, output.lev.units = hPa or mb
> 
> More complicated, the units of output_grid.lev needs to be consistent with the units of target_data if target_data is an input. In this case, I think users should be responsible to provide `units` for output_grid.lev or target_data, and that the units of the output dataset should base on either, if units unchanged, xcdat can use original lev units and print a warning, or flag error if there is conflict between `units` for output_grid.lev and target_data?

Sorry – I've misrepresented this – I said the "units" are incorrect, but it is the metadata in general that is not right. Metadata for a regridded `plev` looks like this, which is obviously not right:

> plev (plev) float64 1e+05 9.167e+04 8.333e+04 ... 1.667e+04 8.334e+03 1.0
> Attributes:
> bounds: lev_bnds
> units: 1
> axis: Z
> positive: down
> long_name: hybrid sigma pressure coordinate
> standard_name: atmosphere_hybrid_sigma_pressure_coordinate
> formula: p = ap + b*ps
> formula_terms: ap: ap b: b ps: ps

chengzhuzhang:
@pochedls thanks for clarifying. I only focused on `units` , which is important to fix. And I didn't realize in general the metadata is problematic.

pochedls:
@jasonb5 – in looking at the [xgcm transform documentation](https://xgcm.readthedocs.io/en/latest/api.html#xgcm.Grid.transform), I noticed that we don't capture some of their optional arguments in our documentation ([here](https://github.com/xCDAT/xcdat/blob/929dbda9e4fc20185a14c517a715eb27dc79f943/xcdat/regridder/xgcm.py#L15), [here](https://github.com/xCDAT/xcdat/blob/929dbda9e4fc20185a14c517a715eb27dc79f943/xcdat/regridder/xgcm.py#L121)). Is there a way to reference `xgcm.Grid.transform` docs?

xcdat/regridder/xgcm.py
```diff
@@ -0,0 +1,244 @@
...
+        grid_positions : Optional[Dict[str, str]]
+            Mapping of dimension positions, by default None. If ``None`` then an
+            attempt is made to derive this argument.
+        periodic : bool
+            Whether the grid is periodic, by default False.
+        extra_init_options : Optional[Dict[str, Any]]
+            Extra options passed to the ``xgcm.Grid`` constructor, by default
+            None.
+        options : Optional[Dict[str, Any]]
+            Extra options passed to the ``xgcm.Grid.trasnform`` method.
```
pochedls, line 61: Typo fix:

```suggestion
            Extra options passed to the ``xgcm.Grid.transform`` method.
```

xcdat/regridder/xgcm.py
```diff
@@ -0,0 +1,244 @@
...
+
+        >>> output_grid = xcdat.create_grid(np.linspace(1000, 1, 20))
+
+        Create regridder:
+
+        >>> regridder = xgcm.XGCMRegridder(ds, output_grid, method="linear", target_data="pressure")
+
+        Regrid data:
+
+        >>> data_new_grid = regridder.verical("T", ds)
```
pochedls, line 155:
```suggestion
        >>> data_new_grid = regridder.vertical("T", ds)
```

xcdat/regridder/accessor.py
```diff
@@ -306,13 +314,78 @@ def horizontal(
...
+
+        Examples
+        --------
+        Create destination grid:
+
+        >>> output_grid = xcdat.create_grid(lev=np.linspace(1000, 1, 20))
+
+        Regrid variable using "xgcm":
+
+        >>> ds.regridder.vertical("T", output_grid, method="linear", theta="pressure")
```
pochedls, line 379: Should `theta` be here?

jasonb5:
@pochedls I'll add links to the xgcm documentation and add some examples of passing extra arguments.

pochedls:
In looking at ocean vertical regridding, I came across one unexpected issue (following up since I mentioned this to @jasonb5): 

```python
import xcdat as xc
# define / open dataset
dspath = '/p/css03/esgf_publish/CMIP6/CMIP/CNRM-CERFACS/CNRM-CM6-1/historical/r3i1p1f2/Omon/thetao/gn/v20190125/thetao_Omon_CNRM-CM6-1_historical_r3i1p1f2_gn_200001-201412.nc'
ds = xc.open_mfdataset(dspath)
# subset for minimal example
ds = ds.isel(time=[0])
# horizontal regridding
dsg = ds.regridder.horizontal('thetao', ngrid, method='conservative_normed')
# vertical regridding
dsv = dsg.regridder.vertical('thetao', target_lev, method="linear")
```

> RuntimeError: Could not determine 'Z' bounds in input dataset

Note that I **do not** have this problem if I skip the horizontal regridding step. It appears the `Z` bounds are dropped during horizontal regridding (`dsg = dsg.bounds.add_missing_bounds(axes='Z')` seems to address this – although I realize I _should_ copy the bounds from `ds` to `dsg`). 

This issue also occurs for `/p/css03/scratch/cmip6/CMIP/IPSL/IPSL-CM6A-LR/historical/r2i1p1f1/Omon/thetao/gn/v20190305/`, but (weirdly) not for other models I tested. 

So in documenting this, I think this is **not** a vertical regridding issue, but may be an issue with the horizontal regridder (simply not copying the bounds under some conditions). @jasonb5 – should I open a separate issue about the dropped vertical bounds? 

The good news – is that the vertical regridder seems to be doing a good job figuring out the vertical axis of interest (I thought this was the problem – but it does not appear to be).

jasonb5:
@pochedls If this is specific to the horizontal regridder, let's open a separate issue.

pochedls:
> @pochedls If this is specific to the horizontal regridder, let's open a separate issue.

@jasonb5 – I was working on opening a ticket about this issue and I realize that the bounds are **not** dropped on the v0.5.0 release – it is happening on this branch. Here's a MCVE (fixed from above):

```python
import xcdat as xc
import numpy as np
# define target grid
nlat = np.arange(-88.75, 90, 2.5)
nlon = np.arange(1.25, 360, 2.5)
ngrid = xc.create_grid(lat=nlat, lon=nlon)
# define / open dataset
dspath = '/p/css03/esgf_publish/CMIP6/CMIP/CNRM-CERFACS/CNRM-CM6-1/historical/r3i1p1f2/Omon/thetao/gn/v20190125/thetao_Omon_CNRM-CM6-1_historical_r3i1p1f2_gn_200001-201412.nc'
# another example path (for the atmospheric grid): 
# /p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2/amip/r1i1p1f1/Amon/ta/gn/v20190218/ta_Amon_CESM2_amip_r1i1p1f1_gn_195001-201412.nc
ds = xc.open_mfdataset(dspath)
# subset for minimal example
ds = ds.isel(time=[0])
# horizontal regridding
dsg = ds.regridder.horizontal('thetao', ngrid, method='conservative_normed')
dsg.lev_bnds
```

> AttributeError: 'Dataset' object has no attribute 'lev_bnds'

One other potential issue only associated with this branch. When I exit iPython I get this warning:

> [WARNING] yaksa: 10 leaked handle pool objects

xcdat/regridder/grid.py
```diff
@@ -425,94 +426,106 @@ def create_zonal_grid(grid: xr.Dataset) -> xr.Dataset:
...
-        Array of latitude values.
-    lon : Union[np.ndarray, xr.DataArray]
-        Array of longitude values.
-    lat_bnds : Optional[Union[np.ndarray, xr.DataArray]]
-        Array of bounds for latitude values.
-    lon_bnds : Optional[Union[np.ndarray, xr.DataArray]]
-        Array of bounds for longitude values.
+    **kwargs : CoordOptionalBnds
+        Mapping of coordinate name and data with optional bounds. See
+        :py:data:`xcdat.axis.VAR_NAME_MAP` for valid coordinate names.
```
tomvothecoder, line 439: Hey Jason, my suggestion here was to make the parameters for `create_grid` more explicit. 

In my opinion, using `**kwargs` over an explicit args/kwargs can create confusion and reduce readability since there isn't set number of clearly defined parameters for the function. Here is a post a read about `**kwargs`: https://softwareengineering.stackexchange.com/a/294486

I would suggest keeping the args and kwargs, then extending as needed.

jasonb5: @pochedls @chengzhuzhang @lee1043 @tomvothecoder 
How do we feel about the following function to replace the current `create_grid`.
```python
def create_grid(x: GridCoord, y: GridCoord, z: GridCoord) -> xr.Dataset:
```
Where `GridCoord` could be any of the following forms.
```
(name, data)
(name, data, attrs)
(name, data, bnds)
(name, data, bnds, attrs)
```

tomvothecoder: From what I understand, this is a tuple where:
- name = string
- data = array of floats
- attrs = dictionary mapping str to str <OPTIONAL>
- bnds = 2d array of floats <OPTIONAL>

Is this correct?

jasonb5: @tomvothecoder Yes that's correct. The `data` and `bnds` values can be a list, ndarray or DataArray (1d/2d respectively).

tomvothecoder: Thanks for confirming. I am curious about the benefits you see in using a tuple of sub-arguments over an `xr.DataArray`? I'm pretty sure this was discussed with the team during our meetings, but I didn't really weigh in because there's a lot to track!

jasonb5: The only benefit I can see is for users who may not want learn how to construct a `xr.DataArray`. This was discussed, I remember it being opposed at first but thought we'd changed the end decision to allow more than just `xr.DataArray`. I'm fine just make the arguments `xr.DataArray` and updating the docstring to include creating the `xr.DataArray`'s.

tomvothecoder: I agree about only seeing that benefit. The tradeoff of API flexibility with simplicity and maintainability doesn't seem enough in my opinion. I'm in favor with keeping the arguments as `xr.DataArray`. This approach encourages the use of xarray objects (this library is an extension of xarray) and keeps the implementation and API usage simple. I noticed other libraries like [xgcm](https://xgcm.readthedocs.io/en/latest/api.html) and [xclim](https://xclim.readthedocs.io/en/stable/index.html) focus on accepting xarray objects.

pochedls: What if we include a `create_axis()` function (similar to the one at bottom)?

Then we could pass in an axis (DataArray derived from `create_axis()`) or a tuple of `(axis, bnds)`. The bnds could be an array or dataarray (if dataarray, we add `bnds.name` as the `bounds` attribute to `axis`, otherwise the bounds name would be `axis.name + "_bnds"` and would be produced automatically). 

```python
def create_axis(aid, axis_values, attrs=None):
    # define default attributes
    lat_attrs = {'units': 'degrees_north',
                 'axis': 'Y',
                 'long_name': 'latitude',
                 'standard_name': 'latitude'}
    lon_attrs = {'units': 'degrees_east',
                 'axis': 'X',
                 'long_name': 'longitude',
                 'standard_name': 'longitude'}
    plev_attrs = {'units': 'Pa',
                  'axis': 'Z',
                  'long_name': 'pressure',
                  'standard_name': 'air_pressure',
                  'positive': 'down'}
    lev_attrs = {'units': 'm',
                  'axis': 'Z',
                  'long_name': 'ocean depth coordinate',
                  'standard_name': 'depth',
                  'positive': 'down'}
    default_attrs = {'lat': lat_attrs,
                     'latitude': lat_attrs,
                     'lon': lon_attrs,
                     'longitude': lon_attrs,
                     'lev': lev_attrs,
                     'plev': plev_attrs}
    if attrs is None:
        attrs = default_attrs[aid]

    newAxis = xr.DataArray(
        name=aid,
        data=axis_values,
        dims=[aid],
        coords={aid: axis_values},
        attrs=attrs)

    return newAxis
```

jasonb5: Are there use cases where `create_axis` would be used without `create_grid`?

pochedls: Yeah – I have instances where I have a numpy array and want to create a DataArray (with correct axis information). I think this is kind of common. @mzelinka – were you saying that you have cases where it would be helpful to generate a new axis (with correct metadata/attributes)?

tomvothecoder: > What if we include a create_axis() function (similar to the one at bottom)?

I can see a function like `create_axis` making the process of creating xr.DataArray axes faster and less prone to error. 
**Maybe this can be scoped out in another GitHub issue?**

I had a few questions (for myself too):

  - When and how often might users want/need to create axes instead of using the ones in the file?
  - How should we decide what is considered the most sensible default attributes for each axis? Xarray sets some like `calendar="standard"` for the time axis, and we set a few others like units for latitude/longitude.
  - If we decide this makes sense to implement, we should consider calling it `create_dim` since we have [`get_dim_keys()`](https://xcdat.readthedocs.io/en/stable/generated/xcdat.get_dim_keys.html#xcdat.get_dim_keys) and [`get_dim_coords()`](https://xcdat.readthedocs.io/en/stable/generated/xcdat.get_dim_coords.html#xcdat.get_dim_coords).



---

I also like the idea of having a static dictionary to store default CF attributes. We have default attrs set in various parts of the code-base (e.g., calendar=`standard` for the time axis). It'll be cleaner to have a single reference point for default attrs.

```python
# axis.py
CF_DEFAULT_ATTRS = {
	"lat": '{'units': 'degrees_north',
                 'axis': 'Y',
                 'long_name': 'latitude',
                 'standard_name': 'latitude'
     },
	....
}
```

jasonb5: @pochedls @tomvothecoder Lets move this discussion to the [new issue](https://github.com/xCDAT/xcdat/issues/496).

jasonb5:
@pochedls This logic sounds good, I'll update accordingly.

tests/test_regrid.py
```diff
@@ -822,24 +1056,43 @@ def test_grid_raises_error_when_dataset_has_multiple_dims_for_an_axis(self):
...
+    def test_vertical_tool_check(self):
+        mock_regridder = mock.MagicMock()
+        mock_regridder.return_value.vertical.return_value = "output data"
+
+        mock_data = mock.MagicMock()
+
+        with mock.patch.dict(accessor.VERTICAL_REGRID_TOOLS, {"xgcm": mock_regridder}):
+            output = self.ac.vertical("ts", mock_data, tool="xgcm", target_data=None)
+
+        assert output == "output data"
```
tomvothecoder, line 1088: The use of MagicMock is neat here. For those interested (probably just me), the xgcm vertical regridder object is being mocked (`mock_regridder`) and set as the `xgcm` dictionary value in `accessor.VERTICAL_REGRID_TOOLS`. The return value of `mock_regridder` is set to `"output_data"`, which is being tested in the assertion on line 1088. This test is to check if the correct vertical regridding tool is being referenced. 

The actual result of the regridder object is not being tested (hence `"output_data"`) because it is already tested through `class TestXGCMRegridder`.

xcdat/regridder/accessor.py
```diff
@@ -306,13 +314,78 @@ def horizontal(
...
-                f"Tool {e!s} does not exist, valid choices {list(REGRID_TOOLS)}"
+                f"Tool {e!s} does not exist, valid choices {list(HORIZONTAL_REGRID_TOOLS)}"
             )
 
         regridder = regrid_tool(self._ds, output_grid, **options)
         output_ds = regridder.horizontal(data_var, self._ds)
 
         return output_ds
+
+    def vertical(
```
tomvothecoder, line 328: Based on #501, are we also considering calling this `vertical_xgcm` and not having a general wrapper with `tool="xgcm"`?

xcdat/regridder/base.py
```diff
@@ -1,68 +1,59 @@
...
+from xcdat.axis import CFAxisKey
 
 logger = _setup_custom_logger(__name__)
 
+Coord = Union[np.ndarray, xr.DataArray]
+
+CoordOptionalBnds = Union[Coord, Tuple[Coord, Coord]]
+
 
 def preserve_bounds(
```
tomvothecoder, line 18: Do you think this function will be used publicly, or should it be private (`_preserve_bounds()`)?

tomvothecoder (commented):
Hi Jason, here is my final review. The suggestions are mostly renaming things.

xcdat/regridder/accessor.py
jasonb5: That's correct, I'll address it in the next PR.

tomvothecoder (approved):

tomvothecoder:
🎉🎊

Issue #412, opened 2023-01-26
[Feature]: Add functionality to accurately set time bounds


pochedls:
### Is your feature request related to a problem?

Following Discussion #411 – xCDAT should modify how it handles bounds in general and time bounds in particular. xCDAT currently uses the bounds within the dataset, but if they are not present it will generate bounds for each axis (by assuming that the bounds are the midpoint between each axis point). This is problematic for time coordinates because a) time points are frequently not saved out at the midpoint value and b) months are not equal in size so the halfway point between two months do not yield the correct bound. 

This is illustrated in Slide 4 of [climate_specific_tools.pdf](https://sharebox.lsce.ipsl.fr/index.php/s/EvtQJZuWR4UBo0X#pdfviewer). 

For example, ERA5 monthly mean data is saved at midnight on the first day of the month:

```python
import xcdat as xc
fn = '/p/user_pub/climate_work/pochedley1/ecmwf/era5/new/tas_era5.nc'
ds = xc.open_dataset(fn)
ds.time[0]
```

> ...
> array(cftime.DatetimeGregorian(1979, 1, 1, 0, 0, 0, 0, has_year_zero=False),
>       dtype=object)
> ...

xCDAT then autogenerates incorrect bounds:

```python
print(ds.bounds.get_bounds('T')[0])
```

> ...
> array([cftime.DatetimeGregorian(1978, 12, 16, 12, 0, 0, 0, has_year_zero=False),
>        cftime.DatetimeGregorian(1979, 1, 16, 12, 0, 0, 0, has_year_zero=False)],
>       dtype=object)
> ...
>     xcdat_bounds:  True

Note that this issue also applies to vertical coordinate which are frequently not equally spaced. 

### Describe the solution you'd like

Addressing this issue would take a few changes:

* By default, auto-generate missing bounds for latitude and longitude only (i.e., do not automatically set bounds for vertical and temporal coordinates). This would require modification of `xcdat.open_dataset()` (in function and documentation).
* Create functions to create sub-daily, daily, and monthly time bounds, e.g., [`setAxisTimeBoundsMonthly()`](https://github.com/CDAT/cdutil/blob/master/cdutil/times.py#L1168) that set the time bounds at the start and end of the period (e.g., for the ERA5 case above, the bounds should be `1979-01-01 00:00:00` and `1979-02-01 00:00:00`).
* Consider allowing the user to set the bounds that xcdat creates
[...]

PR #417, opened 2023-02-13
Update departures and climatology APIs with reference period

### LINKED ISSUE #387: [Feature]: Allow user selection of reference period in departures

The current departures functionality determines the anomalies relative to the average value over the full time series. 

Climate scientists often look at the departures relative to a specific period (rather than the full period). For example, see [IPCC AR6 WG1 Fig. 1.12](https://www.ipcc.ch/report/ar6/wg1/figures/chapter-1/figure-1-12/) which shows the global mean surface temperature departures relative to the average temperature over 1850 – 1900. This shows zero **average** anomaly between 1850 and 1900 and emphasizes the rapid warming that follows. 

We should make it possible for xCDAT users to do this kind of calculation easily. 

xCDAT should allow for flexible selection of the climatological reference period with which to define departures. I think there are several non-mutually exclusive ways to accomplish this:

* Allow the user to provide a a custom climatology when calculating departures, e.g., `ds.temporal.departures(data_var, freq, climatology=ds_climo)`
* Allow the user to specify a climatological base period when calculating departures, e.g., `ds.temporal.departures(data_var, freq, reference_period=("1850-01-01", "1899-12-31"))`
* Or both (possibly by augmenting climatology functionality to accept a `reference_period`, i.e., `ds.temporal.climatology(data_var, freq, reference_period=("1850-01-01", "1899-12-31"))`

An alternative manual method would entail:

* Subset the dataset of interest to the climatological reference period (`ds_climo = ds.sel("1850-01-01", "1899-12-31")`)
* Take the climatology of this subsetted data (`ds_climo = ds.temporal.group_average(data_var, freq="month")`)
* Somehow subtract this climatology from the original timeseries (not sure how to do this or if this would work): `ds_departure[data_var] = ds[data_var] - ds_climo[data_var]`

### Additional context

- Related discussion: https://github.com/xCDAT/xcdat/discussions/404
- Jiwoo's xarray+xcdat implementation: https://github.com/xCDAT/xcdat_test/blob/main/cookbook/seasonal_departure_from_baseline_clim.ipynb

tomvothecoder:
## Description

- Closes #387 
- Add `reference_period` parameter to `climatology()` and `departures()` to subset climatology 
- In `departures()`, calculate the the grouped average of the observation data before calculating anomalies
  - This behavior makes `departures()` align with CDAT's APIs for calculating departures
  - If the input frequency from the dataset is the same as the output frequency `freq` arg, no grouped average is performed since it is unnecessary and incurs a performance cost 
- Update `departures()` notebook with `reference_period` example
 
## Checklist

If applicable:

### DIFF
--- a/xcdat/temporal.py
+++ b/xcdat/temporal.py
@@ -557,8 +557,8 @@ class TemporalAccessor:
             The climatological reference period, which is a subset of the entire
             time series and used for calculating departures. This parameter
             accepts a tuple of strings in the format 'yyyy-mm-dd'. For example,
-            ``('1850-01-01', 1899-12-31')``. If no value is provided, the 
-            climatological reference period will be the full period covered by 
+            ``('1850-01-01', 1899-12-31')``. If no value is provided, the
+            climatological reference period will be the full period covered by
             the dataset.
         season_config: SeasonConfigInput, optional
             A dictionary for "season" frequency configurations. If configs for
@@ -660,13 +660,20 @@ class TemporalAccessor:
         # dataset to create separate instances of the `TemporalAccessor` class.
         # This is done to avoid overriding the attributes of the current
         # instance of `TemporalAccessor` (set in step #1 above).
-        ds_avg = ds.temporal.group_average(
-            data_var,
-            freq,
-            weighted,
-            keep_weights,
-            season_config,
-        )
+
+        # Group averaging is only required if the dataset's frequency (input)
+        # differs from the `freq` arg (output).
+        ds_obs = ds.copy()
+        inferred_freq = self._infer_freq()
+        if inferred_freq != freq:
+            ds_obs = ds_obs.temporal.group_average(
+                data_var,
+                freq,
+                weighted,
+                keep_weights,
+                season_config,
+            )
+
         ds_climo = ds.temporal.climatology(
             data_var,
             freq,
@@ -680,7 +687,7 @@ class TemporalAccessor:
         # ----------------------------------------------------------------------
         # This step allows us to perform xarray's grouped arithmetic to
         # calculate departures.
-        dv_obs = ds_avg[data_var].copy()
+        dv_obs = ds_obs[data_var].copy()
         self._labeled_time = self._label_time_coords(dv_obs[self.dim])
         dv_obs_grouped = self._group_data(dv_obs)
 
@@ -702,18 +709,18 @@ class TemporalAccessor:
         with xr.set_options(keep_attrs=True):
             dv_departs = dv_obs_grouped - dv_climo
             dv_departs = self._add_operation_attrs(dv_departs)
-            ds_avg[data_var] = dv_departs
+            ds_obs[data_var] = dv_departs
 
             # The original time dimension name is restored after grouped
             # arithmetic, so the labeled time dimension name is no longer needed
             # and therefore dropped.
-            ds_avg = ds_avg.drop_vars(self._labeled_time.name)
+            ds_obs = ds_obs.drop_vars(self._labeled_time.name)
 
         if weighted and keep_weights:
             self._weights = ds_climo.time_wts
-            ds_avg = self._keep_weights(ds_avg)
+            ds_obs = self._keep_weights(ds_obs)
 
-        return ds_avg
+        return ds_obs
 
     def _infer_freq(self) -> Frequency:
         """Infers the time frequency from the coordinates.
--- a/xcdat/temporal.py
+++ b/xcdat/temporal.py
@@ -642,7 +642,9 @@ class TemporalAccessor:
         ds = self._preprocess_dataset(ds)
 
         # 1. Get the grouped average of the observation data variable.
-        ds_avg = ds.temporal.group_average(data_var, self._freq)
+        ds_avg = ds.temporal.group_average(
+            data_var, self._freq, weighted, keep_weights, season_config=season_config
+        )
         dv_obs = ds_avg[data_var].copy()
 
         # 2. Group the observation data variable by the departures frequency.
@@ -670,12 +672,13 @@ class TemporalAccessor:
         ds_climo = ds_climo.rename({self.dim: self._labeled_time.name})
         dv_climo = ds_climo[data_var]
 
-        # 4. Calculate the departures for the data variable.
-        # The formula: grouped observation data - climatology
+        # 4. Calculate the departures for the data variable using the formula,
+        # grouped observation data - climatology.
         with xr.set_options(keep_attrs=True):
             dv_departs = dv_obs_grouped - dv_climo
             dv_departs = self._add_operation_attrs(dv_departs)
-            ds_avg[f"{data_var}_departures"] = dv_departs
+            ds_avg[data_var] = dv_departs
+            ds_avg = ds_avg.drop_vars(self._labeled_time.name)
 
         if weighted and keep_weights:
             self._weights = ds_climo.time_wts
@@ -1586,11 +1589,6 @@ class TemporalAccessor:
             # Only keep the original time coordinates, not the ones labeled
             # by group.
             self._weights = self._weights.drop_vars(self._labeled_time.name)
-        # Strip "_original" from the name of the weights` time coordinates
-        # because the final departures Dataset has the original time coordinates
-        # restored after performing grouped subtraction.
-        elif self._mode == "departures":
-            self._weights = self._weights.rename({f"{self.dim}_original": self.dim})
 
         ds[self._weights.name] = self._weights
 
--- a/xcdat/temporal.py
+++ b/xcdat/temporal.py
@@ -1,4 +1,5 @@
 """Module containing temporal functions."""
+from datetime import datetime
 from itertools import chain
 from typing import Dict, List, Literal, Optional, Tuple, TypedDict, get_args
 
@@ -196,7 +197,9 @@ class TemporalAccessor:
 
         freq = self._infer_freq()
 
-        return self._averager(data_var, "average", freq, weighted, keep_weights)
+        return self._averager(
+            data_var, "average", freq, weighted=weighted, keep_weights=keep_weights
+        )
 
     def group_average(
         self,
@@ -332,7 +335,12 @@ class TemporalAccessor:
         self._set_data_var_attrs(data_var)
 
         return self._averager(
-            data_var, "group_average", freq, weighted, keep_weights, season_config
+            data_var,
+            "group_average",
+            freq,
+            weighted=weighted,
+            keep_weights=keep_weights,
+            season_config=season_config,
         )
 
     def climatology(
@@ -341,6 +349,7 @@ class TemporalAccessor:
         freq: Frequency,
         weighted: bool = True,
         keep_weights: bool = False,
+        reference_period: Optional[Tuple[str, str]] = None,
         season_config: SeasonConfigInput = DEFAULT_SEASON_CONFIG,
     ):
         """Returns a Dataset with the climatology of a data variable.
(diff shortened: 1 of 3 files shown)


tomvothecoder (commented):
Hi @lee1043 and @pochedls, this PR for implementing `reference_period` in `climatology()` and `departures()` is ready for review.

I decided to not implement reference period subsetting for `average()` and `group_average()` because the user can easily do that themselves before calling the temporal APIs, e.g., `ds.isel(time=slice(....))`.

### Validation
In addition to the updated unit tests, below is the validation script that compares method 1 (@lee1043's [cookbook](https://github.com/xCDAT/xcdat_test/blob/main/cookbook/seasonal_departure_from_baseline_clim.ipynb)) to method 2 (this PR implementation) for calculating climatologies and departures relative to a reference period. The result of the script shows that **both methods produce identical results.**

```python
# flake8: noqa F401 E256
#%%
import matplotlib.pyplot as plt

import xcdat as xc

#%%
infile = "./input/HadISST_sst.nc"
data_var = "sst"  #  name of data variable

ds = xc.open_dataset(infile)
ds[data_var] = ds[data_var].where(ds[data_var] >= -0.5)  # exclude missing value

#%%
# =======================================
#  METHOD 1 - xCDAT + xarray
# =======================================
# 1. Get seasonal average of the full time series
ds_season = ds.temporal.group_average(data_var, freq="season")
ds_season

#%%
# 2. Temporal subset of original dataset to get reference climatology
ds_reference = ds.sel(time=slice("1971-01-01", "2000-12-31"))
ds_clim1 = ds_reference.temporal.climatology(data_var, freq="season")
ds_clim1[data_var].isel(time=0).plot(cmap="viridis")

#%%
# 3. Calculate departures
ds_departs1 = ds_season.copy()
ds_departs1[data_var + "_departure"] = ds_departs1[data_var].groupby(
    "time.season"
) - ds_clim1[data_var].groupby("time.season").mean("time")
ds_departs1[data_var + "_departure"].isel(time=0).plot()

# %%
# =======================================
#  METHOD 2 - xCDAT
# =======================================
# 1. Temporal subset of original dataset to get reference climatology
ds_clim2 = ds.temporal.climatology(
    data_var, freq="season", reference_period=("1971-01-01", "2000-12-31")
)

ds_clim2[data_var].isel(time=0).plot(cmap="viridis")

# Check equals method 1
ds_clim2["sst"].equals(ds_clim2["sst"])  # True

#%%
# 2. Calculate departures
ds_departs2 = ds.temporal.departures(
    data_var, freq="season", reference_period=("1971-01-01", "2000-12-31")
)
ds_departs2[data_var].isel(time=0).plot()

#%%
# =======================================
#  Compare Outputs
# =======================================
# Drop season sub-coordinates
ds_departs1_new = ds_departs1.drop_vars("season")

# Get the expected and resulting departures
method1_anom = ds_departs1_new["sst_departure"].rename("sst")
method2_anom = ds_departs2["sst"]
#%%
# Compare equality
method2_anom.equals(method1_anom)  # True

#%%
# Compare identical (drop attributes)
expected2 = method1_anom.copy()
result2 = method2_anom.copy()

expected2.attrs = {}
result2.attrs = {}

expected2.identical(result2)  # Truegithub
```

xcdat/temporal.py
```diff
@@ -375,6 +384,10 @@ def climatology(
         keep_weights : bool, optional
             If calculating averages using weights, keep the weights in the
             final dataset output, by default False.
+        reference_period : Optional[Tuple[str, str]], optional
+            The climatological reference period, which is a subset of the entire
+            time series. This parameter accepts a tuple of strings in the format
+            'yyyy-mm-dd'. For example, ``('1850-01-01', 1899-12-31')``.
```
pochedls, line 390:
```suggestion
        reference_period : Optional[Tuple[str, str]], optional
            The climatological reference period, which is a subset of the entire
            time series. This parameter accepts a tuple of strings in the format
            'yyyy-mm-dd'. For example, ``('1850-01-01', 1899-12-31')``. If no
            value is provided, the climatological reference period will be the
            full period covered by the dataset.
```

Simply adds: "If no value is provided, the climatological reference period will be the full period covered by the dataset."

xcdat/temporal.py
```diff
@@ -539,6 +548,11 @@ def departures(
         keep_weights : bool, optional
             If calculating averages using weights, keep the weights in the
             final dataset output, by default False.
+        reference_period : Optional[Tuple[str, str]], optional
+            The climatological reference period, which is a subset of the entire
+            time series and used for calculating departures. This parameter
+            accepts a tuple of strings in the format 'yyyy-mm-dd'. For example,
+            ``('1850-01-01', 1899-12-31')``.
```
pochedls, line 555:
```suggestion
        reference_period : Optional[Tuple[str, str]], optional
            The climatological reference period, which is a subset of the entire
            time series and used for calculating departures. This parameter
            accepts a tuple of strings in the format 'yyyy-mm-dd'. For example,
            ``('1850-01-01', 1899-12-31')``. If no value is provided, the 
            climatological reference period will be the full period covered by 
            the dataset.
```

Simply adds: "If no value is provided, the climatological reference period will be the full period covered by the dataset."

xcdat/temporal.py
```diff
@@ -611,48 +632,59 @@ def departures(
...
+        ds_avg = ds.temporal.group_average(
+            data_var, self._freq, weighted, keep_weights, season_config=season_config
+        )
+        dv_obs = ds_avg[data_var].copy()
+
+        # 2. Group the observation data variable by the departures frequency.
+        # This step allows us to perform xarray's grouped arithmetic to calculate
+        # departures.
         self._labeled_time = self._label_time_coords(dv_obs[self.dim])
         dv_obs_grouped = self._group_data(dv_obs)
```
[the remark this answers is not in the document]
pochedls: Thanks for catching and fixing this. But what if the input/output frequencies are the same? Should we still call `group_average()`?

pochedls (commented):
Overall, this PR seems to work great and makes sense to me. It takes me a long time to re-orient myself to the inner workings of the `.temporal` accessor so I'm hoping @lee1043 can also give this a careful review. 

* One thing I noticed is that calling `ds.temporal.departures` drops `time_bnds` – which I don't think needs to happen (I assume xarray doesn't know how to deal with this during the subtraction of the climatology). I think the time bounds can be added back in after subtraction. I noticed this in verifying that this PR was accurate (I get accuracy of order 10<sup>-4</sup> – 10<sup>-7</sup>, I think because I added time bounds back in using `add_missing_bounds`, which is problematic). 
* Note to others that I did need to update my dev environment to pass the unit tests
* Made some minor documentation suggestions
* Steps zero and one of `departures` call `_set_arg_attrs()`, `.copy()`, `_preprocess_dataset()`, and `group_average()`. This overlaps strongly with `_averager()` – maybe things could be streamlined here (though I see that some of the order-of-operations is different, which may be why they are different. 
* As noted, departures calls `group_average()` on each call – I think this may not be necesary if the input/output frequency are the same?

xcdat/temporal.py
```diff
@@ -669,15 +671,38 @@ def departures(
...
             dv_departs = dv_obs_grouped - dv_climo
             dv_departs = self._add_operation_attrs(dv_departs)
             ds_avg[data_var] = dv_departs
+
+            # The original time dimension name is restored after grouped
+            # arithmetic, so the labeled time dimension name is no longer needed
+            # and therefore dropped.
             ds_avg = ds_avg.drop_vars(self._labeled_time.name)
 
         if weighted and keep_weights:
```
tomvothecoder, line 708: In `departures()`, I updated the order of operations and added clearer comments to outline the steps and why they need to be done this specific way.

tomvothecoder:
>  * One thing I noticed is that calling `ds.temporal.departures` drops `time_bnds` – which I don't think needs to happen (I assume xarray doesn't know how to deal with this during the subtraction of the climatology). I think the time bounds can be added back in after subtraction. I noticed this in verifying that this PR was accurate (I get accuracy of order 10-4 – 10-7, I think because I added time bounds back in using `add_missing_bounds`, which is problematic).

Before this PR:
- `departures()` used to maintain `time_bnds` because the output frequency was the same as the input frequency (no `group_average()` call).

In this PR:
- Since we now call `group_average()`, `time_bnds` are dropped because the `time` coordinates are grouped which changes the `time` frequency and shape.

Solutions/Workarounds:
1. As you mentioned, get new `time_bnds` for the new grouped `time` coordinates using `add_missing_bounds()`. The accuracy _should_ improve with #412.
2. Keep a separate `time_original` dimension/coordinates to keep `time_bnds` in the dataset (otherwise we get `time_bnds` with `NaT` values since `time` frequency and shape changes).

---

>  * Steps zero and one of `departures` call `_set_arg_attrs()`, `.copy()`, `_preprocess_dataset()`, and `group_average()`. This overlaps strongly with `_averager()` – maybe things could be streamlined here (though I see that some of the order-of-operations is different, which may be why they are different.

I saw your point and played around with different implements using `_average()` to tried to find ways to streamline the implementation. I wasn't able to find a simple way to call `self._average()` without making significant modifications to how the `TemporalAccessor` class attributes are set (e.g., `self._mode`). **We can probably try refactoring this in the future if we wanted to.**

I updated the order-of-operations and made clearer comments in the `departures()` method so it is hopefully easier to understand. Please check out this [commit](https://github.com/xCDAT/xcdat/pull/417/commits/25832ddf5357bf043ecbd300aa03e71308840987) diff and let me know if those comments make sense.

---

>   * As noted, departures calls `group_average()` on each call – I think this may not be necesary if the input/output frequency are the same?

I also don't think we need to perform group averaging if the input/output frequencies are the same. **It still works either way** (and converts time coordinates to `cftime` if they aren't already), but it might improve performance by removing unnecessary operations.

We can try determining if `group_average()` is necessary by using the existing `_infer_freq()` method and comparing the result to the input `freq` argument.

xcdat/temporal.py
```diff
@@ -611,48 +639,88 @@ def departures(
...
-        # Group the observation data variable.
-        dv_obs = _get_data_var(ds, data_var)
-        self._labeled_time = self._label_time_coords(dv_obs[self.dim])
-        dv_obs_grouped = self._group_data(dv_obs)
+        # 3. Calculate the grouped average and climatology of the data variable.
+        # ----------------------------------------------------------------------
+        # The climatology and grouped average APIs are called on the copied
+        # dataset to create separate instances of the `TemporalAccessor` class.
+        # This is done to avoid overriding the attributes of the current
+        # instance of `TemporalAccessor` (set in step #1 above).
```
tomvothecoder, line 662: If the `departures()` performance or memory usage is ever an issue, we can look into refactoring these steps to avoid copying. Copying datasets multiple times might not be an issue though because [xr.Dataset.copy()](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.copy.html) defaults to `deep=False` (_"... a shallow copy of each of the component variable is made, so that the underlying memory region of the new dataset is the same as in the original dataset."_)

On the other hand, [xr.DataArray.copy()](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.copy.html) defaults to `deep=True`.

xcdat/temporal.py
```diff
@@ -611,48 +639,88 @@ def departures(
...
+        ds_obs = ds.copy()
+        inferred_freq = self._infer_freq()
+        if inferred_freq != freq:
+            ds_obs = ds_obs.temporal.group_average(
+                data_var,
+                freq,
+                weighted,
+                keep_weights,
+                season_config,
+            )
```
tomvothecoder, line 675: Added conditional for group averaging only if the input and output frequencies are the same. This should save on performance costs for this case.

Related to https://github.com/xCDAT/xcdat/pull/417#discussion_r1113628929 and https://github.com/xCDAT/xcdat/pull/417#pullrequestreview-1308232306.

PR #418, opened 2023-02-24
Functions to produce accurate time bounds

### LINKED ISSUE #412: [Feature]: Add functionality to accurately set time bounds

### LINKED ISSUE #426: [Doc]: Update TemporalAccessor docstrings with info on time bounds and decoded times

This comment (https://github.com/pydata/xarray/discussions/7580#discussioncomment-5227828) pointed out the temporal accessor docstrings don't explicitly mention using time bounds.

We should make it clear that time bounds and decoded times are required. Also we should mention how time bounds are used for averaging.

pochedls:
## Description
PR is intended to reproduce CDAT functionality (e.g., `setAxisTimeBoundsMonthly()`) to produce accurate time bounds for datasets with missing bounds. 

As cleanup tasks, this PR will also a) remove the width parameter from generic bound creation routines, b) only autogenerate bounds for lat/lon (others axes, e.g., time, must be specified). 

- Closes #412
- Closes #426 

## Checklist

If applicable:

### DIFF
--- a/xcdat/bounds.py
+++ b/xcdat/bounds.py
@@ -130,14 +130,17 @@ class BoundsAccessor:
         bounds to its coordinates if they don't exist. The coordinates must meet
         the following criteria in order to add bounds:
 
-          1. The axis for the coordinates are "X", "Y", "T", or "Z"
-          2. Coordinates are a single dimension, not multidimensional
-          3. Coordinates are a length > 1 (not singleton)
-          4. Bounds must not already exist.
-             * Determined by attempting to map the coordinate variable's
-             "bounds" attr (if set) to the bounds data variable of the same key.
-          5. Time axes must be composed of datetime-like objects (`np.datetime64`
-             or `cftime`).
+        1. The axis for the coordinates are "X", "Y", "T", or "Z"
+        2. Coordinates are a single dimension, not multidimensional
+        3. Coordinates are a length > 1 (not singleton)
+        4. Bounds must not already exist
+
+           * Coordinates are mapped to bounds using the "bounds" attr. For
+             example, bounds exist if ``ds.time.attrs["bounds"]`` is set to
+             ``"time_bnds"`` and ``ds.time_bnds`` is present in the dataset.
+
+        5. Time axes must be composed of datetime-like objects
+           (`np.datetime64` or `cftime`).
 
         If coordinates do not meet that criteria, bounds are not added for them.
 
@@ -261,12 +264,14 @@ class BoundsAccessor:
         To add bounds for an axis, its coordinates must be the following
         criteria:
 
-          1. The axis for the coordinates are "X", "Y", "T", or "Z"
-          2. Coordinates are single dimensional, not multidimensional
-          3. Coordinates are a length > 1 (not singleton)
-          4. Bounds must not already exist
-             * Determined by attempting to map the coordinate variable's
-             "bounds" attr (if set) to the bounds data variable of the same key
+        1. The axis for the coordinates are "X", "Y", "T", or "Z"
+        2. Coordinates are single dimensional, not multidimensional
+        3. Coordinates are a length > 1 (not singleton)
+        4. Bounds must not already exist
+
+           * Coordinates are mapped to bounds using the "bounds" attr. For
+             example, bounds exist if ``ds.time.attrs["bounds"]`` is set to
+             ``"time_bnds"`` and ``ds.time_bnds`` is present in the dataset.
 
         Parameters
         ----------
@@ -321,13 +326,16 @@ class BoundsAccessor:
         to add bounds for each of them if they don't exist. To add time bounds
         for the time axis, its coordinates must be the following criteria:
 
-          1. Coordinates are single dimensional, not multidimensional
-          2. Coordinates are a length > 1 (not singleton)
-          3. Bounds must not already exist
-             * Determined by attempting to map the coordinate variable's
-               "bounds" attr (if set) to the bounds data variable of the same key
-          4. If ``method=freq``, coordinates must be composed of datetime-like
-             objects (`np.datetime64` or `cftime`)
+        1. Coordinates are single dimensional, not multidimensional
+        2. Coordinates are a length > 1 (not singleton)
+        3. Bounds must not already exist
+
+           * Coordinates are mapped to bounds using the "bounds" attr. For
+             example, bounds exist if ``ds.time.attrs["bounds"]`` is set to
+             ``"time_bnds"`` and ``ds.time_bnds`` is present in the dataset.
+
+        4. If ``method=freq``, coordinates must be composed of datetime-like
+           objects (``np.datetime64`` or ``cftime``)
 
         Parameters
         ----------
@@ -336,9 +344,9 @@ class BoundsAccessor:
             "freq" or "midpoint".
 
             * "freq": Create time bounds as the start and end of each timestep's
-               period using either the inferred or specified time frequency
-               (``freq`` parameter). For example, the time bounds will be the
-               start and end of each month for each monthly coordinate point.
+              period using either the inferred or specified time frequency
+              (``freq`` parameter). For example, the time bounds will be the
+              start and end of each month for each monthly coordinate point.
             * "midpoint": Create time bounds using time coordinates as the
               midpoint between their upper and lower bounds.
 
@@ -362,11 +370,12 @@ class BoundsAccessor:
         end_of_month : bool, optional
             If ``freq=="month"``, this flag notes that the timepoint is saved
             at the end of the monthly interval (see Note), by default False.
-            Some timepoints are saved at the end of the interval, e.g., Feb. 1
-            00:00 for the time interval Jan. 1 00:00 - Feb. 1 00:00. Since this
-            method determines the month and year from the time vector, the
-            bounds will be set incorrectly if the timepoint is set to the end of
-            the time interval. For these cases, set ``end_of_month=True``.
+
+            * Some timepoints are saved at the end of the interval, e.g., Feb. 1
+              00:00 for the time interval Jan. 1 00:00 - Feb. 1 00:00. Since this
+              method determines the month and year from the time vector, the
+              bounds will be set incorrectly if the timepoint is set to the end of
+              the time interval. For these cases, set ``end_of_month=True``.
 
         Returns
         -------
@@ -418,11 +427,11 @@ class BoundsAccessor:
         loop over coordinates related to an axis and attempts to add bounds if
         they don't exist. If ancillary coordinates are present, "ValueError:
         Cannot generate bounds for coordinate variable 'height' which has a
-        length <= 1 (singleton)" is raised. For the purpose of adding bounds,
-        we temporarily drop any ancillary singletons from dimension 
-        coordinates before looping over those coordinates. Ancillary 
-        singletons will still be present in the final xr.Dataset object to 
-        maintain its integrity.
+        length <= 1 (singleton)" is raised. For the purpose of adding bounds, we
+        temporarily drop any ancillary singletons from dimension coordinates
+        before looping over those coordinates. Ancillary singletons will still
+        be present in the final Dataset object to maintain the Dataset's
+        integrity.
 
         Parameters
         ----------
--- a/xcdat/bounds.py
+++ b/xcdat/bounds.py
@@ -115,7 +115,7 @@ class BoundsAccessor:
             )
         )
 
-    def add_missing_bounds(self) -> xr.Dataset:
+    def add_missing_bounds(self, axes: List[CFAxisKey] = ["X", "Y"]) -> xr.Dataset:
         """Adds missing coordinate bounds for supported axes in the Dataset.
 
         This function loops through the Dataset's axes and attempts to adds
@@ -129,12 +129,16 @@ class BoundsAccessor:
              * Determined by attempting to map the coordinate variable's
              "bounds" attr (if set) to the bounds data variable of the same key.
 
+        Parameters
+        ----------
+        axes : List[str], optional
+            List of CF axes that function should operate on, default ["X", "Y"].
+
         Returns
         -------
         xr.Dataset
         """
         ds = self._dataset.copy()
-        axes = CF_ATTR_MAP.keys()
 
         for axis in axes:
             try:
--- a/xcdat/bounds.py
+++ b/xcdat/bounds.py
@@ -14,7 +14,7 @@ from xarray.coding.cftime_offsets import get_date_type
 from xcdat.axis import CF_ATTR_MAP, CFAxisKey, get_dim_coords
 from xcdat.dataset import _get_data_var
 from xcdat.logger import setup_custom_logger
-from xcdat.temporal import _month_add
+from xcdat.temporal import _infer_freq, _month_add
 
 logger = setup_custom_logger(__name__)
 
@@ -131,6 +131,7 @@ class BoundsAccessor:
           4. Bounds must not already exist.
              * Determined by attempting to map the coordinate variable's
              "bounds" attr (if set) to the bounds data variable of the same key.
+          5. Time axes should be composed of `cftime` objects.
 
         Parameters
         ----------
@@ -245,6 +246,7 @@ class BoundsAccessor:
           4. Bounds must not already exist.
              * Determined by attempting to map the coordinate variable's
              "bounds" attr (if set) to the bounds data variable of the same key.
+          5. Time axes should be composed of `cftime` objects.
 
         Parameters
         ----------
@@ -345,7 +347,10 @@ class BoundsAccessor:
 
         Notes
         -----
-        Based on [1]_ ``iris.coords._guess_bounds`` and [2]_ ``cf_xarray.accessor.add_bounds``
+        Based on [1]_ ``iris.coords._guess_bounds`` and [2]_ ``cf_xarray.accessor.add_bounds.``
+        For temporal coordinates `_create_bounds` will attempt to set the bounds to the start
+        and end of each time step's period. Time axes are expected to be composed of `cftime`
+        objects.
 
         References
         ----------
@@ -374,7 +379,8 @@ class BoundsAccessor:
         # the values of  `diffs` must be casted from `dtype="timedelta64[ns]"`
         # to `timedelta` objects.
         if axis == "T" and issubclass(type(coord_var.values[0]), cftime.datetime):
-            diffs = pd.to_timedelta(diffs)
+            bounds = get_time_bounds(coord_var)
+            return bounds
 
         # width parameter: determines bounds location relative to midpoints
         width = 0.5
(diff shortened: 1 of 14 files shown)


tomvothecoder:
This is an awesome turnaround! Thanks for getting to this so fast Steve. Feel free to tag me when it is ready.

I can do code review and high-level usability testing, and maybe @lee1043 can a more in-depth usability testing if he's available?

pochedls:
> @pochedls thank you for the PR! @tomvothecoder yes I can also take a look. I will work on it and keep you all posted.

This PR is not yet ready for review.

lee1043:
> This PR is not yet ready for review.

@pochedls Okay, let us know when it's ready.

tomvothecoder:
Hey @pochedls I'm following about using `TemporalAccessor._infer_freq()` in this PR. 

In the current implementation, `self._set_data_var_attrs()` needs to be called first before calling `_infer_freq()` because this method requires the `self.dim` class attribute (below)

https://github.com/xCDAT/xcdat/blob/1efb8074e2cbec99c2b410b001c979926d8a44df/xcdat/temporal.py#L673-L674

A possible workaround is to add an optional `dim` kwarg that is used instead of `self.dim` if it is passed.
```python
ds.temporal._infer_freq(dim="time")
```

pochedls (commented):
This is intended to be a summary of the changes made to help with the review. 

**Major changes**

* xcdat no longer adds temporal bounds by default (it now only adds lat/lon by default). To add time (or height) bounds you need to include `T` in the new `bound_axes` argument in `open_dataset` functionality (e.g., `bounds_axes=[‘X’, ‘Y’, ’T’]`) or the `axes` argument in `add_missing_bounds`. 
* Functionality to add bounds now handles temporal bounds as a special case (see trace below); we set the time bounds to the expected start/end of the time period (e.g., the start/end of the month for monthly data).
* A number of functions of the form `get_FREQ_time_bounds` are added. They are wrapped by `get_time_bounds` which is intended to call the correct `get_FREQ_time_bounds` function.
* `infer_freq` is now a function (rather than a temporal accessor method)
* Added a new `_month_add` function to add N months to a cftime object (N can be negative for subtraction)

**API Changes**

* Modified
  * `open_dataset` now adds optional `bounds_axes` argument [`dataset.py`]
  * `open_mfdataset` now adds optional `bounds_axes` argument [`dataset.py`]
  * `postprocess_dataset` now adds optional `bounds_axes` argument [`dataset.py`]
  * `_infer_freq(time_coord)` now requires the time coordinate axis and is no longer an accessor method [`temporal.py`]
  * `add_missing_bounds` remove `width` argument and accept optional `axes` argument [`bounds.py`]
  * `add_bounds` remove `width` argument [`bounds.py`]
  * `_create_bounds` remove `width` argument [`bounds.py`]
* New
  * `_month_add(times, delta, calendar)` [`temporal.py`]
  * `get_yearly_time_bounds(time)` [`bounds.py`]
  * `get_monthly_time_bounds(time)` [`bounds.py`]
  * `get_daily_time_bounds(time, frequency=1)` [`bounds.py`]
  * `get_time_bounds(time)` [`bounds.py`]

**Trace**

`open_dataset` / `open_mfdataset` 
&emsp; `postprocess_dataset` 
&emsp; &emsp; `add_missing_bounds` 
&emsp; &emsp; &emsp; `_create_bounds` 
&emsp; &emsp; &emsp; &emsp; `get_time_bounds` 
&emsp; &emsp; &emsp; &emsp; &emsp;`get_yearly_time_bounds` / `get_monthly_time_bounds` / `get_daily_time_bounds`
&emsp; &emsp; &emsp; &emsp; &emsp; &emsp;`_infer_freq` / `_month_add`

* Add optional `bounds_axes` argument to open dataset functionality: `open_dataset(..., bounds_axes=[..., "T"])` or `open_mfdataset(..., bounds_axes=[..., "T"])`, which call `_postprocess_dataset(..., bounds_axes=[..., "T"])`. By default `bounds_axes=["X", "Y"]`. The rest of this "trace" assumes `"T"` is included in `bounds_axes`.
* Post-processing routine calls `add_missing_bounds`, which now only operates on the `bounds_axes` passed into the function (rather than all `X`/`Y`/`T`/`Z` coordinate variables in dataset). 
* `add_missing_bounds` calls `_create_bounds`. If the coordinate variable being operated on is time, the routine will call the new `get_time_bounds` function to create the time bounds (instead of using the generic bounds creation logic).
* `get_time_bounds` calls the refactored `_infer_freq`. Based on the inferred frequency either `get_yearly_time_bounds`, `get_monthly_time_bounds`, or `get_daily_time_bounds` is called. The function then creates and returns a time coordinate data array. 

**xcdat/bounds.py**

* Modify `add_missing_bounds` such that `width` is no longer an argument (always 0.5) and an optional `axes` argument is added with `[“X”, “Y”]` set as the default. 
* Modify `add_bounds` to remove `width` argument and to accept `bounds` as an optional argument. If `bounds` are passed in it will add these bounds to the dataset rather than generating new bounds. 
* Modify `_create_bounds` to remove `width` argument. Modifies functionality: if `coord_var` is a temporal axis the routine now calls `get_time_bounds` to compute the temporal bounds. 
* Add `get_yearly_time_bounds`, `get_monthly_time_bounds`, and `get_daily_time_bounds` based on CDAT functionality. These functions create time bounds for annual, monthly, and daily time series. `get_daily_time_bounds` will also calculate sub-daily (but not sub-hourly) time bounds if a `frequency` argument is supplied. 
* Add `get_time_bounds` function to call the correct `get_FREQ_time_bounds` function based on the refactored `_infer_freq` function.

**xcdat/dataset.py**
* `open_dataset`, `open_mfdataset`, and `_postprocess_dataset` are modified to accept an optional `bounds_axes` argument, which defines the axes for which xcdat ensures there are bounds

**xcdat/temporal**

* Add `_month_add` function to add N months to cftime objects
* Change `_infer_freq` to a function, which now requires time_coords as an argument
* Update `average` to pass in the now required time coordinate axis argument to `_infer_freq`

**tests/fixtures.py**

* Add sub-hourly, hourly, daily, and yearly time axes and bounds
* Refactor `generate_dataset` so that inputs to `xr.Dataset` are decided with conditional and then `xr.Dataset` is called with those inputs (reduces code length a little)
* Add `generate_dataset_by_frequency` to generate datasets of varying frequencies in order to test functionality to add temporal bounds. I initially added a frequency argument to `generate_dataset`, but it did not pass the pre-commit hooks (one of them complained the function was overly complex: Flake8 C901). 

**tests/test_bounds.py**

* add test_time_bounds_not_added_to_the_dataset_if_not_specified to ensure that time bounds are not added by default when calling `add_missing_bounds`
* Fixed `test_returns_bounds_for_dataset_with_time_coords_as_cftime_objects` test (which was not setting time bounds correctly)
* Add `test_get_monthly_bounds`, `test_get_yearly_bounds`, `test_get_daily_bounds`, `test_get_hourly_bounds` which check that they each compute the correct time bounds. 
* Add `test_generate_monthly_bounds_for_eom_set_true`, which is the same as above, except the `end_of_month` flag for `get_monthly_time_bounds` is set to `True`. 
* Add `test_generic_get_time_bounds_function` which checks to ensure `get_time_bounds` adds the correct time bounds to the dataset for yearly, monthly, daily, and hourly data. Also ensures that sub-hourly data throws a ValueError. 

**tests/test_dataset.py**

* Change `test_adds_missing_lat_and_lon_bounds` to `test_adds_missing_lat_and_lon_bounds_does_not_add_time_bounds` to ensure that time bounds are not added by default when opening a dataset without time bounds
* Add `test_adds_missing_lat_and_lon_and_time_bounds` to ensure time bounds are added when `T` is included in the `bounds_axes` argument in `open_dataset`. 

**tests/test_temporal.py**

* Add `test_month_add_plus_minus` function to test addition / subtraction of months with `_month_add`

pochedls:
One unresolved aspect of this PR is that I needed to add `# type: ignore` in several `get_XYZ_time_bounds()` functions, e.g., [here](https://github.com/xCDAT/xcdat/pull/418/files#diff-6a74d16440d82ae43ed458dc26fafd0427c8568b34915e698fce57b2c9a8a7e1R467).

tomvothecoder:
Hey @pochedls, thanks for your excellent PR summary above. It is really helpful. 

I am walking through your code changes and will do some usability testing. @lee1043 can you do more in-depth usability testing as well?

I am hoping to get this PR merged for the v0.5.0 release (around mid next week).

xcdat/bounds.py
```diff
@@ -429,3 +432,227 @@ def _validate_axis_arg(self, axis: CFAxisKey):
...
+    # Create the bounds data array
+    time_bnds = xr.DataArray(
+        name=f"{time.name}_bnds",
+        data=time_bnds,
+        coords={time.name: time},
+        dims=[*time.dims, "bnds"],
+        attrs={"xcdat_bounds": "True"},
+    )
+
+    return time_bnds
```
tomvothecoder, line 658: @pochedls I had a few questions about these new time bounds APIs:

1. Does you think it makes sense to define these functions as `BoundsAccessor` methods? 
2. Are they intended to be public APIs?
3. Does it make sense to prepend the API names with `create_` instead of `get_`?
   - `add_bounds()` calls `_create_bounds()`, which calls `get_time_bounds()`. Maybe renaming `get_time_bounds()` to `_create_time_bounds()` is more appropriate?
   - There might be a bit of confusion between the usage for the `get_bounds()` method and the `get_time_bounds()` function because they both operate on the time axis.

pochedls: > 1. Does you think it makes sense to define these functions as BoundsAccessor methods?

These are currently not `BoundsAccessor` methods, right?

I think I initially made them accessor methods and then felt I was making life complicated. Accessor methods were either a) unnecessary (if you pass in the time coordinate, you don't need the `dataset` object) or b) a lot of extra work (because then the method has to figure out what time coordinate axis you care about in each of these functions). I guess you could also pass in the `.name` attribute of the time coordinate you care about (and then use the dataset to fetch the time coordinate axis), but then why not just pass in the time axis? 

> 2. Are they intended to be public APIs? 

Yes – I think they should be accessible. They seem like they could be useful functions for users to be able to call. 

> 3. Does it make sense to prepend the API names with `create_` instead of `get_`? 

I think you are right.

tomvothecoder: > I think I initially made them accessor methods and then felt I was making life complicated. 

From what I understand, there are two scenarios here:

1.  We want to add time bounds directly to the Dataset - use `ds.temporal.add_bounds(axis="T")`
    - Loops over _all_ time coordinates and creates bounds for each (calls `create_time_bounds()`)
    - Returns an `xr.Dataset`
    - Example usage: `ds = ds.bounds.add_bounds(axis="T")`
2.  We want to create time bounds for a _specific_ time coordinate variable for more granular operations - use `create_time_bounds()`
    - User specifies the exact time coordinate variable to create bounds.
    - Returns an `xr.DataArray`
    - Example usage: `time_bounds = create_time_bounds(ds.time)`

---

If we want to support scenario 2, I can see how the implementation/usage of `create_time_bounds()` is cumbersome if it were a `BoundsAccessor` method. `BoundsAccessor` methods are intended to return an `xr.Dataset`, not an `xr.DataArray`.

If this all sounds correct, it makes sense to keep your implementation of `create_time_bounds()` as a **function**.

pochedls: I think scenario 1 would be ideal, but I think it would also be useful to allow users to call the methods individually. CDAT didn't automatically create time bounds (as far as I know) and I worry there might be situations where we cannot automatically infer the frequency and add the time bounds. In those cases, enabling users to choose between create_daily/monthly/yearly_bounds and create_time_bounds would be helpful. Doesn't the code, as written, allow you to do both (call individual functions or infer the bounds through `add_bounds()`)? Is there a better way to enable both? 

Semi-relevant: I had initially modified [`create_bounds` to accept a dataarray](https://github.com/xCDAT/xcdat/pull/418/commits/ffb2688d7f33ed1362a8b335e00e2090b548d6b7#diff-6a74d16440d82ae43ed458dc26fafd0427c8568b34915e698fce57b2c9a8a7e1R232) if the user wanted to supply their own bounds (e.g., from `create_monthly_time_bounds`). I undid this since I expected bounds would generally be added through `add_bounds`.

xcdat/bounds.py
```diff
@@ -429,3 +432,227 @@ def _validate_axis_arg(self, axis: CFAxisKey):
...
+
+
+def get_time_bounds(time):
+    """Sets the time bounds for a time coordinate axis.
+
+    Parameters
+    ----------
+    time : xr.DataArray
+        The temporal coordinate variable for the axis.
+    frequency : int, optional, default 1
```
lee1043, line 618: Here, the function description says the function takes two parameters: `time` and `frequency` but it looks the function does not expect getting `frequency` from external, and get `freq` from `_infer_freq` internally.

lee1043:
Thank you, @pochedls, for making this PR. Could you help me learning how to use this new capability in following scenario? A sample snippet of codes would be very helpful.

- Open dataset that has monthly frequency and single variable. Each time step stored at the beginning of month (1/1, 2/1, 3/1, ...)
- I want to set bounds as ([1/1-1/31, 2/1-2/28, 3/1-3/31, ...])

I guess some combination of `create_bounds` and `get_monthly_time_bounds` need to be used, but I am yet to have a good idea how to use them. I think we may want to rename `get_monthly_time_bounds` to `set_monthly_time_bounds` or `create_monthly_time_bounds` because `get_monthly_time_bounds` gives me an impression that I might retrieve time axis from the original dataset.

pochedls:
> Thank you, @pochedls, for making this PR. Could you help me learning how to use this new capability in following scenario? A sample snippet of codes would be very helpful.
> 
>     * Open dataset that has monthly frequency and single variable. Each time step stored at the beginning of month (1/1, 2/1, 3/1, ...)
> 
>     * I want to set bounds as ([1/1-1/31, 2/1-2/28, 3/1-3/31, ...])
> 
> 
> I guess some combination of `create_bounds` and `get_monthly_time_bounds` need to be used, but I am yet to have a good idea how to use them. I think we may want to rename `get_monthly_time_bounds` to `set_monthly_time_bounds` or `create_monthly_time_bounds` because `get_monthly_time_bounds` gives me an impression that I might retrieve time axis from the original dataset.

I think there are two ways to do this (90% sure – I don't have this PR open on my machine right now):

1. On open: `ds = xc.open_dataset(..., bounds_axes=['X', 'Y', 'T'])`
2. After open: `ds = ds.add_missing_bounds(axes=['X', 'Y', 'T'])`

@tomvothecoder also suggested renaming `get_time_bounds` functions to `create_time_bounds` (I agree with both of you).

xcdat/bounds.py
```diff
@@ -429,3 +432,227 @@ def _validate_axis_arg(self, axis: CFAxisKey):
...
+        Flag to note set the number of timepoints per day.
```
pochedls, line 619:
```suggestion
```

Suggested edit based on @lee1043's catch (frequency is not an input).

xcdat/bounds.py
pochedls: Thanks @lee1043! I suggested removing this. I think if this inference becomes problematic we could add it back as an optional argument (and infer frequency if the optional argument is not provided).

lee1043:
>  1. On open: ds = xc.open_dataset(..., bounds_axes=['X', 'Y', 'T'])
> 2. After open: ds = ds.bounds.add_missing_bounds(axes=['X', 'Y', 'T'])

@pochedls In the above my example scenario how do I know `add_missing_bounds` set time bounds as begin/end of month instead of -15 and +15 days from the time of each time step?

pochedls:
> > 1. On open: ds = xc.open_dataset(..., bounds_axes=['X', 'Y', 'T'])
> > 2. After open: ds = ds.bounds.add_missing_bounds(axes=['X', 'Y', 'T'])
> 
> @pochedls In the above my example scenario how do I know `add_missing_bounds` set time bounds as begin/end of month instead of -15 and +15 days from the time of each time step?

`add_missing_bounds` effectively calls `get_time_bounds` (to be renamed `create_time_bounds`), which infers that the frequency is monthly and then calls `get_monthly_time_bounds`. So the bounds functionality now is intended to figure out the appropriate time bounds based on frequency (rather than doing the +/- 15 day stuff). Do you think the +/- 15 day stuff should be preserved?

lee1043:
> `add_missing_bounds` effectively calls `get_time_bounds` (to be renamed `create_time_bounds`), which infers that the frequency is monthly and then calls `get_monthly_time_bounds` (to be renamed `create_monthly_time_bounds`). So the bounds functionality now is intended to figure out the appropriate time bounds based on frequency (rather than doing the +/- 15 day stuff). Do you think the +/- 15 day stuff should be preserved?

@pochedls thank you for explaining. I was thinking `add_missing_bounds` add for example `time_bnds` data array when it was not existing in the original dataset, but `add_missing_bounds` will **update** bounds information as well. Is this correct?

pochedls:
> > `add_missing_bounds` effectively calls `get_time_bounds` (to be renamed `create_time_bounds`), which infers that the frequency is monthly and then calls `get_monthly_time_bounds` (to be renamed `create_monthly_time_bounds`). So the bounds functionality now is intended to figure out the appropriate time bounds based on frequency (rather than doing the +/- 15 day stuff). Do you think the +/- 15 day stuff should be preserved?
> 
> @pochedls thank you for explaining. I was thinking `add_missing_bounds` add for example `time_bnds` data array when it was not existing in the original dataset, but `add_missing_bounds` will **update** bounds information as well. Is this correct?

No – I don't think `add_missing_bounds` replaces bounds if they are present in the dataset. To do that you'd have to drop the existing bounds first.

xcdat/bounds.py
tomvothecoder: Thanks for clarifying Steve. It sounds like we're on the same page in regards to the usefulness of supporting both scenarios. 

I'll try to see if there is an intuitive way of implementing and using `create_time_bounds()` as a `BoundsAccessor` method. If I can't find one, I'm happy to keep your function-based implementation.

tomvothecoder (commented):
Here's my initial review for this PR with some changes ([link](https://github.com/xCDAT/xcdat/pull/418/commits/031316b5096f7ae625de40974f34dd08026c589c) to diff with latest commit vs. previous commit).

I'm exploring the potential conversion of the `create` functions to accessor methods in a separate review.

Summary of changes:

- Rename `get` time bounds functions to `create`
- Update docstrings and comments for `create` time bounds functions
- Add TODO comment to raise exception when freq cannot be inferred in `_infer_freq`
- Add `create` time bounds functions to top-level xcdat module `__init__.py` and API references `api.rst`

xcdat/dataset.py
```diff
@@ -41,6 +41,7 @@ def open_dataset(
     path: str,
     data_var: Optional[str] = None,
     add_bounds: bool = True,
+    bounds_axes: Optional[List[str]] = ["X", "Y"],
```
tomvothecoder, line 44: What do you think about combining `add_bounds` and `bounds_axes` into a single parameter (e.g., `add_bounds = ["X, "Y"]`, also accepts `None`)?

**I'll need to think about this more** to see if it is confusing or makes the API less intuitive compared to having two separate parameters.

xcdat/bounds.py
tomvothecoder: I fixed this in my latest commit (https://github.com/xCDAT/xcdat/pull/418/commits/031316b5096f7ae625de40974f34dd08026c589c)

xcdat/bounds.py
tomvothecoder: I fixed this in my latest commit (https://github.com/xCDAT/xcdat/commit/031316b5096f7ae625de40974f34dd08026c589c)

xcdat/bounds.py
tomvothecoder: After some thought, I wasn't able to find an _intuitive_ accessor method implementation of `create_time_bounds()`. I am leaning towards keeping your implementation.

### 1. Function approach (current)

Example:
```python

def create_time_bounds(time: xr.DataArray) -> xr.DataArray
	pass

def create_yearly_time_bounds(self, xr.DataArray) -> xr.DataArray
	pass

def create_monthly_time_bounds(self, xr.DataArray) -> xr.DataArray
	pass

def create_daily_time_bounds(self, xr.DataArray) -> xr.DataArray
	pass

ds["time_bnds"] = create_time_bounds(ds.time)
ds["time_bnds"] = create_yearly_time_bounds(ds.time)
ds["time_bnds"] = create_monthly_time_bounds(ds.time)
ds["time_bnds"] = create_daily_time_bounds(ds.time)
```

Pros:
  - Simpler API naming convention (no need to reference `ds.bounds`)
  - User references the coordinate variable directly rather than xCDAT checking the dataset for the coordinate variable  -- earlier validation in the pipeline
  - `ds.bounds.add_bounds()` will handle _most_ cases for adding bounds -- these functions are for edge cases or users who want finer grain control

Cons:
 - Functionality is related to the `BoundsAccessor` class so grouping kinda makes sense

### 2. Accessor approach

Example:
```python
def create_time_bounds(self, time: str) -> xr.DataArray
	pass

def create_yearly_time_bounds(self, time:str) -> xr.DataArray
	pass

def create_monthly_time_bounds(self, time:str) -> xr.DataArray
	pass

def create_daily_time_bounds(self, time:str) -> xr.DataArray
	pass

ds["time_bnds"] = ds.bounds.create_time_bounds("time")
ds["time_bnds"] = ds.bounds.create_yearly_time_bounds("time")
ds["time_bnds"] = ds.bounds.create_monthly_time_bounds("time")
ds["time_bnds"]  = ds.bounds.create_daily_time_bounds("time")
```
Pros:
  - Encapsulates related bounds methods together
  - References the dataset and attempts to grab `ds.time` coordinate variable

Cons:
  - Verbose method naming convention makes it tedious to call when it's under the `.bounds` namespace
  - Returns a DataArray -- user needs to set bounds as data variable in the Dataset regardless if it were method or a function (e.g., `ds["time_bnds"] = ...`) 
  - We don't reference any class attribute in the methods except for `self._dataset` to get the `ds.time` coordinate variable

xcdat/bounds.py
```diff
@@ -560,91 +573,86 @@ def get_daily_time_bounds(time, frequency=1):
...
+        time_bnds = create_daily_time_bounds(time)
+    elif freq == "hour":
+        # Determine the daily frequency for generating time  bounds.
         diff = time.values[1] - time.values[0]
         hrs = diff.seconds / 3600
         daily_freq = int(24 / hrs)
-        # get sub-daily time bounds
-        time_bnds = get_daily_time_bounds(time, frequency=daily_freq)
+
+        time_bnds = create_daily_time_bounds(time, freq=daily_freq)  # type: ignore
```
tomvothecoder, line 656: I forgot to mention that we only need to call `create_daily_time_bounds()` once here.

xcdat/bounds.py
```diff
@@ -560,91 +573,86 @@ def get_daily_time_bounds(time, frequency=1):
...
         attrs={"xcdat_bounds": "True"},
     )
 
-    return time_bnds
+    return da_time_bnds
 
 
-def get_time_bounds(time):
-    """Sets the time bounds for a time coordinate axis.
+def create_time_bounds(time: xr.DataArray) -> xr.DataArray:
```
tomvothecoder, line 620: This higher-level function should ordered based on the call stack and ahead of its children functions (`create_X_time_bounds()`).

pochedls:
Stumbled on an xarray frequency inference method working on something else...parking here for now:

* https://github.com/pydata/xarray/blob/main/xarray/coding/frequencies.py#L59-L100 
* https://docs.xarray.dev/en/stable/generated/xarray.infer_freq.html#xarray.infer_freq

tomvothecoder:
> Stumbled on an xarray frequency inference method working on something else...parking here for now:
> 
>     * [pydata/xarray@`main`/xarray/coding/frequencies.py#L59-L100](https://github.com/pydata/xarray/blob/main/xarray/coding/frequencies.py?rgh-link-date=2023-03-16T05%3A36%3A01Z#L59-L100)
> 
>     * [docs.xarray.dev/en/stable/generated/xarray.infer_freq.html#xarray.infer_freq](https://docs.xarray.dev/en/stable/generated/xarray.infer_freq.html#xarray.infer_freq)

Thanks for this! I'll get in the habit of checking xarray or other packages for utilities (not just functionalities) before I implement them in xcdat.

I don't remember if I ran in to `xarray.infer_freq()`. Let's investigate to see if `xcdat.infer_freq()` is a redundant/obsolete version of `xarray.infer_freq()` and if it can be dropped from xcdat.

pochedls (commented):
@tomvothecoder – I'm good with this PR and look forward to reviewing the next PR (adding this functionality into accessors). 

I think adding accessor methods will be necessary – there are some items that can't be figured out automatically (e.g., `end_of_month = True`) and it would be a pain to have these options passed around through every intermediate function. Instead, users in these special situations should open their dataset and create the time bounds manually with the method and options of their choice (create_XYZ_time_bounds).

xcdat/bounds.py
pochedls: You mean by combining "hour" and "day"? I kind of like the logical separation here.

xcdat/dataset.py
pochedls: I'd be okay with this (assuming it defaults to `["X", "Y"]`).

xcdat/bounds.py
pochedls: Note that I didn't add to this PR again – so this is still a to do item.

xcdat/dataset.py
tomvothecoder: This would be considered a breaking change IMO. I assume most users keep the default `add_bounds=True` argument, and xCDAT is not at v1.0.0 yet so we can kinda get away with it (for now).

I'll take a look at this in PR #434.

tomvothecoder:
> Stumbled on an xarray frequency inference method working on something else...parking here for now:
> 
>     * [pydata/xarray@`main`/xarray/coding/frequencies.py#L59-L100](https://github.com/pydata/xarray/blob/main/xarray/coding/frequencies.py?rgh-link-date=2023-03-16T05%3A36%3A01Z#L59-L100)
> 
>     * [docs.xarray.dev/en/stable/generated/xarray.infer_freq.html#xarray.infer_freq](https://docs.xarray.dev/en/stable/generated/xarray.infer_freq.html#xarray.infer_freq)

Apparently I already investigated `xarray.infer_freq` when I implemented `xcdat.temporal._infer_freq` according to the docstring. I guess I gave myself a hard time for thinking I didn't look at xarray first lol.

https://github.com/xCDAT/xcdat/blob/d0a48920e84af3a80375dad64918ac1c9b7a0097/xcdat/temporal.py#L725-L740

---
`xarray.infer_freq` calls `pandas.infer_freq` under the hood, which returns frequencies in a string format based on a set list: https://pandas.pydata.org/docs/user_guide/timeseries.html#dateoffset-objects 

This format is not useful for `groupby` operations, so that's why I needed to implement my own utility.

Example:
```python

print(dataset.time)
<xarray.DataArray 'time' (time: 5)>
array(['2000-01-01T00:00:00.000000000', '2001-01-01T00:00:00.000000000',
       '2002-01-01T00:00:00.000000000', '2003-01-01T00:00:00.000000000',
       '2004-01-01T00:00:00.000000000'], dtype='datetime64[ns]')
Coordinates:
  * time     (time) datetime64[ns] 2000-01-01 2001-01-01 ... 2004-01-01
Attributes:
    axis:           T
    long_name:      time
    standard_name:  time
    bounds:         time_bnds

# Not useful for groupby operations
xr.infer_freq(dataset.time)
'AS-JAN'

# Useful for groupby operations
xcdat.temporal_infer_freq(dataset.time)
"year"
```

tomvothecoder:
Hey @lee1043, is ready for your usability testing. Thanks!

tomvothecoder:
Here's the API reference for `create_time_bounds()` to get you rolling:
https://xcdat.readthedocs.io/en/feature-412-accurate-time-bounds/generated/xarray.Dataset.bounds.add_time_bounds.html

lee1043:
> Hey @lee1043, is ready for your usability testing. Thanks!

Thank you @tomvothecoder @pochedls! I will work on this tomorrow.

xcdat/bounds.py
```diff
@@ -278,13 +301,152 @@ def add_bounds(self, axis: CFAxisKey, width: float = 0.5) -> xr.Dataset:
...
+
+        This method loops over the time axis coordinate variables and attempts
+        to add bounds for each of them if they don't exist. To add time bounds
+        for an axis, its coordinates must be the following criteria:
+
+          1. Coordinates are single dimensional, not multidimensional
+          2. Coordinates are a length > 1 (not singleton)
+          3. Bounds must not already exist
+             * Determined by attempting to map the coordinate variable's
+             "bounds" attr (if set) to the bounds data variable of the same key
```
tomvothecoder, line 328:
```suggestion
          3. Bounds must not already exist
             * Determined by attempting to map the coordinate variable's
              "bounds" attr (if set) to the bounds data variable of the same key
```

xcdat/bounds.py
```diff
@@ -278,13 +301,152 @@ def add_bounds(self, axis: CFAxisKey, width: float = 0.5) -> xr.Dataset:
...
+        method: Literal["freq", "midpoint"],
+        freq: Optional[Literal["year", "month", "day", "hour"]] = None,
+        daily_subfreq: Optional[Literal[1, 2, 3, 4, 6, 8, 12, 24]] = None,
+        end_of_month: bool = False,
+    ) -> xr.Dataset:
+        """Add bounds for an axis using its coordinate points.
+
+        This method loops over the time axis coordinate variables and attempts
+        to add bounds for each of them if they don't exist. To add time bounds
+        for an axis, its coordinates must be the following criteria:
```
tomvothecoder, line 322:
```suggestion
        to add bounds for each of them if they don't exist. To add time bounds
        for the time axis, its coordinates must be the following criteria:
```

xcdat/bounds.py
```diff
@@ -278,13 +301,152 @@ def add_bounds(self, axis: CFAxisKey, width: float = 0.5) -> xr.Dataset:
...
+        "height" singleton coordinate will be attached to "time" coordinates
+        even though "height" is related to the "Z" axis, not the "T" axis.
+
+        This is an undesirable behavior in xCDAT because the add bounds methods
+        loop over coordinates related to an axis and attempts to add bounds if
+        they don't exist. If ancillary coordinates are present, "ValueError:
+        Cannot generate bounds for coordinate variable 'height' which has a
+        length <= 1 (singleton)" is raised. To work around this Xarray behavior,
+        we drop the ancilliary singleton coordinates before adding bounds. Refer
+        to [1]_ for more info on this behavior.
```
tomvothecoder, line 422:
```suggestion
        length <= 1 (singleton)" is raised. For the purpose of adding bounds,
        we temporarily drop any ancillary singletons from dimension 
        coordinates before looping over those coordinates. Ancillary 
        singletons will still present in the final xr.Dataset object to maintain
        its integrity. Refer to [1]_ for more info on this behavior.
```

xcdat/bounds.py
```diff
@@ -278,13 +301,152 @@ def add_bounds(self, axis: CFAxisKey, width: float = 0.5) -> xr.Dataset:
...
+    def _drop_ancillary_singleton_coords(
+        self, coord_vars: Union[xr.Dataset, xr.DataArray]
+    ) -> Union[xr.Dataset, xr.DataArray]:
+        """Drop ancillary singleton coordinates from dimension coordinates.
+
+        Xarray coordinate variables retain all coordinates from the parent
+        object. This means if singleton coordinates exist, they are attached to
+        dimension coordinates as ancillary coordinates. For example, the
+        "height" singleton coordinate will be attached to "time" coordinates
+        even though "height" is related to the "Z" axis, not the "T" axis.
```
tomvothecoder, line 414:
```suggestion
        Xarray coordinate variables retain all coordinates from the parent
        object. This means if singleton coordinates exist, they are attached to
        dimension coordinates as ancillary coordinates. For example, the
        "height" singleton coordinate will be attached to "time" coordinates
        even though "height" is related to the "Z" axis, not the "T" axis.
        Refer to [1]_ for more info on this Xarray behavior.
```

lee1043:
@tomvothecoder @pochedls I confirm the `add_time_bounds` function works as expected and the documentation is very helpful! Many thanks for this PR! I agree to merge this PR to the main.

xcdat/dataset.py
tomvothecoder: Hey @pochedls  and @lee1043, I removed `bounds_axes` and refactored `add_bounds` to accept a list axes string, `None`, and `False`.

I kept `False` as an option for API backwards compatibility. I'm certain people have been using `add_bounds=False` to fix metadata issues and then calling `add_bounds()` or `add_missing_bounds()`.

Refer to this commit diff for more info: https://github.com/xCDAT/xcdat/pull/418/commits/c28b2277e6e46abeaa03241a9d1f0a73074201c8

pochedls: I view this as another improvement to the API I originally proposed. Thanks.

tomvothecoder:
Thank you @pochedls for your work and @lee1043 for validating.

I am merging this milestone PR!

PR #392, opened 2022-11-17
Fix `swap_lon_axis()` breaking when sorting with singleton coords

tomvothecoder:
## Description

- Closes #389 

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules

If applicable:

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass with my changes (locally and CI/CD build)
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have noted that this is a breaking change for a major release (fix or feature that would cause existing functionality to not work as expected)

Issue #389, opened 2022-11-11
[Bug]: Singletons breaks default `sort_ascending=True` in `swap_lon_axis()`


tomvothecoder:
### What happened?

> Using xcdat v0.4.0, I just got an error when trying to to swap the longitude axes. I wonder if the issue is that coord_keys = get_dim_coords(ds, "X").coords.keys() on [L236](https://github.com/xCDAT/xcdat/blob/main/xcdat/axis.py#L236) of swap_lon_axis() returns not just the coordinate key for the "X" dimension but also "height" since this is a 2-meter air temperature.

Hey @mzelinka and @pochedls, this issue is related to  `height` being a singleton coordinate. Singleton coordinate variables are  included alongside other coordinate variables (refer to code example below). There is an open xarray issue regarding singletons here: https://github.com/pydata/xarray/issues/6196

```python
import xcdat as xc

fnames = "/p/css03/esgf_publish/CMIP6/CMIP/BCC/BCC-CSM2-MR/historical/r1i1p1f1/day/tasmax/gn/v20181126/*.nc"

ds = xc.open_mfdataset(fnames)

print(xc.get_dim_coords(ds, axis="X"))

# Notice how `height` is included in the `lon` coordinate var.

# <xarray.DataArray 'lon' (lon: 320)>
# array([  0.   ,   1.125,   2.25 , ..., 356.625, 357.75 , 358.875])
# Coordinates:
#   * lon      (lon) float64 0.0 1.125 2.25 3.375 4.5 ... 355.5 356.6 357.8 358.9
#     height   float64 2.0
# Attributes:
#     bounds:         lon_bnds
#     units:          degrees_east
#     axis:           X
#     long_name:      Longitude
#     standard_name:  longitude
```

### How it affects `swap_lon_axis()`
In `swap_lon_axis()`, we are getting all of the coordinates keys using `lon`, which includes `height`.
This causes `swap_lon_axis(ds, sort_ascending=True)` to break on this file.
https://github.com/xCDAT/xcdat/blob/d5d4fd77cbe0e23cc6452c2a1d421923d800b31b/xcdat/axis.py#L235-L236
https://github.com/xCDAT/xcdat/blob/d5d4fd77cbe0e23cc6452c2a1d421923d800b31b/xcdat/axis.py#L259-L260

### Temporary workaround: 
* Set `sort_ascending=False` in `swap_lon_axis()` and perform sorting manually

[...]

PR #434, opened 2023-03-15
Accessor implementation of PR #418

tomvothecoder:
## Description

An extension of #418 

## Checklist

If applicable:

- [x] I have noted that this is a breaking change for a major release (fix or feature that would cause existing functionality to not work as expected) -- **technically, removing the `add_missing_bounds()` `axes` default value is a breaking change**


tests/test_bounds.py
```diff
@@ -440,120 +433,120 @@ def test_returns_bounds_for_dataset_with_time_coords_as_cftime_objects(self):
...
-        # sub hourly data is not supported
-        with pytest.raises(ValueError):
-            create_time_bounds(ds_subhrly_wo_bnds.time)
-
-        # ensure identical
-        assert hourly_bounds.identical(ds_hrly_with_bnds.time_bnds)
-        assert daily_bounds.identical(ds_daily_with_bnds.time_bnds)
-        assert monthly_bounds.identical(ds_monthly_with_bnds.time_bnds)
-        assert yearly_bounds.identical(ds_yearly_with_bnds.time_bnds)
+    # def test_create_monthly_bounds(self):
```
tomvothecoder, line 436: Update tests after confirming implementation direction.

xcdat/bounds.py
```diff
@@ -231,25 +256,41 @@ def get_bounds(
...
-        add bounds for each of them if they don't exist. The coordinates must
-        meet the following criteria in order to add bounds:
-
-          1. The axis for the coordinates are "X", "Y", "T", or "Z"
-          2. Coordinates are single dimensional, not multidimensional
-          3. Coordinates are a length > 1 (not singleton)
-          4. Bounds must not already exist.
-             * Determined by attempting to map the coordinate variable's
-             "bounds" attr (if set) to the bounds data variable of the same key.
-          5. Time axes should be composed of `cftime` objects.
```
tomvothecoder, line 247: In addition to `cftime` objects, we should also support time coordinates composed of `datetime` objects. I removed #5 from the criteria.

xcdat/bounds.py
```diff
@@ -374,26 +427,12 @@ def _create_bounds(self, axis: CFAxisKey, coord_var: xr.DataArray) -> xr.DataArr
         diffs = np.insert(diffs, 0, diffs[0])
         diffs = np.append(diffs, diffs[-1])
 
-        # `cftime` objects only support arithmetic using `timedelta` objects, so
-        # the values of  `diffs` must be casted from `dtype="timedelta64[ns]"`
-        # to `timedelta` objects.
-        if axis == "T" and issubclass(type(coord_var.values[0]), cftime.datetime):
-            bounds = create_time_bounds(coord_var)
-            return bounds
```
tomvothecoder, line 382: `if axis == "T"` is not necessary in `_create_bounds()`, and was moved to `add_bounds()`/`add_missing_bounds()`.

xcdat/bounds.py
```diff
@@ -374,26 +427,12 @@ def _create_bounds(self, axis: CFAxisKey, coord_var: xr.DataArray) -> xr.DataArr
...
         # width parameter: determines bounds location relative to midpoints
         width = 0.5
 
-        # FIXME: These lines produces the warning: `PerformanceWarning:
-        # Adding/subtracting object-dtype array to TimedeltaArray not
-        # vectorized` after converting diffs to `timedelta`. I (Tom) was not
-        # able to find an alternative, vectorized solution at the time of this
-        # implementation.
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore", category=pd.errors.PerformanceWarning)
```
tomvothecoder, line 393: These lines are no longer needed since we aren't do `timedelta` arithmetic in `_create_bounds()`.

xcdat/bounds.py
```diff
@@ -428,239 +467,238 @@ def _create_bounds(self, axis: CFAxisKey, coord_var: xr.DataArray) -> xr.DataArr
...
+        elif freq == "day":
+            time_bnds = self._create_daily_time_bounds(time)
+        elif freq == "hour":
+            # Determine the daily frequency for generating time  bounds.
+            if daily_freq is None:
+                diff = time.values[1] - time.values[0]
+                hrs = diff.seconds / 3600
+                daily_freq = int(24 / hrs)  # type: ignore
+
+            time_bnds = self._create_daily_time_bounds(time, freq=daily_freq)  # type: ignore
```
tomvothecoder, line 529: Do these functions support time coordinates composed of `datetime` objects? We should keep this support in this PR.

tomvothecoder (commented):
Hi @pochedls, here's a quick draft PR with our latest implementation decisions based on the meeting discussion. The PR description lists the changes. Thanks!

xcdat/bounds.py
tomvothecoder: Also we should be supporting time coordinates as `datetime` objects (removed `issubclass(...)` conditional)

xcdat/bounds.py
```diff
@@ -231,25 +256,41 @@ def get_bounds(
...
-          3. Coordinates are a length > 1 (not singleton)
-          4. Bounds must not already exist.
-             * Determined by attempting to map the coordinate variable's
-             "bounds" attr (if set) to the bounds data variable of the same key.
-          5. Time axes should be composed of `cftime` objects.
+        add bounds for each of them if they don't exist. If ``axis`` is "X",
+        "Y", or "Z", each coordinate point is the midpoint between their lower
+        and upper bounds. If the axis is "T", bounds will be set to the start
+        and end of each time step's period using either the inferred or specified
+        time frequency (``time_freq`` parameter).
```
tomvothecoder, line 272: TODO: This information should also be in `add_missing_bounds()`

xcdat/bounds.py
```diff
@@ -428,239 +467,238 @@ def _create_bounds(self, axis: CFAxisKey, coord_var: xr.DataArray) -> xr.DataArr
...
 
-        if axis not in cf_axis_keys:
-            keys = ", ".join(f"'{key}'" for key in cf_axis_keys)
-            raise ValueError(
-                f"Incorrect 'axis' argument value. Supported values include {keys}."
-            )
+        This function uses ``_infer_freq`` to determine the temporal resolution and
+        then calls ``get_yearly_time_bounds``, ``get_monthly_time_bounds``, or
+        ``get_daily_time_bounds`` as appropriate. See the docstrings for these
+        functions for more details.
```
tomvothecoder, line 481: FIXME: These method names have changed from `get_...` to `_create...`.

tomvothecoder, line 481:
```suggestion
```

tomvothecoder: I don't think we need this documentation anymore since `_create_time_bounds()` is a private method and the "Parameters" docstrings include this info.

xcdat/bounds.py
```diff
@@ -428,239 +467,238 @@ def _create_bounds(self, axis: CFAxisKey, coord_var: xr.DataArray) -> xr.DataArr
...
+        List[cftime.datetime]
+            A list of time bound values.
+        """
+        # Get the calendar type from the time coordinates to determine the `cftime`
+        # object to represent bounds values.
+        calendar = time.encoding["calendar"]
+        cf_obj = get_date_type(calendar)
 
-        time_bnds.append([l_bnd, u_bnd])
+        time_bnds: List[cftime.datetime] = []
```
tomvothecoder, line 563: FIXME: We should detect if the input time coordinates are `datetime` or `cftime` and keep the time bounds the same object type.

xcdat/bounds.py
tomvothecoder: I added logic in the latest commit to support datetime objects.

xcdat/bounds.py
tomvothecoder: I added logic in the latest commit to support datetime objects.

xcdat/bounds.py
tomvothecoder: This is fixed in the latest commit.

xcdat/bounds.py
```diff
@@ -428,239 +468,319 @@ def _create_bounds(self, axis: CFAxisKey, coord_var: xr.DataArray) -> xr.DataArr
...
+
+        # If day is greater than days in month use days in month as day.
+        day = timestep.day
+        dim = obj_type(year, month, 1).daysinmonth
+        day = min(day, dim)
+
+        # Create the new timestep.
+        new_timestep = obj_type(year, month, day, hour, minute, second, microsecond)
+
+        return new_timestep
```
tomvothecoder, line 704: `_add_months_to_timestep` method is based on `_month_add()` and replaces `_month_add()`. 

I added this method to `BoundsAccessor` because it is a private method that is only called in `_create_monthly_time_bounds()`. I removed logic related to arrays because we only pass a single timestep (`cftime.datetime` or `pd.Timestamp`) internally.

tests/test_bounds.py
```diff
@@ -544,16 +466,53 @@ def test_generic_create_time_bounds_function(self):
...
+    #     # expect time bounds minus one month
+    #     expected_time_bnds = ds_with_bnds.time_bnds
+    #     lower = _month_add(expected_time_bnds[:, 0], -1, "standard")
+    #     upper = _month_add(expected_time_bnds[:, 1], -1, "standard")
+    #     expected_time_bnds[:, 0] = lower
+    #     expected_time_bnds[:, 1] = upper
+    #     # test bounds generation
+    #     result = create_monthly_time_bounds(ds.time, end_of_month=True)
+
+    #     assert result.identical(expected_time_bnds)
```
tomvothecoder, line 518: TODO: Add internal logic for `end_of_month` and fix this test

tomvothecoder (commented):
Second pass refactor

xcdat/bounds.py
```diff
@@ -260,12 +275,6 @@ def add_bounds(self, axis: CFAxisKey) -> xr.Dataset:
         ------
         ValueError
             If bounds already exist. They must be dropped first.
```
tomvothecoder, line 277:
```suggestion
```

This error is no longer thrown, the coordinates are just ignored if they already have bounds.

xcdat/bounds.py
```diff
@@ -290,6 +299,101 @@ def add_bounds(self, axis: CFAxisKey) -> xr.Dataset:
...
+
+        Returns
+        -------
+        xr.Dataset
+            The dataset with bounds added.
+
+        Raises
+        ------
+        ValueError
+            If bounds already exist. They must be dropped first.
```
tomvothecoder, line 371:
```suggestion
```

This error is no longer thrown, the coordinates are just ignored if they already have bounds.

xcdat/bounds.py
```diff
@@ -290,6 +302,101 @@ def add_bounds(self, axis: CFAxisKey) -> xr.Dataset:
...
+            ``freq`` or ``midpoint``.
+
+            * ``freq``: Create time bounds as the start and end of each
+                timestep's period using either the inferred or specified time
+                frequency (``freq`` parameter). For example, the time bounds
+                will be the start and end of each month for each monthly
+                coordinate point.
+            * ``midpoint``: Create time bounds using time coordinates as the
+                midpoint between their upper and lower bounds.
+
```
tomvothecoder, line 339:
```suggestion
            The method for creating time bounds for time coordinates, either
            ``"freq"`` or ``"midpoint"``.

            * ``"freq"``: Create time bounds as the start and end of each
                timestep's period using either the inferred or specified time
                frequency (``freq`` parameter). For example, the time bounds
                will be the start and end of each month for each monthly
                coordinate point.
            * ``"midpoint"``: Create time bounds using time coordinates as the
                midpoint between their upper and lower bounds.
                
```

xcdat/bounds.py
```diff
@@ -290,6 +302,101 @@ def add_bounds(self, axis: CFAxisKey) -> xr.Dataset:
...
+            The method for creating time bounds for time coordinates, either
+            ``freq`` or ``midpoint``.
+
+            * ``freq``: Create time bounds as the start and end of each
+                timestep's period using either the inferred or specified time
+                frequency (``freq`` parameter). For example, the time bounds
+                will be the start and end of each month for each monthly
+                coordinate point.
+            * ``midpoint``: Create time bounds using time coordinates as the
+                midpoint between their upper and lower bounds.
```
tomvothecoder, line 338:
```suggestion
        method : Literal["freq", "midpoint"]
            The method for creating time bounds for time coordinates, either
            "freq" or "midpoint".

            * "freq": Create time bounds as the start and end of each
                timestep's period using either the inferred or specified time
                frequency (``freq`` parameter). For example, the time bounds
                will be the start and end of each month for each monthly
                coordinate point.
            * "midpoint": Create time bounds using time coordinates as the
                midpoint between their upper and lower bounds.
```

xcdat/bounds.py
```diff
@@ -232,40 +248,33 @@ def get_bounds(
...
              * Determined by attempting to map the coordinate variable's
-             "bounds" attr (if set) to the bounds data variable of the same key.
-          5. Time axes should be composed of `cftime` objects.
+             "bounds" attr (if set) to the bounds data variable of the same key
 
         Parameters
         ----------
         axis : CFAxisKey
-            The CF axis key ("X", "Y", "T", or "Z").
+            The CF axis key ("X", "Y", "Z").
```
tomvothecoder, line 270:
```suggestion
            The CF axis key ("X", "Y", "T", "Z").
```

tomvothecoder (commented):
Hi @pochedls, this PR is now finally and officially ready for review. I promise I won't make any more changes while you review. 

For each file under the ["Files Changed"](https://github.com/xCDAT/xcdat/pull/434/files) tab, I listed descriptions for their changes. Let me know how it looks. Thanks!

**docs/api.rst**
- Removed `create_time_bounds()` and `create_XYZ_time_bounds()` from list of APIs
- Add `contains_datetime_like_objects()` and `get_datetime_like_type()` utility functions which I thought were useful to make public
- Add `Dataset.bounds.add_time_bounds()` accessor method

**xcdat/\_\_init\_\_.py**
- Add top-level imports for `contains_datetime_like_objects()` and `get_datetime_like_type()` utility functions 

**xcdat/bounds.py**
- Update `add_missing_bounds()` docstrings and modify `axes` to not accept a default value (`axes = ["X", "Y"]`) -- _xCDAT open functions will set this default value, while the user should explicitly pass `axes` if they call this function directly_
- Update `add_missing_bounds()` to call `self._create_time_bounds()`, which generates bounds based on frequency (not generic midpoint)
- Update `add_bounds()` docstring stating that this method uses coordinates as midpoints for bounds
  - - Update `_create_bounds()` docstring with info on using coordinates as midpoints
- Add new method `add_time_bounds()` which adds bounds using the frequency and time steps of time coordinates -- _splitting this method separately makes the most sense to me and is the cleanest API solution IMO_
  - Add `_create_time_bounds()` -- private method that calls `_create_XYZ_time_bounds()` methods
  - Replace `xcdat/temporal.py/_month_add()` function with `_add_months_to_timestep()` accessor method -- _this method is only called internally and only needs to operate on a single timestep rather than an array/vector of timesteps_

**test/test_bounds.py**
- Replace deprecated tests for `create_XYZ_time_bounds()` functions with tests for bounds accessor method `create_time_bounds()`
 
**xcdat/dataset.py**
- Update `open_dataset()` and `open_mfdataset()` docstring for `bounds_axes` regarding how bounds are generated by default

**xcdat/regridder/*.py**
- Update all regridder references to `add_missing_bounds()` to explicitly pass `axes=["X", "Y", "T"]` 

**xcdat/temporal.py**
- Add `contains_datetime_like_objects()` function
- Add `get_datetime_like_type()` function
- Update `_set_data_var_attrs()` to use `contains_datetime_like_objects()` 
- Remove `_month_add()` (replaced by bounds accessor method `_add_months_to_timestep()`)

**test/temporal.py**
- Add tests for `contains_datetime_like_objects()` and `get_datetimelike_type()`
- Remove tests for deprecated `_month_add()` function

xcdat/regridder/xesmf.py
```diff
@@ -200,6 +200,6 @@ def horizontal(self, data_var: str, ds: xr.Dataset) -> xr.Dataset:
         # preserve non-spatial bounds
         output_ds = preserve_bounds(ds, self._output_grid, output_ds)
 
-        output_ds = output_ds.bounds.add_missing_bounds()
+        output_ds = output_ds.bounds.add_missing_bounds(axes=["X", "Y", "T"])
```
pochedls, line 203:
```suggestion
        output_ds = output_ds.bounds.add_missing_bounds(axes=["X", "Y"])
```

Do we need to include `T` here? We don't include it by default in `open_dataset`. The bounds also don't appear to get dropped during horizontal regridding (so if they are in the original dataset, they are preserved during regridding).

xcdat/regridder/accessor.py
```diff
@@ -78,7 +78,7 @@ def grid(self) -> xr.Dataset:
 
         ds = xr.Dataset(coords, attrs=self._ds.attrs)
 
-        ds = ds.bounds.add_missing_bounds()
+        ds = ds.bounds.add_missing_bounds(axes=["X", "Y", "T"])
```
pochedls, line 81:
```suggestion
        ds = ds.bounds.add_missing_bounds(axes=["X", "Y"])
```

xcdat/regridder/grid.py
```diff
@@ -508,7 +508,7 @@ def create_grid(
 
     grid = xr.Dataset(data_vars=data_vars, coords={"lat": lat, "lon": lon})
 
-    grid = grid.bounds.add_missing_bounds()
+    grid = grid.bounds.add_missing_bounds(axes=["X", "Y", "T"])
```
pochedls, line 511:
```suggestion
    grid = grid.bounds.add_missing_bounds(axes=["X", "Y"])
```

xcdat/regridder/regrid2.py
```diff
@@ -150,7 +150,7 @@ def horizontal(self, data_var: str, ds: xr.Dataset) -> xr.Dataset:
         # preserve non-spatial bounds
         output_ds = preserve_bounds(ds, self._output_grid, output_ds)
 
-        output_ds = output_ds.bounds.add_missing_bounds()
+        output_ds = output_ds.bounds.add_missing_bounds(axes=["X", "Y", "T"])
```
pochedls, line 153:
```suggestion
        output_ds = output_ds.bounds.add_missing_bounds(axes=["X", "Y"])
```

xcdat/bounds.py
```diff
@@ -131,12 +136,16 @@ def add_missing_bounds(self, axes: List[CFAxisKey] = ["X", "Y"]) -> xr.Dataset:
           4. Bounds must not already exist.
              * Determined by attempting to map the coordinate variable's
              "bounds" attr (if set) to the bounds data variable of the same key.
-          5. Time axes should be composed of `cftime` objects.
+          5. Time axes must be composed of datetime-like objects (`np.datetime`
```
pochedls, line 139:
```suggestion
          5. Time axes must be composed of datetime-like objects (`np.datetime64`
```

xcdat/bounds.py
```diff
@@ -232,40 +248,33 @@ def get_bounds(
...
+        criteria:
 
           1. The axis for the coordinates are "X", "Y", "T", or "Z"
           2. Coordinates are single dimensional, not multidimensional
           3. Coordinates are a length > 1 (not singleton)
-          4. Bounds must not already exist.
+          4. Bounds must not already exist
              * Determined by attempting to map the coordinate variable's
-             "bounds" attr (if set) to the bounds data variable of the same key.
-          5. Time axes should be composed of `cftime` objects.
```
pochedls, line 247: I'm curious why this was removed – it seems like you specifically require this in `_create_bounds`.

xcdat/bounds.py
```diff
@@ -374,15 +477,18 @@ def _create_bounds(self, axis: CFAxisKey, coord_var: xr.DataArray) -> xr.DataArr
...
+                raise TypeError(
+                    f"Bounds cannot be created for '{coord_var.name}' coordinates "
+                    "because it is not decoded as `cftime.datetime` or `np.datetime`. "
+                    f"Try decoding '{coord_var.name}' first then adding bounds."
+                )
+            # `cftime` objects only support arithmetic using `timedelta` objects, so
+            # the values of  `diffs` must be casted from `dtype="timedelta64[ns]"`
+            # to `timedelta` objects.
+            if contains_cftime_datetimes(xr.as_variable(coord_var)):
+                diffs = pd.to_timedelta(diffs)
```
pochedls, line 491: I wonder if we should require this for time axes in the generic `create_bounds` function. If the user opened the dataset without decoding then the time values would be floats/ints (e.g., hours since... with values like [0, 30, 60, ...]) and this function could come up with reasonable (un-decoded) bounds. I think this requirement could be eliminated.

xcdat/bounds.py
```diff
@@ -428,239 +535,341 @@ def _create_bounds(self, axis: CFAxisKey, coord_var: xr.DataArray) -> xr.DataArr
...
+        The delta value can be positive or negative (for subtraction). Refer to [1]_
+        for logic.
 
-def create_daily_time_bounds(
-    time: xr.DataArray, freq: Literal[1, 2, 3, 4, 6, 8, 12, 24] = 1
-) -> xr.DataArray:
-    """Creates time bounds for each timestep with the start and end of the day.
+        Parameters
+        ----------
+        timesep : Union[cftime.datime, pd.Timestamp]
```
pochedls, line 758:
```suggestion
        timestep : Union[cftime.datime, pd.Timestamp]
```

xcdat/__init__.py
```diff
@@ -22,7 +16,11 @@
     create_zonal_grid,
 )
 from xcdat.spatial import SpatialAccessor  # noqa: F401
-from xcdat.temporal import TemporalAccessor  # noqa: F401
+from xcdat.temporal import (  # noqa: F401
+    TemporalAccessor,
+    contains_cftime_datetimes,
+    get_datetime_like_type,
```
pochedls, line 22: I don't mind making these so visible, but I'm curious why the end-user would need/want these? I see them mostly used behind-the-scenes.

pochedls (commented):
Looks great! This organization is way better (compared to my earlier PR). I think this accounts for all the comments that were made on the first version of this PR. 

I made a few comments / suggestions to consider (one would require modifying a few unit tests), but I think this is okay to merge once you've looked over the comments.

xcdat/bounds.py
pochedls: With that said...I don't think we need to require datetime-like objects here (see next comment below).

xcdat/bounds.py
tomvothecoder: Thanks for catching this. I accidentally created a regression from the `main` code because we supported non-decoded time coordinates before. I'll delete this line.

xcdat/regridder/xesmf.py
tomvothecoder: I included `T` here because the old `add_missing_bounds()` added bounds for X, Y, T, and Z (if these axes are present in the dataset). I included all of these axes (except `Z` because I forgot to here) to make sure `output_ds` would be the same.

tomvothecoder (commented):
Minor fixes

xcdat/__init__.py
tomvothecoder: I initially implemented them as private methods, but I thought it would be useful to have them available for users who wanted to implement some more advanced functionality (which is also why I think Xarray made their version of [`contains_cftime_datetime()`](https://github.com/pydata/xarray/blob/c3224ef599e701efb00bbf4c20f8cf5e247a021d/xarray/core/common.py#L1816-L1818)) public.

I can see an argument for keeping them private because most users of xCDAT are probably focused on the core functionality of temporal averaging, regridding, etc. Unless there is a demand for this kind of feature (there probably won't be), I will revert them back to being private.

xcdat/bounds.py
tomvothecoder: Good catch, we previously supported numerically encoded time coordinates so this conditional creates a new regression. I'll remove this conditional.

xcdat/bounds.py
```diff
@@ -290,6 +299,97 @@ def add_bounds(self, axis: CFAxisKey) -> xr.Dataset:
...
+            except KeyError:
+                if method == "freq":
+                    bounds = self._create_time_bounds(
+                        coord, freq, daily_subfreq, end_of_month
+                    )
+                elif method == "midpoint":
+                    bounds = self._create_bounds("T", coord)
+
+                ds[bounds.name] = bounds
+                ds[coord.name].attrs["bounds"] = bounds.name
```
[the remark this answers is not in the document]
tomvothecoder: `add_missing_bounds()` has these try and except statements because we want to make sure datasets can be opened with `open_dataset()` and `open_mfdataset()`.

`add_bounds()` and `add_time_bounds()` are standalone public APIs that are called on datasets that were successfully opened. I lean towards keeping these try and except statements, otherwise those errors will be silent. The user has more control over how to use these APIs and can add a try and except statement in their own code.

```python
ds = xc.open_mfdataset(...)

try:
  ds = ds.add_bounds(axis="X")
except (ValueError, TypeError):
  continue

try:
  ds = ds.add_time_bounds()
except (ValueError, TypeError):
  continue
```

xcdat/regridder/xesmf.py
tomvothecoder: `add_missing_bounds()` originally added bounds for all axes including "T" and "Z".
To avoid regression, I decided to include "T" (and I forgot to include "Z") in the new `axes` argument.

I'm not sure if it is absolutely required to add "T" bounds during horizontal regridding. @jasonb5 any thoughts?

pochedls: I don't think `T` (or `Z`) should be added here. It would kind of be a surprise to not have time bounds on open and then have them added during regridding.

xcdat/bounds.py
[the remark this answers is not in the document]
pochedls: Is `height` a time axis? It seems like the user shouldn't get a ValueError trying to add bounds for `time` when there is a non-dimensional height coordinate embedded in the data.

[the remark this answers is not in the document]
tomvothecoder: SIngleton coordinate variables are included alongside other coordinate variables. This issue is related to https://github.com/xCDAT/xcdat/issues/389.

I'll investigate a solution similar to https://github.com/xCDAT/xcdat/pull/392/files.

xcdat/regridder/xesmf.py
tomvothecoder: Gotcha, makes sense to me. If I recall correctly, there was a discussion somewhere with Karl where somebody mentions that trying to add "Z" bounds using the generic midpoint algorithm can produce incorrect results.

tomvothecoder:
Reviewed all comments and implemented most suggestions. Merging this PR into #418 for final review.

PR #460, opened 2023-04-25
xml open update

### LINKED ISSUE #454: [Bug]: xml open error
### What happened?

I was opening a xml file that is indexed for multiple netcdf files. Opening as `open_mfdataset('*.nc')` worked but `open_mfdataset('test.xml')` returned error. 

Input data for the below minimal example code: [input.zip](https://github.com/xCDAT/xcdat/files/11315375/input.zip)

I expected `open_mfdataset('*.nc')` and `open_mfdataset('test.xml')` result the same dataset.

```python
import xcdat as xc
ds = xc.open_mfdataset('precip.V1.0.1948.nc')
ds = xc.open_mfdataset('test_1948.xml')
```

### Relevant log output

Log from `ds = xc.open_mfdataset('precip.V1.0.1948.nc')`
```
2023-04-24 16:52:26,768 [WARNING]: dataset.py(decode_time:321) >> 'time' does not have a calendar attribute set. Defaulting to CF 'standard' calendar.
```

Log from `ds = xc.open_mfdataset('test_1948.xml')`
```
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
Cell In[4], line 1
----> 1 ds = xc.open_mfdataset('test_1948.xml')

File ~/mambaforge/envs/xcdat_devel_20230419/lib/python3.10/site-packages/xcdat/dataset.py:233, in open_mfdataset(paths, data_var, add_bounds, bounds_axes, decode_times, center_times, lon_orient, data_vars, preprocess, **kwargs)
    229         paths = _parse_dir_for_nc_glob(paths)  # type: ignore
    231 preprocess = partial(_preprocess, decode_times=decode_times, callable=preprocess)
--> 233 ds = xr.open_mfdataset(
    234     paths,
    235     decode_times=False,
    236     data_vars=data_vars,
    237     preprocess=preprocess,
    238     **kwargs,  # type: ignore
    239 )
    241 ds = _postprocess_dataset(
    242     ds, data_var, center_times, add_bounds, bounds_axes, lon_orient
    243 )
    245 return ds

File ~/mambaforge/envs/xcdat_devel_20230419/lib/python3.10/site-packages/xarray/backends/api.py:948, in open_mfdataset(paths, chunks, concat_dim, compat, preprocess, engine, data_vars, coords, combine, parallel, join, attrs_file, combine_attrs, **kwargs)
    945     paths = [os.fspath(p) if isinstance(p, os.PathLike) else p for p in paths]
    947 if not paths:
--> 948     raise OSError("no files to open")
    950 if combine == "nested":
    951     if isinstance(concat_dim, (str, DataArray)) or concat_dim is None:

OSError: no files to open
```

_No response_

### Environment

xcdat 0.5.0

### LINKED ISSUE #459: [Doc]: update open_mfdataset documentation to include information on cdml/xml

From https://github.com/xCDAT/xcdat/issues/454, we realized that xcdat functionality to open cdml files is not included in the docstring for `open_mfdataset`. This should be done.

lee1043:
## Description
* Open only specific nc file(s) that are listed in the xml file, instead of opening all nc files in the directory
* When xml was generated using relative path, in particular in the same directory where data is, assume nc files are in the current directory (solution suggested by @pochedls) 

- Closes #454 
- Closes #459

## Checklist

If applicable:

### DIFF
--- a/xcdat/dataset.py
+++ b/xcdat/dataset.py
@@ -414,8 +414,8 @@ def _parse_xml_for_nc_glob(xml_path: Union[str, pathlib.Path]) -> Union[str, lis
 
     Returns
     -------
-    str
-A list of files (if specified with a cdms_filemap attribute) or a string glob of `*.nc` paths. 
+    list | str
+        A list of files (if specified with a cdms_filemap attribute) or a string glob of `*.nc` paths.
 
     Notes
     -----
--- a/xcdat/dataset.py
+++ b/xcdat/dataset.py
@@ -434,14 +434,24 @@ def _parse_xml_for_nc_glob(xml_path: Union[str, pathlib.Path]) -> str:
     root = tree.getroot()
 
     dir_attr = root.attrib.get("directory")
-    if dir_attr is None:
+    if dir_attr == "":
+        dir_attr = "."
+    elif dir_attr is None:
         raise KeyError(
             f"The XML file ({xml_path}) does not have a 'directory' attribute "
             "that points to a directory of `.nc` dataset files."
         )
 
-    glob_path = dir_attr + "/*.nc"
+    filemap_attr = root.attrib.get("cdms_filemap")
+    if filemap_attr is not None:
+        file_name_list = sorted(list(set([x for x in filemap_attr.replace('[', ' ').replace(']', ' ').replace(',', ' ').split(' ') if '.nc' in x])))
+        glob_path = list()
+        for file_name in file_name_list:
+            glob_path.append(os.path.join(dir_attr, file_name))
+    else:
+        glob_path = dir_attr + "/*.nc"
 
+    print('glob_path:', glob_path)
     return glob_path
 
 
--- a/xcdat/dataset.py
+++ b/xcdat/dataset.py
@@ -450,13 +450,15 @@ A list of files (if specified with a cdms_filemap attribute) or a string glob of
         fm_values = fma.split(" ")
         # keep values that are nc files (and remove white space)
         file_name_list = [fn.replace(" ", "") for fn in fm_values if ".nc" in fn]
+        # remove duplicates and sort by alphabetical order
+        file_name_list = sorted(list(set(file_name_list)))
         # create empty list
         glob_path = list()
         # combine directory and file names
         for file_name in file_name_list:
             glob_path.append(os.path.join(dir_attr, file_name))
     else:
-        glob_path = dir_attr + "/*.nc"
+        glob_path = os.path.join(dir_attr, "*.nc")
 
     return glob_path
 
--- a/xcdat/dataset.py
+++ b/xcdat/dataset.py
@@ -458,7 +458,6 @@ def _parse_xml_for_nc_glob(xml_path: Union[str, pathlib.Path]) -> str:
     else:
         glob_path = dir_attr + "/*.nc"
 
-    print("glob_path:", glob_path)
     return glob_path
 
 
--- a/xcdat/dataset.py
+++ b/xcdat/dataset.py
@@ -1,8 +1,11 @@
 """Dataset module for functions related to an xarray.Dataset."""
+from __future__ import annotations
+
 import os
 import pathlib
 from datetime import datetime
 from functools import partial
+from io import BufferedIOBase
 from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Union
 
 import numpy as np
@@ -10,9 +13,11 @@ import xarray as xr
 from dateutil import parser
 from dateutil import relativedelta as rd
 from lxml import etree
+from xarray.backends.common import AbstractDataStore
 from xarray.coding.cftime_offsets import get_date_type
 from xarray.coding.times import convert_times, decode_cf_datetime
 from xarray.coding.variables import lazy_elemwise_func, pop_to, unpack_for_decoding
+from xarray.core.types import NestedSequence
 from xarray.core.variable import as_variable
 
 from xcdat import bounds as bounds_accessor  # noqa: F401
@@ -38,9 +43,9 @@ Paths = Union[
 
 
 def open_dataset(
-    path: str,
+    filename_or_obj: str | os.PathLike[Any] | BufferedIOBase | AbstractDataStore,
     data_var: Optional[str] = None,
-    add_bounds: Union[List[CFAxisKey], None, Literal[False]] = ["X", "Y"],
+    add_bounds: List[CFAxisKey] | None | Literal[False] = ["X", "Y"],
     decode_times: bool = True,
     center_times: bool = False,
     lon_orient: Optional[Tuple[float, float]] = None,
@@ -50,12 +55,16 @@ def open_dataset(
 
     Parameters
     ----------
-    path : str
-        Path to Dataset.
+    filename_or_obj : str, Path, file-like or DataStore
+        Strings and Path objects are interpreted as a path to a netCDF file
+        or an OpenDAP URL and opened with python-netCDF4, unless the filename
+        ends with .gz, in which case the file is gunzipped and opened with
+        scipy.io.netcdf (only netCDF3 supported). Byte-strings or file-like
+        objects are opened by scipy.io.netcdf (netCDF3) or h5py (netCDF4/HDF).
     data_var: Optional[str], optional
         The key of the non-bounds data variable to keep in the Dataset,
         alongside any existing bounds data variables, by default None.
-    add_bounds: Union[List[CFAxisKey], None, Literal[False]]
+    add_bounds: List[CFAxisKey] | None | Literal[False]
         List of CF axes to try to add bounds for (if missing), default
         ["X", "Y"]. Set to ``None`` or ``False`` to not try to add any missing
         bounds.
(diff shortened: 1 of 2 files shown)


xcdat/dataset.py
```diff
@@ -434,14 +434,37 @@ def _parse_xml_for_nc_glob(xml_path: Union[str, pathlib.Path]) -> str:
...
+                        .replace(",", " ")
+                        .split(" ")
+                        if ".nc" in x
+                    ]
+                )
+            )
+        )
+        glob_path = list()
+        for file_name in file_name_list:
+            glob_path.append(os.path.join(dir_attr, file_name))
```
pochedls, line 463: Tried to make this more succint, readable, and added comments. 

```suggestion
        # remove brackets and commas from filemap_attr
        fma = filemap_attr.replace("[", " ").replace("]", " ").replace(",", " ")
        # split string filemap_attr up (by spaces)
        fm_values = fma.split(" ")
        # keep values that are nc files (and remove white space)
        file_name_list = [fn.replace(" ", "") for fn in fm_values if ".nc" in fn]
        # create empty list
        glob_path = list()
        # combine directory and file names
        for file_name in file_name_list:
            glob_path.append(os.path.join(dir_attr, file_name))
```

xcdat/dataset.py
```diff
@@ -434,14 +434,37 @@ def _parse_xml_for_nc_glob(xml_path: Union[str, pathlib.Path]) -> str:
...
+    else:
+        glob_path = dir_attr + "/*.nc"
 
+    print("glob_path:", glob_path)
```
pochedls, line 467: We don't want to print this do we?

pochedls (commented):
@lee1043 – this looks good to me - thanks. I suggested some minor changes (@tomvothecoder might prefer your syntax though). Regardless – comments would be helpful. 

Do you have the dev environment setup? If so, could you deal with the pre-commit checks that seem to be failing (mypy / codecov)? 

While you have this open – could you handle this issue, too: https://github.com/xCDAT/xcdat/issues/459?

xcdat/dataset.py
lee1043: No, thanks for pointing out. That was just for test which I missed to remove.

xcdat/dataset.py
lee1043: @pochedls this is great! Thank you for the suggestion!

xcdat/dataset.py
```diff
@@ -415,7 +415,7 @@ def _parse_xml_for_nc_glob(xml_path: Union[str, pathlib.Path]) -> str:
     Returns
     -------
     str
```
pochedls, line 417: I'm not sure how this is denoted, but I think it should be `str | List`

xcdat/dataset.py
```diff
@@ -415,7 +415,7 @@ def _parse_xml_for_nc_glob(xml_path: Union[str, pathlib.Path]) -> str:
...
-        A glob of `*.nc` paths in the directory.
+        A glob of `*.nc` paths in the directory or list of nc files included in the CDML XML file.
```
pochedls, line 418: A list of files (if specified in a cdms_filemap attribute) or a string glob of `*.nc` paths.

pochedls (commented):
A couple more minor comments. Also, I wonder if we could update the `paths` parameter description under `open_mfdataset` to something like: 

>       Either a string glob in the form ``"path/to/my/files/*.nc"``, a (string) path
>       to a cdml/xml file that includes a directory attribute, or an explicit list of 
>       files to open. Paths can be given as strings or as pathlib Paths. If concatenation along                
>       more than one dimension is desired, then ``paths`` must be a nested list-of-lists (see         
>       ``combine_nested`` for details). (A string glob will be expanded to a 1-dimensional list.)

xcdat/dataset.py
lee1043: Thanks for the checking! Yes, that is correct. In the last [commit](https://github.com/xCDAT/xcdat/pull/460/commits/37d6fbab67f3057f743cdacffd1a1e56166be96a#diff-e5b713e4bf50c61361496f77fdadcf9d3ffc2fcb39c4c953932cd780ae951e55R402) I revised this to `def _parse_xml_for_nc_glob(xml_path: Union[str, pathlib.Path]) -> Union[str, list]:` (`str | List` syntax works for python >= 3.10, but fails for python <= 3.9)

tomvothecoder:
> When xml was generated using relative path, in particular in the same directory where data is, assume nc files are in the current directory (solution suggested by @pochedls)

After trying to write a unit test to cover this case, I found that substituting `directory=""` with the current working directory (`"."`) will introduce issues. 

1. The current directory (".") is relative to where `xcdat` is ran
   - If the user doesn't run `xcdat` in the same directory as the data, this will raise `OSError: no files to open`
   - I am confident there is a high possibility that users will fail to do this
2. This substitution behavior seems to be handling very specific I/O use cases, especially with the deprecated CDML format
   - I don't think this logic should be implemented in `xcdat` if it is only applicable to a small subset of potential `xcdat` users (who are CDML users)
   - We should eventually be moving away from CDML since CDAT is being deprecated

My proposed solution:
- CDML users should supply the string path explicitly if no `directory` attribute is found in the XML 
  - I think they should know where the `.nc` files are generated?
  - Alternatively, CDML users can implement `directory` attribute substitution logic in their own code instead of baking it into xCDAT

tomvothecoder:
> When xml was generated using relative path, in particular in the same directory where data is, assume nc files are in the current directory (solution suggested by @pochedls)

From my understanding, we are trying assuming `.nc` files are in the same directory as the `.xml` file. After trying to write a unit test to cover this case, I found that substituting `directory=""` with `"."` introduces issues:

1. The current directory (".") is relative to where `xcdat` is imported/ran, not the supplied XML file
   - If the user doesn't run `xcdat` in the same directory as the data, this will raise `OSError: no files to open`
2. This substitution behavior is handling specific I/O cases for CDML users


My proposed solution: CDML users should supply the string path explicitly if no `directory` attribute is found in the XML

Example:
```python
xml_path =  "/path/to/data/xml"

# Breaks because to `directory` key
xc.open_mfdataset(path)

# User assumes or knows files are in the same directory as the XML
xc.open_mfdataset("/path/to/data/*.nc")
```

lee1043:
@tomvothecoder I like the idea, but I am not clear where I can include this logic under the current structure. Would you mind adding it into the code in this PR?

tomvothecoder:
>My proposed solution: CDML users should supply the string path explicitly if no directory attribute is found in the XML
> @tomvothecoder I like the idea, but I am not clear where I can include this logic under the current structure. Would you mind adding it into the code in this PR?

What I mean by my proposed solution is that the **user implement the `try and except` statement in their code** if they are using XML files to open up datasets. This would catch instances where the `directory` key is potentially blank in the XML file(s). 

`xcdat` shouldn't assume a specific directory (e.g., current directory or same directory as XML) if this key is blank since it might not always be correct.

tomvothecoder:
> > Didn't this PR arise in part to address a blank directory attribute in the xml?
> > Can't we just add something like this [here](https://github.com/xCDAT/xcdat/pull/460/files#diff-e5b713e4bf50c61361496f77fdadcf9d3ffc2fcb39c4c953932cd780ae951e55R463):
> > `os.path.join(os.path.dirname(xml_path), '')` # wrap dirname with join to ensure you have the trailing slash
> > If the xml is missing the directory attribute it means that it was created with data from the same directory.
> 
> I think for this particular PR, I'm fine with adding the `cdms_filemap` parsing and assuming datasets are in the same directory as the XML file if it has `directory=""` (although we get the same OSError if the files are not in there).
> 
> In hindsight, I think CDML/XML support is more ideal for a PCMDI utilities package. We're starting to integrate features that are too specific to an end-of-life software (CDAT/CDML) and for a smaller user group. I'm being cautious about how much support we have for CDML in xCDAT because it is much more difficult to remove code than to add code, especially in a mature package.

Actually on second-thought, I still have concerns about this scope of this feature. Let's talk about it at our next meeting since isn't the highest priority item.

tomvothecoder:
Path forward:

- Scope this feature 
- Figure out timeline 
  - deprecate this feature by end of the year when CDAT/CDML gets deprecated?
- Review this PR

tomvothecoder:
We have feature request guidelines here that we overlooked https://xcdat.readthedocs.io/en/latest/contributing.html#feature-request

pochedls:
I have noted in the past that you could create your own xml/cdml file without cdscan, which could be a nice feature (e.g., handling spliced datasets), but it [is true that](https://cdms.readthedocs.io/en/latest/manual/cdms_6.html)

> Climate Data Markup Language (CDML) is the markup language used to represent metadata in CDMS

Does this mark the beginnings of the xcdml format? We will see.

tomvothecoder:
Much to the dismay of @lee1043 and @pochedls  I have decided to veto CDML support due to this [reason](https://github.com/xCDAT/xcdat/pull/460#issuecomment-1581290603).

Next steps will be:
1.  Close this PR
2.  Add a deprecation warning when opening CDML/XML files with xcdat in v0.6.0 -- this is standard practice and the user can silence them
3. Remove CDML support in the subsequent release (maybe v0.7.0) -- this timeline should parallel CDAT's end-of-life at the end of this year

### Suggested Alternative
I suggest we drive #492 forward, which discusses the idea of developing a package with more specialized features that meet the needs of LLNL Climate Program staff.

This new package can provide an API to parse CDML files for `.nc` filepaths (PR #460 and #385). I'm aware that it will take time for some users and packages (e.g.,  pcmdi_metrics) that rely on the CDML files hosted at LLNL to transition away from them. I don't think a path forward has been scoped for CDML users yet (maybe use [xsearch](https://github.com/pochedls/xsearch)?).

PR #495, opened 2023-06-05
Fix multi-file dataset spatial average orientation and weights when lon bounds span prime meridian 

### LINKED ISSUE #494: [Bug]:`.spatial.average(lat_bnds=(a,b)` does not appear to be working properly
### What happened?

I am a new user to XCDAT so this could definitely be user error.  I have a data set that I have am loading on `perlmutter` with this code
```
# Load SCREAMv1 regridded files with the Surface related output
directory_prefix='/global/cfs/cdirs/e3sm/terai/SCREAM/v1_production'
f_Jan_SurfVars=xcdat.open_mfdataset(directory_prefix+'/ne1024pg2_ne1024pg2_DY2/output.scream.SurfVars.INSTANT.nmins_x15.20*nc', add_bounds=True, parallel=True)
```
and then I attempt to take a spatial average with latitude bounds.  As a sanity check I took the average for wildly different areas of the globe expecting them to be different, but they end up matching exactly.  
I noticed this when plotting the average for more realistic cases and noticing the plots matched.

I expected the two different lat bounds to produce different answers.

```python
# Load SCREAMv1 regridded files with the Surface related output
directory_prefix='/global/cfs/cdirs/e3sm/terai/SCREAM/v1_production'
f_Jan_SurfVars=xcdat.open_mfdataset(directory_prefix+'/ne1024pg2_ne1024pg2_DY2/output.scream.SurfVars.INSTANT.nmins_x15.20*nc', add_bounds=True, parallel=True)

varname = "T_2m"
filespec = f_Jan_SurfVars
a = np.array(filespec.spatial.average(varname, lat_bounds=(0,10),axis=["X","Y"], weights=v1_weights)[varname].sel(time=slice('2020-01-20 00:00:00','2020-02-28 23:59:59')))
b = np.array(filespec.spatial.average(varname, lat_bounds=(80,90),axis=["X","Y"], weights=v1_weights)[varname].sel(time=slice('2020-01-20 00:00:00','2020-02-28 23:59:59')))
errmax = np.amax(np.abs(a-b))
print("%e, %e, %e" %(a[0],b[0], errmax))
```

produces:

```
2.852244e+02, 2.852244e+02, 0.000000e+00
```

### Relevant log output

_No response_

_No response_

### Environment

TBH I don't know how to do this.  `xcdat.show_versions()` throws an error.  I am using the e3sm_unified environment on NERSC - perlmutter which I loaded with
```
source /global/common/software/e3sm/anaconda_envs/load_latest_e3sm_unified_pm-cpu.sh
```

xarray.show_versions produces:
```
INSTALLED VERSIONS
------------------
commit: None
python: 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:23:14) [GCC 10.4.0]
python-bits: 64
OS: Linux
OS-release: 5.14.21-150400.24.46_12.0.73-cray_shasta_c
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: ('en_US', 'UTF-8')
libhdf5: 1.12.2
libnetcdf: 4.8.1

xarray: 2022.10.0
pandas: 1.5.2
numpy: 1.23.5
scipy: 1.9.3
netCDF4: 1.6.1
pydap: None
h5netcdf: None
h5py: None
Nio: None
zarr: None
cftime: 1.6.2
nc_time_axis: None
PseudoNetCDF: None
rasterio: None
cfgrib: None
iris: None
bottleneck: 1.3.5
dask: 2022.10.0
distributed: 2022.10.0
matplotlib: 3.6.2
cartopy: 0.21.0
seaborn: 0.12.1
numbagg: None
fsspec: 2022.11.0
cupy: None
pint: 0.20.1
sparse: 0.13.0
flox: None
numpy_groupies: None
setuptools: 65.5.1
pip: 22.3.1
conda: None
pytest: 7.2.0
IPython: 8.7.0
sphinx: None
```

### LINKED ISSUE #500: Investigate / document CDAT-xCDAT spatial differences (again)
A [recent issue](https://github.com/xCDAT/xcdat/issues/494) prompted me to look through ([and update](https://github.com/xCDAT/xcdat/pull/495)) the spatial averaging routine. 

I validated this PR by looking at monthly CMIP5/6 historical tas files (1 per model for 117 climate models) and [21 averaging domains](https://github.com/PCMDI/pcmdi_metrics/blob/27903488adefe1a2941100dfb56dceceb0321310/share/default_regions.py) (I skipped NAFM due to [this discussion](https://github.com/xCDAT/xcdat/issues/166#issuecomment-974591123)).

In validating the proposed changes, I have (again) come across CDAT-xCDAT differences. In 28 cases of 2415 cases (1.1%) the spatial averages are different (as judged by `np.allclose()`). The monthly differences are mostly below 0.01 K, but exceed 0.1 (in rare cases). These results apply to both the current release v0.5.0 and the proposed PR. 

I plan to summarize the reasons for the differences in this issue.

pochedls:
## Description
In #494, `.spatial.get_weights` and `.spatial.average()` do not work for multifile datasets when the longitude bounds span the prime meridian. This PR updates that logic to fix issues.

Issues and root causes:

> The next issue was that the longitude bounds wrap around a prime meridian in this dataset. This creates problems, because weights are determined by the difference between the longitude bound values (and a prime meridian introduces a discontinuity, e.g., `[359.6, 0.35]` in this case). The logic to correct this does not work for multifile datasets (which led to the error `xarray can't set arrays with multiple array indices to dask yet.` which came from a function `_force_domain_order_low_to_high`).
> 
> I also noticed that the `_force_domain_order_low_to_high` could produce incorrect weights for datasets that wrap a prime meridian (this didn't occur because of the above issues). 
> 
> I've addressed the multifile dataset issue and the prime meridian issues in PR #495, which will hopefully be included in our next release.
> 
> _Originally posted by @pochedls in https://github.com/xCDAT/xcdat/issues/494#issuecomment-1577568022_
>             

- Closes #494
- Closes #500 

## Checklist

If applicable:

### DIFF
--- a/tests/test_spatial.py
+++ b/tests/test_spatial.py
@@ -269,26 +269,17 @@ class TestGetWeights:
             decode_times=True, cf_compliant=False, has_bounds=True
         )
 
-    def test_bounds_reordered_when_upper_indexed_first(self):
+    def test_value_error_thrown_for_multiple_out_of_order_lon_bounds(self):
         domain_bounds = xr.DataArray(
             name="lon_bnds",
-            data=np.array(
-                [[-89.375, -90], [0.0, -89.375], [0.0, 89.375], [89.375, 90]]
-            ),
+            data=np.array([[3, 1], [5, 3], [5, 7], [7, 9]]),
             coords={"lat": self.ds.lat},
             dims=["lat", "bnds"],
         )
-        result = self.ds.spatial._force_domain_order_low_to_high(domain_bounds)
-
-        expected_domain_bounds = xr.DataArray(
-            name="lon_bnds",
-            data=np.array(
-                [[-90, -89.375], [-89.375, 0.0], [0.0, 89.375], [89.375, 90]]
-            ),
-            coords={"lat": self.ds.lat},
-            dims=["lat", "bnds"],
-        )
-        assert result.identical(expected_domain_bounds)
+        # Check _get_longitude_weights raises error when there are
+        # > 1 out-of-order bounds for the dataset.
+        with pytest.raises(ValueError):
+            self.ds.spatial._get_longitude_weights(domain_bounds, region_bounds=None)
 
     def test_raises_error_if_dataset_has_multiple_bounds_variables_for_an_axis(self):
         ds = self.ds.copy()
--- a/tests/test_spatial.py
+++ b/tests/test_spatial.py
@@ -180,6 +180,24 @@ class TestAverage:
 
         assert result.identical(expected)
 
+    def test_spatial_average_for_domain_wrapping_p_meridian_non_cf_conventions(
+        self,
+    ):
+        ds = self.ds.copy()
+
+        # get spatial average for original dataset
+        ref = ds.spatial.average("ts").ts
+
+        # change first bound from -0.9375 to 359.0625
+        lon_bnds = ds.lon_bnds.copy()
+        lon_bnds[0, 0] = 359.0625
+        ds["lon_bnds"] = lon_bnds
+
+        # check spatial average with new (bad) bound
+        result = ds.spatial.average("ts").ts
+
+        assert result.identical(ref)
+
     @requires_dask
     def test_spatial_average_for_lat_region_and_keep_weights_with_dask(self):
         ds = self.ds.copy().chunk(2)
--- a/xcdat/spatial.py
+++ b/xcdat/spatial.py
@@ -418,20 +418,15 @@ class SpatialAccessor:
         p_meridian_index: Optional[np.ndarray] = None
         d_bounds = domain_bounds.copy()
 
-        # check if there is more than one potential prime meridian cell
-        # there should only be one for rectilinear data
-        pmcells = np.where(domain_bounds[:, 1] - domain_bounds[:, 0] < 0)[0]
-        if len(pmcells) > 1:
+        pm_cells = np.where(domain_bounds[:, 1] - domain_bounds[:, 0] < 0)[0]
+        if len(pm_cells) > 1:
             raise ValueError(
-                "More than one longitude bound is out of order. Only one bound \n\
-                              spanning the prime meridian is permitted"
+                "More than one longitude bound is out of order. Only one bound "
+                "value spanning the prime meridian is permitted in data on "
+                "a rectilinear grid."
             )
-        # convert longitude bounds to 360 degree domain
         d_bounds: xr.DataArray = self._swap_lon_axis(d_bounds, to=360)  # type: ignore
-        # check for bounds spanning prime meridian
         p_meridian_index = _get_prime_meridian_index(d_bounds)
-        # if bounds span a prime meridian, ensure bounds are organized to span 0-360
-        # (without losing weight)
         if p_meridian_index is not None:
             d_bounds = _align_lon_bounds_to_360(d_bounds, p_meridian_index)
 
--- a/xcdat/spatial.py
+++ b/xcdat/spatial.py
@@ -281,12 +281,6 @@ class SpatialAccessor:
                     "to reference a specific data variable's axis bounds."
                 )
 
-            # The logic for generating longitude weights depends on the
-            # bounds being ordered such that d_bounds[:, 0] < d_bounds[:, 1].
-            # They are re-ordered (if need be) for the purpose of creating
-            # weights.
-            d_bounds = self._force_domain_order_low_to_high(d_bounds)
-
             r_bounds = axis_bounds[key]["region"]
 
             weights = axis_bounds[key]["weights_method"](d_bounds, r_bounds)
@@ -323,33 +317,6 @@ class SpatialAccessor:
 
             # Check the axis coordinate variable exists in the Dataset.
             get_dim_coords(self._dataset, key)
-
-    def _force_domain_order_low_to_high(self, domain_bounds: xr.DataArray):
-        """Reorders the ``domain_bounds`` low-to-high.
-
-        This method ensures all lower bound values are less than the upper bound
-        values (``domain_bounds[:, 1] < domain_bounds[:, 1]``).
-
-        Parameters
-        ----------
-        domain_bounds: xr.DataArray
-            The bounds of an axis.
-
-        Returns
-        ------
-        xr.DataArray
-            The bounds of an axis (re-ordered if applicable).
-        """
-        index_bad_cells = np.where(domain_bounds[:, 1] - domain_bounds[:, 0] < 0)[0]
-
-        if len(index_bad_cells) > 0:
-            new_domain_bounds = domain_bounds.copy()
-            new_domain_bounds[index_bad_cells, 0] = domain_bounds[index_bad_cells, 1]
-            new_domain_bounds[index_bad_cells, 1] = domain_bounds[index_bad_cells, 0]
-
-            return new_domain_bounds
-
-        return domain_bounds
 
     def _validate_region_bounds(self, axis: SpatialAxis, bounds: RegionAxisBounds):
         """Validates the ``bounds`` arg based on a set of criteria.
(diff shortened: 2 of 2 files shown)


tests/test_spatial.py
```diff
@@ -269,26 +269,17 @@ def setup(self):
...
-                [[-90, -89.375], [-89.375, 0.0], [0.0, 89.375], [89.375, 90]]
-            ),
-            coords={"lat": self.ds.lat},
-            dims=["lat", "bnds"],
-        )
-        assert result.identical(expected_domain_bounds)
+        # Check _get_longitude_weights raises error when there are
+        # > 1 out-of-order bounds for the dataset.
+        with pytest.raises(ValueError):
+            self.ds.spatial._get_longitude_weights(domain_bounds, region_bounds=None)
```
pochedls, line 282: The test here is much simpler than has been included in past iterations. 

It basically checks to see if there is more than one longitude bound in which the lower bound (i.e., `bound[0]`) is greater than the upper bound (i.e., `bounds[1]`). This can happen once for longitude (when spanning a prime meridian) but not more than once for rectilinear datasets. If it happens more than once, a `ValueError` is thrown.

xcdat/spatial.py
```diff
@@ -281,12 +281,6 @@ def get_weights(
                     "to reference a specific data variable's axis bounds."
                 )
 
-            # The logic for generating longitude weights depends on the
-            # bounds being ordered such that d_bounds[:, 0] < d_bounds[:, 1].
-            # They are re-ordered (if need be) for the purpose of creating
-            # weights.
-            d_bounds = self._force_domain_order_low_to_high(d_bounds)
-
```
pochedls, line 289: This is not necessary for latitude because we use the [absolute difference of the domain bounds](https://github.com/xCDAT/xcdat/blob/main/xcdat/spatial.py#L515). So the logic was updated in the `_get_longitude_weights()` function.

xcdat/spatial.py
```diff
@@ -324,33 +318,6 @@ def _validate_axis_arg(self, axis: List[SpatialAxis]):
...
-        if len(index_bad_cells) > 0:
-            new_domain_bounds = domain_bounds.copy()
-            new_domain_bounds[index_bad_cells, 0] = domain_bounds[index_bad_cells, 1]
-            new_domain_bounds[index_bad_cells, 1] = domain_bounds[index_bad_cells, 0]
-
-            return new_domain_bounds
-
-        return domain_bounds
-
```
pochedls, line 353: This is removed and replaced with a simpler consistency check and existing logic.

xcdat/spatial.py
```diff
@@ -441,12 +408,34 @@ def _get_longitude_weights(
...
         d_bounds = domain_bounds.copy()
 
+        # check if there is more than one potential prime meridian cell
+        # there should only be one for rectilinear data
+        pmcells = np.where(domain_bounds[:, 1] - domain_bounds[:, 0] < 0)[0]
+        if len(pmcells) > 1:
+            raise ValueError(
+                "More than one longitude bound is out of order. Only one bound \n\
+                              spanning the prime meridian is permitted"
+            )
```
pochedls, line 428: This is the consistency check to ensure there is only one instance in which `domain_bounds[:, 1] - domain_bounds[:, 0]` (this can happen for grid cells that span a prime meridian, but only once for rectilinear data).

xcdat/spatial.py
```diff
@@ -441,12 +408,34 @@ def _get_longitude_weights(
...
+        # convert longitude bounds to 360 degree domain
+        d_bounds: xr.DataArray = self._swap_lon_axis(d_bounds, to=360)  # type: ignore
+        # check for bounds spanning prime meridian
+        p_meridian_index = _get_prime_meridian_index(d_bounds)
+        # if bounds span a prime meridian, ensure bounds are organized to span 0-360
+        # (without losing weight)
+        if p_meridian_index is not None:
+            d_bounds = _align_lon_bounds_to_360(d_bounds, p_meridian_index)
+
```
pochedls, line 437: I pulled these lines of code out of the `if region_bounds is not None` block (below). This logic should apply whether there is a `region_bound` provided or not, because it helps to ensure the longitude bounds span 0 to 360 and are linear (high-to-low) which is important during the weights calculation.

tomvothecoder:
FYI I updated the PR title because it is used as the default commit message when the PR with "Squash and merge".
I also updated the description with a comment from #494 that covers the issue and causes.

xcdat/spatial.py
pochedls: This was originally put in this if statement, because the user could pass in region bounds in (-180, 180) or (0, 360). If the user passed in region bounds, we converted both to a (0, 360) domain and then handled prime meridian issues. 

The issue prompting this shows that for some datasets, we need to deal with the prime meridian issue (even if there are no region bounds). This doesn't occur in CMIP data (I think because CF-conventions state that bounds should be increasing). If bounds are increasing, there is no need for this complex logic.

xcdat/spatial.py
pochedls: Note that this also caused the error

> xarray can't set arrays with multiple array indices to dask yet.

because `d_bounds` were not loaded / copied.

tests/test_spatial.py
```diff
@@ -180,6 +180,24 @@ def test_spatial_average_for_lat_region(self):
...
+        # change first bound from -0.9375 to 359.0625
+        lon_bnds = ds.lon_bnds.copy()
+        lon_bnds[0, 0] = 359.0625
+        ds["lon_bnds"] = lon_bnds
+
+        # check spatial average with new (bad) bound
+        result = ds.spatial.average("ts").ts
+
+        assert result.identical(ref)
+
```
pochedls, line 200: Test for non-CF compliant bounds.

pochedls, line 200: Test that non-CF compliant bounds still lead to correct spatial averaging.

pochedls:
I think this can be reviewed now @tomvothecoder. I think I understand the spatial averaging issues we've had in the past (see #500).

xcdat/spatial.py
```diff
@@ -441,12 +408,34 @@ def _get_longitude_weights(
...
+                              spanning the prime meridian is permitted"
+            )
+        # convert longitude bounds to 360 degree domain
+        d_bounds: xr.DataArray = self._swap_lon_axis(d_bounds, to=360)  # type: ignore
+        # check for bounds spanning prime meridian
+        p_meridian_index = _get_prime_meridian_index(d_bounds)
+        # if bounds span a prime meridian, ensure bounds are organized to span 0-360
+        # (without losing weight)
+        if p_meridian_index is not None:
+            d_bounds = _align_lon_bounds_to_360(d_bounds, p_meridian_index)
```
tomvothecoder, line 436:
```suggestion
        # The longitude bounds are converted to the 360 degree domain if it
        # is not already. Afterwards, the index of the bounds with a prime
        # meridian cell is extracted (if it exists). If a prime meridian cell
        # exists, bounds are organized to span 0-360 in ascending order (without
        # losing weight).
        d_bounds: xr.DataArray = self._swap_lon_axis(d_bounds, to=360)  # type: ignore
        p_meridian_index = _get_prime_meridian_index(d_bounds)
        if p_meridian_index is not None:
            d_bounds = _align_lon_bounds_to_360(d_bounds, p_meridian_index)
```

xcdat/spatial.py
tomvothecoder, line 428:
```suggestion
        pm_cells = np.where(domain_bounds[:, 1] - domain_bounds[:, 0] < 0)[0]
        if len(pm_cells) > 1:
            raise ValueError(
                "More than one longitude bound is out of order. Only one bound "
                "value spanning the prime meridian is permitted in data on "
                "rectilinear grids."
            )
```

Removed comment and updated error message formatting and specifies rectilinear grids.

xcdat/spatial.py
tomvothecoder: Joined comments together and updated to specify "bounds are organized to span 0-360 in ascending order".

tomvothecoder (commented):
Hey @pochedls, your changes make sense to me. I added a few minor code suggestions. If you looped over datasets for validation like your normally do and the results checked out, feel free to merge. Thanks!

xcdat/spatial.py
tomvothecoder: @pochedls Is this the logic you mentioned we removed for forcing order of bounds? Just clarifying because I didn't remember if I was the one who removed it or not.

pochedls: @tomvothecoder – yes – this is the logic. Sorry I said you removed the logic (it seems that I removed it from this thread). Removing this check is okay for spatial average calculations.

When I was looking back at this, I think I realized one issue is that we don't call `_validate_region_bounds` when [generating spatial weights](https://github.com/xCDAT/xcdat/blob/main/xcdat/spatial.py#L206). This isn't a problem if you compute a spatial average (since the spatial averager does this check), but we might want to put this check in `get_weights`.

PR #539, opened 2023-08-31
Fix `create_grid` breaking with tuple containing `None`

### LINKED ISSUE #536: [Bug]: `create_grid` breaks if an arg is a tuple with the second element being `None`
### What happened?

I ran into this issue during e3sm_diags development while testing a function to reconstruct pressure levels. 
I passed a single pressure level (e.g., `plevs=[800]`) to `create_axis("lev", plevs, generate_bounds=True)` and it threw an error about generating bounds requiring  > 1 data points.

To workaround this, I tried setting `generate_bounds=False`. However, I ran into the issue of `create_grid` breaking because of a bug in the conditional for copying bounds as described below.

Steps to reproduce (refer to MVCE):

1. Use `z_axis = xc.create_axis(...generate_bounds=False)`, which will return a tuple consisting of `(xr.DataArray, None)`. 
2. Pass this tuple as an argument to `xc.create_grid(z=z_axis)`
3. `xc.create_grid` breaks on line 544 -- `AttributeError: 'NoneType' object has no attribute 'copy'`
   - This line attempts to deep copy bounds (the second element of the tuple), which breaks since it is `None`.
   - https://github.com/xCDAT/xcdat/blob/08cfea37cf47454abea83de66cf7677ab157708b/xcdat/regridder/grid.py#L544

The conditional to check the argument type for `create_grid` should also check if the second element is `None` and execute the appropriate logic

https://github.com/xCDAT/xcdat/blob/08cfea37cf47454abea83de66cf7677ab157708b/xcdat/regridder/grid.py#L532-L553

```python
import xcdat as xc

plevs = [800, 200]

# Create the output pressure grid to regrid to using the `plevs` array.
z_axis = xc.create_axis("lev", plevs, generate_bounds=False)
pressure_grid = xc.create_grid(z=z_axis)
```

### Relevant log output

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/global/homes/v/vo13/mambaforge/envs/e3sm_diags_dev_658/lib/python3.10/site-packages/xcdat/regridder/grid.py", line 544, in create_grid
    axis, bnds = item[0].copy(deep=True), item[1].copy(deep=True)  # type: ignore[union-attr]
AttributeError: 'NoneType' object has no attribute 'copy'
```

_No response_

### Environment

xcdat=0.6.0rc1 and main branch

tomvothecoder:
## Description

- Closes #536 

## Checklist

If applicable:

### DIFF
--- a/xcdat/regridder/grid.py
+++ b/xcdat/regridder/grid.py
@@ -529,28 +529,30 @@ def create_grid(
     axes = {"x": x, "y": y, "z": z}
     ds = xr.Dataset(attrs={} if attrs is None else attrs.copy())
 
-    for key, item in axes.items():
+    for axis, item in axes.items():
         if item is None:
             continue
 
         if isinstance(item, (tuple, list)):
             if len(item) != 2:
                 raise ValueError(
-                    f"Argument {key!r} should be an xr.DataArray representing "
+                    f"Argument {axis!r} should be an xr.DataArray representing "
                     "coordinates or a tuple (xr.DataArray, xr.DataArray) representing "
                     "coordinates and bounds."
                 )
 
-            axis, bnds = item[0].copy(deep=True), item[1].copy(deep=True)  # type: ignore[union-attr]
+            coords = item[0].copy(deep=True)
 
-            # ensure bnds attribute is set
-            axis.attrs["bounds"] = bnds.name
+            if item[1] is not None:
+                bnds = item[1].copy(deep=True)
 
-            ds = ds.assign({bnds.name: bnds})
+                coords.attrs["bounds"] = bnds.name
+
+                ds = ds.assign({bnds.name: bnds})
         else:
-            axis = item.copy(deep=True)
+            coords = item.copy(deep=True)
 
-        ds = ds.assign_coords({axis.name: axis})
+        ds = ds.assign_coords({coords.name: coords})
 
     return ds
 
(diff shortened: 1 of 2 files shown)


xcdat/regridder/grid.py
```diff
@@ -529,28 +529,30 @@ def create_grid(
...
-    for key, item in axes.items():
+    for axis, item in axes.items():
         if item is None:
             continue
 
         if isinstance(item, (tuple, list)):
             if len(item) != 2:
                 raise ValueError(
-                    f"Argument {key!r} should be an xr.DataArray representing "
+                    f"Argument {axis!r} should be an xr.DataArray representing "
```
tomvothecoder, line 539:
```suggestion
                    f"Argument `{axis}` should be an xr.DataArray representing "
```

tomvothecoder: I don't think we need `!r`.

jasonb5: `!r` adds the quotes so you don't need to include them in the f-string.

jasonb5 (approved):
LGTM

xcdat/regridder/grid.py
tomvothecoder: Thanks, that's a cool feature. I briefly Googled it and thought it meant "raw string" to make things like backslash a string.

jasonb5: Aw yea `r""` is a raw string, `!r` is just the `repr` of the variable.

PR #606, opened 2024-02-14
Feature/596 vertical averaging

### LINKED ISSUE #596: Add Z axis support for spatial averaging
### Discussed in https://github.com/xCDAT/xcdat/discussions/591

<div type='discussions-op-text'>

<sup>Originally posted by **tomvothecoder** January 29, 2024</sup>

I'm running into a case where I need to average over the Z axis. The CDAT code being used is `cdutil.averager(tvar, axis="z")`, which I need to replace.

xCDAT's [spatial averager](https://xcdat.readthedocs.io/en/stable/generated/xarray.Dataset.spatial.average.html#xarray.Dataset.spatial.average), which is based on `cdutil.averager()`, only supports rectilinear grids ("X" and "Y"). However, we do mention that the `get_weights()` method can be extended to support other axes:
https://github.com/xCDAT/xcdat/blob/fbf1db68fd6328d5a534e0c1e57e22e6fdd70c76/xcdat/spatial.py#L246-L252

### Are there are any possible answers you came across?
_No response_

_No response_

### Additional context

For the short-term, I might be able to use [xarray.DataArray.weighted](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.weighted.html) directly. However, I still need to generate the weights xr.DataArray beforehand.

```python
datarray.weighted(weights).mean("height")
```

pochedls:
## Description
This PR adds functionality to compute vertical averages. 

- Closes #596 

## Checklist

If applicable:


pochedls:
I'd appreciate testing with real-world data (and comparisons with CDAT).

tests/test_spatial.py
```diff
@@ -45,6 +45,35 @@ def test_raises_error_if_data_var_not_in_dataset(self):
...
+
+        # check that vertical averaging returns the correct weights
+        expected = xr.DataArray(
+            data=np.array([2000, 2000, 1000, 0.0]),
+            coords={"time": ds.time, "lev": ds.lev, "lat": ds.lat, "lon": ds.lon},
+            dims=["lev"],
+            attrs={"xcdat_bounds": True},
+        )
+
+        xr.testing.assert_allclose(result["lev_wts"], expected)
```
pochedls, line 75: Are two-in-one tests permitted?

xcdat/spatial.py
```diff
@@ -73,10 +73,12 @@ def average(
         keep_weights: bool = False,
         lat_bounds: Optional[RegionAxisBounds] = None,
         lon_bounds: Optional[RegionAxisBounds] = None,
+        lev_bounds: Optional[RegionAxisBounds] = None,
```
pochedls, line 76: I think `lev_bounds` is a good generic name, but I'm open to other possibilities.

xcdat/spatial.py
```diff
@@ -73,10 +73,12 @@ def average(
...
     ) -> xr.Dataset:
         """
-        Calculates the spatial average for a rectilinear grid over an optionally
-        specified regional domain.
+        Calculates the weighted spatial and/or vertical average for a
+        rectilinear grid over an optionally specified regional and/or vertical
+        domain.
```
pochedls, line 81:
```suggestion
        Calculates the weighted spatial and/or vertical average for a
        rectilinear grid for an optionally specified regional and/or vertical
        domain.
```

xcdat/spatial.py
```diff
@@ -476,6 +509,32 @@ def _get_latitude_weights(
...
+        -------
+        xr.DataArray
+            The vertical axis weights.
+        """
+        if region_bounds is not None:
+            domain_bounds = self._scale_domain_to_region(domain_bounds, region_bounds)
+
+        weights = self._calculate_weights(domain_bounds)
+        return weights
+
```
pochedls, line 537: I think this is sufficiently different from `_get_longitude_weights` (which deals with the prime meridian) and `_get_latitude_weights` (sine of bounds) to justify its own function.

tomvothecoder:
Tom:
- [ ] Compare with cdms2 average with Z axis and get floating point diffs
- [ ] Try out in e3sm_diags

tomvothecoder:
I'm going to pick this back up for review. We might be able to get this in for v0.7.0.

xcdat/spatial.py
tomvothecoder: I agree

xcdat/spatial.py
tomvothecoder: Sounds good to me

tests/test_spatial.py
tomvothecoder: Yeah there's no problem with that here. As long as the tests are relatively easy to maintain.

tests/test_spatial.py
```diff
@@ -45,6 +45,35 @@ def test_raises_error_if_data_var_not_in_dataset(self):
...
 
+    def test_vertical_average_with_weights(self):
+        # check that vertical averaging returns the correct answer
+        # get dataset with vertical levels
+        ds = generate_lev_dataset()
+        # subset to one column for testing (and shake up data)
+        ds = ds.isel(time=[0], lat=[0], lon=[0]).squeeze()
+        so = ds["so"]
+        so[:] = np.array([1, 2, 3, 4])
+        ds["so"] = so
```
tomvothecoder, line 56:
```suggestion
        ds["so"][:] = np.array([1, 2, 3, 4])
```

Assigning the numpy array directly to the DataArray will work too

pochedls:
> > I'd appreciate testing with real-world data (and comparisons with CDAT).
> 
> Can you provide me some datasets to test with? I wrote a script and tested a single E3SM dataset, which produced nearly identical results (1e-16 max rel diff).

Do you have xsearch running? I think you could look for monthly, historical data, for the `thetao` or `ta` (Amon table) variables.

tomvothecoder:
> > > I'd appreciate testing with real-world data (and comparisons with CDAT).
> > 
> > 
> > Can you provide me some datasets to test with? I wrote a script and tested a single E3SM dataset, which produced nearly identical results (1e-16 max rel diff).
> 
> Do you have xsearch running? I think you could look for monthly, historical data, for the `thetao` or `ta` (Amon table) variables.

I'll check on `xsearch`. Thanks for the pointers.

PR #672, opened 2024-06-28
Add weight threshold option for spatial averaging 

### LINKED ISSUE #531: [Enhancement]: Add weight threshold option for averaging operations

When xCDAT performs an averaging operation (spatial or temporal) with missing data, it assigns missing values a weight of zero (which is correct). But this can be misleading if some or the majority of data is missing. Imagine if data was missing in spring, fall, and summer (leaving only winter data). xCDAT would take the annual average and report a very cold annual average temperature. A user is usually aware of this kind of thing, but may miss it if a small number of grid cells are missing part of the dataset (or are missing anomaly values, which would be harder to recognize as "weird"). 

See the example below: 

```
import xcdat as xc
import numpy as np
# load example dataset
fn = '/p/css03/esgf_publish/CMIP6/CMIP/NCAR/CESM2/amip/r1i1p1f1/Amon/tas/gn/v20190218/tas_Amon_CESM2_amip_r1i1p1f1_gn_195001-201412.nc'
ds = xc.open_dataset(fn)
# select 12 months of data (for some midlatitude grid cell over Asia)
ds = ds.isel(time=np.arange(0, 12), lat=[150], lon=[50]).squeeze()
print('range of seasonal cycle: ' + str(np.min(ds.tas.values)) + ' - ' + str(np.max(ds.tas.values)))
print('annual average: ' + str(ds.temporal.group_average('tas', freq='year').tas.values[0]))
# NaN out data over Jan - Nov
ds['tas'][0:11] = np.nan
print('annual average (with NaNs): ' + str(ds.temporal.group_average('tas', freq='year').tas.values[0]))
```

> range of seasonal cycle: 260.47635 - 297.10477
> annual average: 278.48889160156256
> annual average (with NaNs Jan - Nov): 265.1025406006844

A similar situation can arise if a timestep of observations is missing part of the field during spatial averaging operations (e.g., missing the tropics, leading to too cold global temperatures). 

This would need to be mapped out more, but it might be useful to have a `weight_threshold` optional argument that allows the user to specify the minimum temporal/spatial weight needed to produce a value. For example `weight_threshold=0.9` would require 90% of the spatial / temporal data within the spatial or temporal averaging window be present. 

CDAT had similar functionality for temporal averaging calculations (see [Specifying Data Coverage Criteria](https://cdat.llnl.gov/documentation/utilities/utilities-1.html). I'm not sure if there was any similar functionality for spatial averaging. 

_No response_

### Additional context

_No response_

pochedls:
## Description

This PR adds an optional argument that requires a minimum fraction of data be available to perform a spatial average. The initial PR is for spatial averaging only (it would need to be expanded to handle temporal averaging). 

- Closes #531

## Checklist

If applicable:

### DIFF
--- a/tests/test_spatial.py
+++ b/tests/test_spatial.py
@@ -15,13 +15,13 @@ class TestSpatialAccessor:
         )
 
     def test__init__(self):
-        ds = self.ds.copy()
+        ds = self.ds.copy(deep=True)
         obj = SpatialAccessor(ds)
 
         assert obj._dataset.identical(ds)
 
     def test_decorator_call(self):
-        ds = self.ds.copy()
+        ds = self.ds.copy(deep=True)
         obj = ds.spatial
 
         assert obj._dataset.identical(ds)
@@ -50,7 +50,7 @@ class TestAverage:
             self.ds.spatial.average("ts", axis=["Y", "incorrect_axis"])
 
     def test_raises_error_if_lat_axis_coords_cant_be_found(self):
-        ds = self.ds.copy()
+        ds = self.ds.copy(deep=True)
 
         # Update CF metadata to invalid values so cf_xarray can't interpret them.
         del ds.lat.attrs["axis"]
@@ -64,7 +64,7 @@ class TestAverage:
             ds.spatial.average("ts", axis=["X", "Y"])
 
     def test_raises_error_if_lon_axis_coords_cant_be_found(self):
-        ds = self.ds.copy()
+        ds = self.ds.copy(deep=True)
 
         # Update CF metadata to invalid values so cf_xarray can't interpret them.
         del ds.lon.attrs["axis"]
@@ -141,13 +141,13 @@ class TestAverage:
             self.ds.spatial.average("ts", axis=["X", "Y"], weights=weights)
 
     def test_spatial_average_for_lat_region_and_keep_weights(self):
-        ds = self.ds.copy()
+        ds = self.ds.copy(deep=True)
 
         result = ds.spatial.average(
             "ts", axis=["Y"], lat_bounds=(-5.0, 5), keep_weights=True
         )
 
-        expected = self.ds.copy()
+        expected = self.ds.copy(deep=True)
         expected["ts"] = xr.DataArray(
             data=np.array(
                 [[2.25, 2.25, 2.25, 2.25], [1.0, 1.0, 1.0, 1.0], [1.0, 1.0, 1.0, 1.0]]
@@ -164,12 +164,12 @@ class TestAverage:
         xr.testing.assert_allclose(result, expected)
 
     def test_spatial_average_for_lat_region(self):
-        ds = self.ds.copy()
+        ds = self.ds.copy(deep=True)
 
         # Specifying axis as a str instead of list of str.
         result = ds.spatial.average("ts", axis=["Y"], lat_bounds=(-5.0, 5))
 
-        expected = self.ds.copy()
+        expected = self.ds.copy(deep=True)
         expected["ts"] = xr.DataArray(
             data=np.array(
                 [[2.25, 2.25, 2.25, 2.25], [1.0, 1.0, 1.0, 1.0], [1.0, 1.0, 1.0, 1.0]]
@@ -205,7 +205,7 @@ class TestAverage:
     def test_spatial_average_for_domain_wrapping_p_meridian_non_cf_conventions(
         self,
     ):
-        ds = self.ds.copy()
+        ds = self.ds.copy(deep=True)
 
         # get spatial average for original dataset
         ref = ds.spatial.average("ts").ts
@@ -222,14 +222,14 @@ class TestAverage:
 
     @requires_dask
     def test_spatial_average_for_lat_region_and_keep_weights_with_dask(self):
-        ds = self.ds.copy().chunk(2)
+        ds = self.ds.copy(deep=True).chunk(2)
 
         # Specifying axis as a str instead of list of str.
         result = ds.spatial.average(
             "ts", axis=["Y"], lat_bounds=(-5.0, 5), keep_weights=True
         )
 
-        expected = self.ds.copy()
+        expected = self.ds.copy(deep=True)
         expected["ts"] = xr.DataArray(
             data=np.array(
                 [[2.25, 2.25, 2.25, 2.25], [1.0, 1.0, 1.0, 1.0], [1.0, 1.0, 1.0, 1.0]]
@@ -246,7 +246,7 @@ class TestAverage:
         xr.testing.assert_allclose(result, expected)
 
     def test_spatial_average_for_lat_and_lon_region_and_keep_weights(self):
-        ds = self.ds.copy()
+        ds = self.ds.copy(deep=True)
         result = ds.spatial.average(
             "ts",
             axis=["X", "Y"],
@@ -255,7 +255,7 @@ class TestAverage:
             keep_weights=True,
         )
 
-        expected = self.ds.copy()
+        expected = self.ds.copy(deep=True)
         expected["ts"] = xr.DataArray(
             data=np.array([2.25, 1.0, 1.0]),
             coords={"time": expected.time},
@@ -277,7 +277,7 @@ class TestAverage:
         xr.testing.assert_allclose(result, expected)
 
     def test_spatial_average_for_lat_and_lon_region_with_custom_weights(self):
-        ds = self.ds.copy()
+        ds = self.ds.copy(deep=True)
 
         weights = xr.DataArray(
             data=np.array([[1, 2, 3, 4], [2, 4, 6, 8], [3, 6, 9, 12], [4, 8, 12, 16]]),
@@ -292,7 +292,7 @@ class TestAverage:
             data_var="ts",
         )
 
-        expected = self.ds.copy()
+        expected = self.ds.copy(deep=True)
         expected["ts"] = xr.DataArray(
             data=np.array([2.25, 1.0, 1.0]),
             coords={"time": expected.time},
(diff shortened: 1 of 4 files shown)


xcdat/spatial.py
```diff
@@ -729,11 +741,48 @@ def _averager(self, data_var: xr.DataArray, axis: List[SpatialAxis]):
...
+        # if weight thresholds applied, calculate fraction of data availability
+        # replace values that do not meet minimum weight with nan
+        if required_weight > 0.0:
+            # sum all weights (assuming no missing values exist)
+            print(dim)
+            weight_sum_all = weights.sum(dim=dim)  # type: ignore
+            # zero out cells with missing values in data_var
+            weights = xr.where(~np.isnan(data_var), weights, 0)
+            # sum all weights (including zero for missing values)
+            weight_sum_masked = weights.sum(dim=dim)  # type: ignore
```
pochedls, line 782: Needed to add `type: ignore` for unclear reasons...

xcdat/spatial.py
```diff
@@ -716,6 +725,9 @@ def _averager(self, data_var: xr.DataArray, axis: List[SpatialAxis]):
             Data variable inside a Dataset.
         axis : List[SpatialAxis]
             List of axis dimensions to average over.
+        required_weight : optional, float
```
pochedls, line 728: Is `required_weight` a good parameter name? What should the default be (currently zero to make default behavior backwards compatible).

xcdat/spatial.py
```diff
@@ -729,11 +741,48 @@ def _averager(self, data_var: xr.DataArray, axis: List[SpatialAxis]):
         """
         weights = self._weights.fillna(0)
```
pochedls, line 742: I don't think this does anything.

xcdat/spatial.py
tomvothecoder: Maybe `required_weight_pct`?  `required_weight` sounds like a boolean parameter.

I think 0 is the appropriate default value.

xcdat/spatial.py
tomvothecoder: I recall we ran into issues with Xarray's weighted mean API when weights contain `np.nan`. The notes above it say:

https://github.com/xCDAT/xcdat/blob/cdae8263b0ce279ffebc36cad98a95e3ee20a5e5/xcdat/spatial.py#L727-L728

xcdat/spatial.py
tomvothecoder: Actually I like `None` as a default value to indicate the arg is not set by default.

tomvothecoder:
I'm going to open a separate PR to address this enhancement with the temporal APIs because those require more work and we can merge this earlier when it is ready.

tests/test_spatial.py
```diff
@@ -254,6 +265,49 @@ def test_spatial_average_for_lat_and_lon_region_and_keep_weights(self):
...
+        )
+
+        expected = self.ds.copy()
+        expected["ts"] = xr.DataArray(
+            data=np.array([2.25, 1.0, 1.0]),
+            coords={"time": expected.time},
+            dims="time",
+        )
+
+        xr.testing.assert_allclose(result, expected)
```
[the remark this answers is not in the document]
tomvothecoder: Using `Optional` indicates that the argument can also be `None`. You can just specify `float` as the type annotation if `None` is not expected.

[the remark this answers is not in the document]
tomvothecoder: I like `None` as the default value

xcdat/spatial.py
tomvothecoder: I think `dim` is expected to be a string but we pass a Hashable (dict key).

`# type: ignore` is fine here

xcdat/spatial.py
```diff
@@ -729,11 +741,47 @@ def _averager(self, data_var: xr.DataArray, axis: List[SpatialAxis]):
...
+
+        if required_weight > 1.0:
+            raise ValueError(
+                "required_weight argment is greater than zero. "
+                "required_weight must be between 0 and 1."
+            )
+
+        # need weights to match data_var dimensionality
+        if required_weight > 0.0:
+            weights, data_var = xr.broadcast(weights, data_var)
```
tomvothecoder, line 762:
```suggestion
        # ensure required weight is between 0 and 1
        if required_weight is None:
            required_weight = 0.0
        elif required_weight < 0.0:
            raise ValueError(
                "required_weight argument is less than 0. "
                "required_weight must be between 0 and 1."
            )

        if required_weight > 1.0:
            raise ValueError(
                "required_weight argument is greater than 1. "
                "required_weight must be between 0 and 1."
            )

        # need weights to match data_var dimensionality
        if required_weight > 0.0:
            weights, data_var = xr.broadcast(weights, data_var)
```

tomvothecoder:
I started PR #683 for temporal operations. I used some of your code and split them up into reusable functions. We can think about making these functions generalizable across the spatial and temporal classes. 

Check them out here: https://github.com/xCDAT/xcdat/pull/683/files

pochedls (commented):
Review review and apply suggestions.

xcdat/spatial.py
tomvothecoder: We can just keep `# type: ignore`. The [docstring](https://github.com/pydata/xarray/blob/19a0428005f255da1ae538d8db682f7f6f83cc4f/xarray/core/_aggregations.py#L1783-L1785) for `dim` says:


> Name of dimension[s] along which to apply sum. For e.g. dim="x" or dim=["x", "y"]. If “…” or None, will reduce over all dimensions.

Although the type annotation for dim is `Dims = Union[str, Collection[Hashable], "ellipsis", None]`, the example includes a list (`dim=["x", "y"])`,

xcdat/spatial.py
pochedls: `min_weight` is good with me!

tomvothecoder:
I've created general functions for weight thresholds [here](https://github.com/xCDAT/xcdat/pull/683/files). You can copy them over to `utils.py` in your branch and then use them in the `SpatialAccessor` class as needed.

xcdat/spatial.py
```diff
@@ -729,11 +743,45 @@ def _averager(self, data_var: xr.DataArray, axis: List[SpatialAxis]):
...
+            )
+        elif minimum_weight > 1.0:
+            raise ValueError(
+                "minimum_weight argument is greater than 1. "
+                "minimum_weight must be between 0 and 1."
+            )
+
+        # need weights to match data_var dimensionality
+        if minimum_weight > 0.0:
+            weights, data_var = xr.broadcast(weights, data_var)
```
tomvothecoder, line 762: I think we can remove this line.

Per docs for [xr.broadcast](https://docs.xarray.dev/en/stable/generated/xarray.broadcast.html):

xarray objects automatically broadcast against each other in arithmetic operations, so this function should not be necessary for normal use.

pochedls: I **think** the weights are probably of dimensions `[lat, lon]` and this casts them to `[time, lat, lon]`. That way when you sum them below you get `[time]` (instead of a dimensionless scalar) which is then used to figure out which values do not meet the weight threshold and need to be masked.

tomvothecoder: Xarray should be handling the automatic broadcasting when using the `xr.where()` API to mask the data using weights. Otherwise, it would probably break with an error that the dimensions don't align. 

According to the [docs](https://docs.xarray.dev/en/stable/generated/xarray.where.html) for `xr.where()`:

> Performs xarray-like broadcasting across input arguments.

tomvothecoder: The unit tests pass with this line commented out. I also didn't implement `xr.broadcast()` explicitly in PR #683.

tomvothecoder:
Rebased this branch on latest `main` and pushed commit [`b56befe` (#672)](https://github.com/xCDAT/xcdat/pull/672/commits/b56befeae0e872e0c26a83bbb36b55132c2cc102) with code review updates (refer to commit message for list of changes).

tomvothecoder:
FYI Just did a review of the `utils.py` and it is the same as PR #683. We can proceed with testing this PR with real datasets. I'll do another review of the code.

pochedls:
@tomvothecoder – this feature stops working correctly at [this commit](https://github.com/xCDAT/xcdat/pull/672/commits/636366b2ae3bd19c0cbe2846b8dc8e93dcf2140c). 

I think the issue arises, because [this chunk](https://github.com/xCDAT/xcdat/blob/1c5c3c1af136fda7f86a767323ae2c9367427334/xcdat/spatial.py#L778-L788) of code is put into a separate function (`_mask_var_with_weight_threshold` and `_get_masked_weights`) in the [subsequent commit](https://github.com/xCDAT/xcdat/pull/672/commits/636366b2ae3bd19c0cbe2846b8dc8e93dcf2140c). 

The problem is that the original code operates on the full `[time, lat, lon]` dataarray when it computes the amount of weight if you exclude missing values: `weights = xr.where(~np.isnan(data_var), weights, 0)`. In the subsequent commit, it seems like the logic is updated to operate on the *already averaged* array (dimensions of `[time]`). It therefore can't accurately determine the amount of weight in the original 3D dataarray (it is giving an answer and not throwing an error because of broadcasting). 

Should I roll this back to the commit that was working? Or try to refactor from the latest commit? I think part of the issue is that you moved some of the logic to utils (which may work for the temporal version of this code). 

For testing, this should yield some `NaN` values:

```python
import xcdat as xc
fn = '/p/user_pub/climate_work/pochedley1/surface/HadCRUT.4.6.0.0.median.nc'
ds = xc.open_dataset(fn)
ts = ds.spatial.average('temperature_anomaly', min_weight=1.0)
```

xcdat/spatial.py
```diff
@@ -746,45 +753,68 @@ def _averager(
...
+        weight_sum_masked = masked_weights.sum(dim=dim)
+
+        # Get fraction of the available weight.
+        frac = weight_sum_masked / weight_sum_all
+
+        # Nan out values that don't meet specified weight threshold.
+        dv_new = xr.where(frac >= min_weight, dv, np.nan, keep_attrs=True)
+        dv_new.name = dv.name
+
+        return dv_new
```
tomvothecoder: > @tomvothecoder – this feature stops working correctly at [this commit](https://github.com/xCDAT/xcdat/pull/672/commits/636366b2ae3bd19c0cbe2846b8dc8e93dcf2140c).
> 
> I think the issue arises, because [this chunk](https://github.com/xCDAT/xcdat/blob/1c5c3c1af136fda7f86a767323ae2c9367427334/xcdat/spatial.py#L778-L788) of code is put into a separate function (`_mask_var_with_weight_threshold` and `_get_masked_weights`) in the [subsequent commit](https://github.com/xCDAT/xcdat/pull/672/commits/636366b2ae3bd19c0cbe2846b8dc8e93dcf2140c).
> 
> The problem is that the original code operates on the full `[time, lat, lon]` dataarray when it computes the amount of weight if you exclude missing values: `weights = xr.where(~np.isnan(data_var), weights, 0)`. In the subsequent commit, it seems like the logic is updated to operate on the _already averaged_ array (dimensions of `[time]`). It therefore can't accurately determine the amount of weight in the original 3D dataarray (it is giving an answer and not throwing an error because of broadcasting).
> 
> Should I roll this back to the commit that was working? Or try to refactor from the latest commit? I think part of the issue is that you moved some of the logic to utils (which may work for the temporal version of this code).
> 
> For testing, this should yield some `NaN` values:
> 
> ```python
> import xcdat as xc
> fn = '/p/user_pub/climate_work/pochedley1/surface/HadCRUT.4.6.0.0.median.nc'
> ds = xc.open_dataset(fn)
> ts = ds.spatial.average('temperature_anomaly', min_weight=1.0)
> ```

I see what you're saying. Let's keep the current commit.

I think we just need to update `_mask_var_with_weight_threshold()` with parameters for `dv` (original var) and `dv_mean` (averaged var). All current references to `dv` should be `dv_mean`, while line 810 (`masked_weights = `_get_masked_weights(dv, weights)` should be kept the same to properly calculate weights using `dv`.

I can make this update soon unless you want to give it a shot.

tomvothecoder (commented):
@pochedls This commit should fix the issue mentioned [here](https://github.com/xCDAT/xcdat/pull/672#issuecomment-2675901221). I did not test yet.

pochedls:
I wrote a script to sanity check this (it seemed to do what I thought it should): 

```python
# %% imports
import sys
sys.path.append('/home/pochedley1/code/xcdat/')
import xcdat as xc
import numpy as np
import xarray as xr

# %% parameters
fn = '/p/user_pub/climate_work/pochedley1/surface/HadCRUT.4.6.0.0.median.nc'
weight_levels = np.arange(0., 1.01, 0.05)
vid = 'temperature_anomaly'
lat_bounds = (-90, 90)

# open dataset
ds = xc.open_dataset(fn)
# get weight matrix
weights = ds.spatial.get_weights(axis=["X", "Y"], lat_bounds=lat_bounds)
# loop over weight thresholds
for wl in weight_levels:
    # get matching dimensionality
    wb, da = xr.broadcast(weights, ds[vid])
    # put time first
    wb = wb.transpose("time", "latitude", "longitude")
    da = da.transpose("time", "latitude", "longitude")
    # zero weights where the underlying dataarray is NaN
    wb = wb.where(~np.isnan(da), 0)
    # compute the fraction of weight available at each time step
    f = wb.sum(dim=['latitude', 'longitude']) / weights.sum(dim=['latitude', 'longitude'])
    # take the spatial average with the new weight thresholding functionality
    ts = ds.spatial.average(vid, min_weight=wl, lat_bounds=lat_bounds)[vid]
    # get indices where there should be data (i.e., more data is available than the fractional threshold)
    I1 = np.where(f >= wl)[0]
    # get indices where the spatial averager returns a non-nan solution
    I2 = np.where(~np.isnan(ts))[0]
    # ensure arrays are equal
    # else throw a valueerror to alert us to a problem
    if np.array_equal(I1, I2):
        print(str(wl) + ' checks out...')
        n = len(ds.time)
        print(str(n - len(I1)) + ' timesteps left out')
        print()
        continue
    else:
        raise ValueError('Mismatch for weight threshold ' + str(wl))
```

I think it would be helpful if @lee1043 or someone else could come up with a different way to test this functionality.

tomvothecoder:
CDAT's spatial averaging function ([genutil.averager()](https://github.com/CDAT/genutil/blob/59517ab54e65c03098502f63434270f96020f3eb/Lib/averager.py#L856-L857)) does not have a `criteriaarg` argument for specifying weight threshold. This argument only exists for temporal averaging (["Specifying Data Coverage Criteria"](https://cdat.llnl.gov/documentation/utilities/utilities-1.html)).

There are unit tests already written with extremely simple cases. I'll look to expand the test suite if there are more edge cases to cover. 

The unit tests along with some manual testing on different model data should be sufficient enough since the implementation is relatively simple. Once we're confident, we can move forward with this PR.

xcdat/spatial.py
```diff
@@ -114,17 +119,21 @@ def average(
...
+        lon_bounds : RegionAxisBounds | None, optional
             A tuple of floats/ints for the regional longitude lower and upper
             boundaries. This arg is used when calculating axis weights, but is
             ignored if ``weights`` are supplied. The lower bound can be larger
             than the upper bound (e.g., across the prime meridian, dateline), by
             default None.
+        min_weight : optional, float
+            Fraction of data coverage (i.e, weight) needed to return a
+            spatial average value. Value must range from 0 to 1, by default None
+            (equivalent to ``min_weight=0.0``).
```
tomvothecoder, line 136: I suggest adding how/why the `min_weight` is useful to set.

lee1043:
The new parameter `min_weight` is working as expected for `spatial.average`.  Parameter name is straightforward. Please see the detailed testing with a real world data -- HadISST from obs4MIPs, which has NaN values over land (time-fixed) and sea-ice (time-varying). 

[test_spatial_avg_20250423.pdf](https://github.com/user-attachments/files/19873289/test_spatial_avg_20250423.pdf)

lee1043 (approved):

tomvothecoder:
> The new parameter `min_weight` is working as expected for `spatial.average`. Parameter name is straightforward. Please see the detailed testing with a real world data -- HadISST from obs4MIPs, which has NaN values over land (time-fixed) and sea-ice (time-varying).
> 
> [test_spatial_avg_20250423.pdf](https://github.com/user-attachments/files/19873289/test_spatial_avg_20250423.pdf)

Thank you Jiwoo! I will do a final review and merge this PR soon.

xcdat/spatial.py
```diff
@@ -732,13 +764,81 @@ def _averager(
...
+        dim: List[str],
+        weights: xr.DataArray,
+        min_weight: float,
+    ) -> xr.DataArray:
+        """Mask values that do not meet the minimum weight threshold with np.nan.
+
+        This function is useful for cases where the weighting of data might be
+        skewed based on the availability of data. For example, if a portion of
+        cells in a region has significantly more missing data than other other
+        regions, it can result in inaccurate calculations of spatial averaging.
```
tomvothecoder, line 797:
```suggestion
        regions, it can result in inaccurate spatial average calculations.
```

tomvothecoder (commented):
LGTM, will merge.

tomvothecoder (approved):

PR #696, opened 2024-09-21
Update temporal.py to properly handle piControl simulations

### LINKED ISSUE #695: [Bug]: Potential bug in temporal.py module
### What happened?

I recently used the PCMDI metrics package to process the diagnostics for the piControl simulations from the E3SM model. As the model time for the piControl starts from year 1,  issues arises when the PCMDI calls the function **_drop_incomplete_djf ()** in "xcdat/temporal.py", specifically, I found the error are pointed to line [1087](https://github.com/xCDAT/xcdat/blob/753a0467b3b9df8925f3fadfda7158c188fd36c7/xcdat/temporal.py#L1087): 
        
`incomplete_seasons = (f"{start_year}-01", f"{start_year}-02", f"{end_year}-12")
`

The problem to my understanding here is that the **start_year**  and **end_year** in the piControl simulations (e.g. year 1 to year 50) are not 4-digit format, then the **_drop_incomplete_djf ()** complains in my case that: 

> warning: failed for  v3-SORRM 0051 no ISO-8601 or cftime-string-like match for string: 1-01 
  
I think the function actually expect the "0001-01" rather than "1-01", which caused the problems for my case. 

I think that it would be safer to adjust the [1087](https://github.com/xCDAT/xcdat/blob/753a0467b3b9df8925f3fadfda7158c188fd36c7/xcdat/temporal.py#L1087) in  "xcdat/temporal.py" to 

`incomplete_seasons = (f"{start_year:04d}-01", f"{start_year:04d}-02", f"{end_year:04d}-12")
`
In this way, the 4-digt year will be guaranteed, and my test simulation can pass the above error for piControl simulations. 

I reported this use case here to see if this would be an indicator for some code improvements. 

_No response_

_No response_

### Relevant log output

_No response_

_No response_

### Environment

The xCDAT is part of library for PCMDI metrics (https://github.com/PCMDI/pcmdi_metrics). In my case,  xcdat version 0.7.1 was complied with the python 3.10 and used as the library for the mcmdi metrics.

lee1043:
## Description
PiControl simulations can have time step labels as like "0001-01-01". To not lose the first "000" in year, #695 proposed this fix.

- Closes #695
- Convert `self.dim` in `TemporalAccessor` from Hashable to a string to fix mypy warning and remove type ignore comments

## Checklist

If applicable:

### DIFF
--- a/xcdat/temporal.py
+++ b/xcdat/temporal.py
@@ -1084,7 +1084,11 @@ class TemporalAccessor:
             ds[self.dim].dt.year.values[0],
             ds[self.dim].dt.year.values[-1],
         )
-        incomplete_seasons = (f"{start_year:04d}-01", f"{start_year:04d}-02", f"{end_year:04d}-12")
+        incomplete_seasons = (
+            f"{start_year:04d}-01",
+            f"{start_year:04d}-02",
+            f"{end_year:04d}-12",
+        )
 
         for year_month in incomplete_seasons:
             try:
--- a/xcdat/temporal.py
+++ b/xcdat/temporal.py
@@ -815,7 +815,7 @@ class TemporalAccessor:
         # it becomes obsolete after the data variable is averaged. When the
         # averaged data variable is added to the dataset, the new time dimension
         # and its associated coordinates are also added.
-        ds = ds.drop_dims(self.dim)  # type: ignore
+        ds = ds.drop_dims(self.dim)
         ds[dv_avg.name] = dv_avg
 
         if keep_weights:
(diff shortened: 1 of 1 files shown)


lee1043:
pre-commit shows following error for `mypy` failing, which I don't have a good idea. @tomvothecoder can you help on this?

```
$ pre-commit run --files temporal.py 
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...........................................(no files to check)Skipped
black....................................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

xcdat/temporal.py:1288: error: Argument 1 to "groupby" of "DataArray" has incompatible type "Any | Hashable"; expected "str | DataArray | IndexVariable | Sequence[Hashable] | Mapping[Any, Grouper] | None"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

tomvothecoder (approved):
I fixed the `mypy` warning that was appearing with the latest version of `xarray`. This PR looks good to me. Thank you Jiwoo.

lee1043:
@tomvothecoder thank you for your review! I am merging it now.

PR #708, opened 2024-10-07
Update `get_bounds()` to support mappable non-CF axes using `"bounds"` attr

### LINKED ISSUE #707: [Bug]: `get_bounds()` can't get bounds for non-CF axes without `var_key` argument
### What happened?

Notice in the `get_bounds()` conditional below, we retrieve bounds for the Dataset object using `_get_bounds_keys()` (which does not work with non-CF axes). On the other hand, if we pass `var_key`, we get the bounds using the "bounds" attribute of non-CF axes (which works fine).
https://github.com/xCDAT/xcdat/blob/e3dda64feab046046b6a3ee8047faec1ab01f8ec/xcdat/bounds.py#L233-L250 

https://github.com/xCDAT/xcdat/blob/e3dda64feab046046b6a3ee8047faec1ab01f8ec/xcdat/bounds.py#L477-L508

We should be able to get bounds for non-CF compliant axes that xCDAT can map to via dimension name (e.g., "lev", "lat", "lon"). We need to update `_get_bounds_keys()` to consider the "bounds" attribute.

```python
# %%
import xcdat as xc
import xarray as xr

# Create a dummy dataset
data = xr.DataArray([[1, 2, 3], [4, 5, 6]], dims=("lat", "lon"))
ds = xr.Dataset({"data": data})

# Add a "lev" Z axis with bounds
lev = xr.DataArray([1000, 2000], dims=("lev"))
ds["lev"] = lev
ds = ds.bounds.add_missing_bounds(axes="Z")

# %%
# Notice how we can map to "lev" without CF attributes because "lev" is a standard
# Z axis name.
xc.get_dim_coords(ds, "Z")

# %%
# However, we can't map to "lev_bnds".
ds.bounds.get_bounds(axis="Z")
# KeyError: "No bounds data variables were found for the 'Z' axis. Make sure the dataset has bound data vars and their names match the 'bounds' attributes found on their related time coordinate variables. Alternatively, you can add bounds with `ds.bounds.add_missing_bounds()` or `ds.bounds.add_bounds()`."

# %%
# Workaround 1 - add CF attribute(s)
ds_cf = ds.copy()
ds_cf["lev"].attrs["axis"] = "Z"
ds_cf.bounds.get_bounds(axis="Z")

# %%
# Workaround 2 - pass the var_key arg to get bounds via "bounds" key
ds.bounds.get_bounds(axis="Z", var_key="lev")
```

### Relevant log output

_No response_

Found this bug through e3sm_diags GH issue [863](https://github.com/E3SM-Project/e3sm_diags/issues/863)

### Environment

Latest xCDAT and `main`

tomvothecoder:
## Description

- Closes #707 

## Checklist

If applicable:

### DIFF
--- a/tests/test_bounds.py
+++ b/tests/test_bounds.py
@@ -287,6 +287,27 @@ class TestGetBounds:
 
         assert result.identical(expected)
 
+    def test_returns_single_dataset_axis_bounds_as_a_dataarray_object_for_non_cf_axis(
+        self,
+    ):
+        ds = xr.Dataset(
+            coords={
+                "lat": xr.DataArray(
+                    data=np.ones(3),
+                    dims="lat",
+                    attrs={"bounds": "lat_bnds"},
+                )
+            },
+            data_vars={
+                "lat_bnds": xr.DataArray(data=np.ones((3, 3)), dims=["lat", "bnds"])
+            },
+        )
+
+        result = ds.bounds.get_bounds("Y")
+        expected = ds.lat_bnds
+
+        assert result.identical(expected)
+
     def test_returns_multiple_dataset_axis_bounds_as_a_dataset_object(self):
         ds = xr.Dataset(
             coords={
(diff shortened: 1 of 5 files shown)


tests/test_bounds.py
```diff
@@ -321,6 +342,41 @@ def test_returns_multiple_dataset_axis_bounds_as_a_dataset_object(self):
 
         assert result.identical(expected)
 
+    def test_returns_multiple_dataset_axis_bounds_as_a_dataset_object_for_non_cf_axis(
+        self,
+    ):
+        # Can only map to "lat" dim name
```
tomvothecoder, line 348:
```suggestion
```

PR #735, opened 2025-02-10
More sophisticated bounds handling for temporal averaging

### LINKED ISSUE #594: [Feature]: More Sophisticated Bounds Handling in Temporal Averaging Operations

xcdat temporal averaging operations currently bin data by the labelled time point with the weights derived from the difference in the time bounds. This works for most conventional climate data: a timepoint of `2020-01-16 12:00` with bounds of `[2020-01-01 00:00, 2020-02-01 00:00]` would be given 31 days of weight in January (e.g., in creating an annual average or climatology), which is correct. 

There are reasonable instances where this wouldn't work. Imagine pentad data with a time point of `2020-02-02 12:00` with bounds of `[2020-01-31 00:00, 2020-02-05 00:00]`. This time point should be given one day of weight in January and four days of weight in February. The current algorithm (e.g., for monthly averaging) assigns all five days of weight in February (the labelled time point). 

Weights should be assigned based on the time period that they fall into. This would mean that a given time point can contribute to averages in more than one time interval 

Solutions for the time being would be to update documentation to note that:

* Weights are determined from the difference in the bounds
* Data is grouped into the labelled time point for averaging operations
* If one time point spans across the time intervals that you are averaging into, then weights are not properly assigned
* `get_time_bounds` generally assumes data with frequencies of annual, monthly, daily, or sub-daily (I thought of this while writing this issue)

### Additional context

I'm not sure cdms / cdutil covers this case; it would be helpful to determine what cdutil does. 

This seems like it could be challenging issue to address in general and might require a major refactor of the logic use for existing temporal averaging calculations.

pochedls:
## Description
This PR has more sophisticated handling of bounds for temporal operations. In particular, it will compare the dataset time bounds with that period that is targeted. For example, if you have a time point with arbitrary bounds of `["2019-12-28 00:00", "2020-01-03 00:00"]` and want to create an average value for January 2020 (i.e., `["2020-01-01", "2020-01-01"]`) this PR will correctly include 2 days of weight in that January average (since only 2 days of data are in January). 

- Closes #594 

## Checklist

If applicable:

### DIFF
--- a/xcdat/temporal.py
+++ b/xcdat/temporal.py
@@ -2,7 +2,7 @@
 
 import warnings
 from datetime import datetime
-from itertools import chain
+from itertools import chain, product
 from typing import Dict, List, Literal, Optional, Tuple, TypedDict, Union, get_args
 
 import cf_xarray  # noqa: F401
@@ -17,7 +17,7 @@ from xarray.core.groupby import DataArrayGroupBy
 
 from xcdat import bounds  # noqa: F401
 from xcdat._logger import _setup_custom_logger
-from xcdat.axis import get_dim_coords
+from xcdat.axis import get_dim_coords, get_dim_keys
 from xcdat.dataset import _get_data_var
 
 logger = _setup_custom_logger(__name__)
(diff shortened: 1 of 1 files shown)


xcdat/temporal.py
```diff
@@ -2025,6 +2025,210 @@ def _calculate_departures(
         return ds_departs
 
 
+    def compute_monthly_average(self, data_var):
```
pochedls, line 2028: This function wraps several steps/functions (that are defined below) in order to compute monthly averages (e.g., from hourly/daily/pentad data to monthly means):
* `ensure_bounds_order`: function ensures that dataset bounds are in order [earlier time, later time] (since PR logic depends on this)
* `generate_monthly_bounds`: function creates monthly bounds
* `get_temporal_weights`: function computes weights for averaging source dataset into targeted time periods
* `_experimental_averager`: function uses temporal weights to average data into targeted time periods

pochedls: I think we could generalize this function (kind of like `.temporal.group_average()`) by having it call different functions to generate target bounds (e.g., `generate_daily_bounds`, `generate_seasonal_bounds`, `generate_yearly_bounds`). The other steps would work as-is.

xcdat/temporal.py
```diff
@@ -2025,6 +2025,210 @@ def _calculate_departures(
...
+        ds.temporal.ensure_bounds_order()
+        # get target time and bounds
+        target_time, target_bnds = ds.temporal.generate_monthly_bounds()
+        # get temporal weights
+        weights = ds.temporal.get_temporal_weights(target_bnds)
+        # compute average and return resulting dataset
+        return ds.temporal._experimental_averager(data_var, weights, target_bnds)
+
+
+    def _experimental_averager(self, data_var, weights, target_bnds):
```
pochedls, line 2066: This is intended to be a generic average averaging data variable information into targeted time periods (using the supplied weights).

xcdat/temporal.py
```diff
@@ -2025,6 +2025,210 @@ def _calculate_departures(
...
+        # and its associated coordinates are also added.
+        dsmean = dsmean.drop_dims(time_key)
+        # add weighted mean data array to output dataset
+        dsmean[data_var] = da_mean
+        # add the time bounds to the dataset
+        dsmean[time_key + '_bnds'] = target_bnds
+        return dsmean
+
+
+    def get_temporal_weights(self, target_bnds):
```
pochedls, line 2113: This function basically gets the intersection between the dataset's own time bounds and the targeted time bounds (i.e., averaging periods). For a given time step, it assigns weight proportional to the duration in which a given timestep is within the a given averaging period.

pochedls: Note that this PR is ~10x slower than existing functionality. The slowdown is almost entirely in this function. If we could speed this step up, that would be great (but we likely can tolerate this slowdown, since the approach in this PR should be more robust/accurate).

xcdat/temporal.py
```diff
@@ -2025,6 +2025,210 @@ def _calculate_departures(
...
+        # Convert weights to xarray DataArray
+        weights = xr.DataArray(
+            data=weights,
+            dims=['target_time', 'time'],
+            coords={'target_time': target_time.values, 'time': ds[time_key].values}
+        )
+        return weights
+
+
+    def generate_monthly_bounds(self):
```
pochedls, line 2161: Prototype function for generating target bounds (i.e., what bins do you want to average your source data into). We could make other functions for other frequencies (e.g., daily, seasonal, yearly).

xcdat/temporal.py
```diff
@@ -2025,6 +2025,210 @@ def _calculate_departures(
...
+        monthly_time.encoding = ds[time_key].encoding
+        target_time = monthly_time.assign_attrs({'bounds': time_key + '_bnds'})
+        monthly_bnds = xr.DataArray(data=monthly_bnds,
+                                    dims=[time_key, 'bnds'],
+                                    coords={time_key: monthly_time})
+        monthly_bnds.encoding = ds[time_key].encoding
+        return monthly_time, monthly_bnds
+
+
+    def ensure_bounds_order(self):
```
pochedls, line 2217: This function just makes sure the bounds are ordered as expected.

pochedls (commented):
A very small start on this issue, but worth going over carefully because if we get the workflow right, we can use this PR (and modifications) as a template for more temporal operations. This mainly handles group averaging operations (we'd need to think more about how to apply this to climatologies, for example).

tomvothecoder:
> I tried to find if there were other packages that provided this feature, but none do (even native Xarray doesn't). This is a great opportunity to fill a gap for more accurate results.

> Re: other packages: Yes...but are these other packages purporting to apply temporal weights? I think what we are doing is slightly different (and requires more sophisticated logic to ensure that we do it correctly).

I'm not sure if I'm following you here. What I meant was I tried to find if any other package that offered #594 just to see if anybody has tried implementing it so far (and to analyze existing logic), but it doesn't like there are any. Does CDAT have this feature? Great opportunity for xCDAT to provide this capability.

PR #752, opened 2025-04-11
Adds option to output regridding weights

### LINKED ISSUE #550: [Feature]: Update regridding to allow preservation of weights in output

Adding an option to keep weights in the output will allow the user to perform validation on them if needed.

Additional work:
* Confirm internal weights are correct by returning them to the user, and also ability to reuse them 
* xESMF returns sparse array, need to figure out how to return weight matrix that matches input data shape
* Still understanding how weights are generated, applied
  * In regrid2 weights are a list of arrays, just need to format them to return 
  * Calculates weight for one time step and apply to every time step, stays the same
* Also Karl’s algorithm to take weight and make them conservative

_No response_

_No response_

### Additional context

_No response_

jasonb5:
## Description

Adds `output_weights` option to `regrid2` and `xesmf` regridders.
Can be set to true/false to include weights as `weights` variable in output dataset.
If it's a `str`, this will be the name of the variable in the output dataset.

The `regrid2` regridder formats it's weights as a sparse array with dimensions (`out_y`, `out_x`, `in_y`, `in_x`).
This matches the behavior of the xesmf regridder. Sparse array is used for space efficiency.

- Closes #550 

## Checklist

If applicable:

### DIFF
--- a/tests/test_regrid.py
+++ b/tests/test_regrid.py
@@ -605,14 +605,21 @@ class TestRegrid2Regridder:
         assert np.all(output_data.ts == 1)
 
     def test_output_weigths(self):
-        regridder = regrid2.Regrid2Regridder(self.coarse_2d_ds, self.fine_2d_ds, output_weights=True)
+        regridder = regrid2.Regrid2Regridder(
+            self.coarse_2d_ds, self.fine_2d_ds, output_weights=True
+        )
 
         output_ds = regridder.horizontal("ts", self.coarse_2d_ds)
 
         assert "weights" in output_ds
-        assert output_ds["weights"].shape ==  self.fine_2d_ds["ts"].shape + self.coarse_2d_ds["ts"].shape
+        assert (
+            output_ds["weights"].shape
+            == self.fine_2d_ds["ts"].shape + self.coarse_2d_ds["ts"].shape
+        )
 
-        regridder = regrid2.Regrid2Regridder(self.coarse_2d_ds, self.fine_2d_ds, output_weights="ts_weights")
+        regridder = regrid2.Regrid2Regridder(
+            self.coarse_2d_ds, self.fine_2d_ds, output_weights="ts_weights"
+        )
 
         output_ds = regridder.horizontal("ts", self.coarse_2d_ds)
 
@@ -760,14 +767,21 @@ class TestXESMFRegridder:
     def test_output_weights(self):
         ds = self.ds.copy()
 
-        regridder = xesmf.XESMFRegridder(ds, self.new_grid, "bilinear", output_weights=True)
+        regridder = xesmf.XESMFRegridder(
+            ds, self.new_grid, "bilinear", output_weights=True
+        )
 
         output = regridder.horizontal("ts", ds)
 
         assert "weights" in output
-        assert output["weights"].shape == (self.new_grid["lat"].shape[0], self.new_grid["lon"].shape[0]) + (ds["lat"].shape[0], ds["lon"].shape[0])
+        assert output["weights"].shape == (
+            self.new_grid["lat"].shape[0],
+            self.new_grid["lon"].shape[0],
+        ) + (ds["lat"].shape[0], ds["lon"].shape[0])
 
-        regridder = xesmf.XESMFRegridder(ds, self.new_grid, "bilinear", output_weights="ts_weights")
+        regridder = xesmf.XESMFRegridder(
+            ds, self.new_grid, "bilinear", output_weights="ts_weights"
+        )
 
         output = regridder.horizontal("ts", ds)
 
--- a/tests/test_regrid.py
+++ b/tests/test_regrid.py
@@ -604,6 +604,20 @@ class TestRegrid2Regridder:
 
         assert np.all(output_data.ts == 1)
 
+    def test_output_weigths(self):
+        regridder = regrid2.Regrid2Regridder(self.coarse_2d_ds, self.fine_2d_ds, output_weights=True)
+
+        output_ds = regridder.horizontal("ts", self.coarse_2d_ds)
+
+        assert "weights" in output_ds
+        assert output_ds["weights"].shape ==  self.fine_2d_ds["ts"].shape + self.coarse_2d_ds["ts"].shape
+
+        regridder = regrid2.Regrid2Regridder(self.coarse_2d_ds, self.fine_2d_ds, output_weights="ts_weights")
+
+        output_ds = regridder.horizontal("ts", self.coarse_2d_ds)
+
+        assert "ts_weights" in output_ds
+
     def test_map_longitude_coarse_to_fine(self):
         mapping, weights = regrid2._map_longitude(
             self.coarse_lon_bnds.values, self.fine_lon_bnds.values
@@ -742,6 +756,22 @@ class TestXESMFRegridder:
         assert "lat_bnds" in output
         assert "lon_bnds" in output
         assert "time_bnds" in output
+
+    def test_output_weights(self):
+        ds = self.ds.copy()
+
+        regridder = xesmf.XESMFRegridder(ds, self.new_grid, "bilinear", output_weights=True)
+
+        output = regridder.horizontal("ts", ds)
+
+        assert "weights" in output
+        assert output["weights"].shape == (self.new_grid["lat"].shape[0], self.new_grid["lon"].shape[0]) + (ds["lat"].shape[0], ds["lon"].shape[0])
+
+        regridder = xesmf.XESMFRegridder(ds, self.new_grid, "bilinear", output_weights="ts_weights")
+
+        output = regridder.horizontal("ts", ds)
+
+        assert "ts_weights" in output
 
     @pytest.mark.parametrize(
         "name,value,_",
--- a/xcdat/regridder/regrid2.py
+++ b/xcdat/regridder/regrid2.py
@@ -32,7 +32,12 @@ class Regrid2Regridder(BaseRegridder):
             Dataset containing the source grid.
         output_grid : xr.Dataset
             Dataset containing the destination grid.
-        options : Any
+        unmapped_to_nan : bool
+            If True, unmapped values are set to NaN. Default is True.
+        output_weights : Union[bool, str]
+            If True, output weights are added to the output dataset as weights.
+            If str, the name of the variable to store the weights. Default is False.
+        **options : Any
             Dictionary with extra parameters for the regridder.
 
         Examples
--- a/xcdat/regridder/regrid2.py
+++ b/xcdat/regridder/regrid2.py
@@ -1,6 +1,7 @@
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
+import sparse as sp
 import xarray as xr
 
 import xcdat as xc
@@ -13,7 +14,8 @@ class Regrid2Regridder(BaseRegridder):
         self,
         input_grid: xr.Dataset,
         output_grid: xr.Dataset,
-        unmapped_to_nan=True,
+        unmapped_to_nan: bool = True,
+        output_weights: Union[bool, str] = False,
         **options: Any,
     ):
         """
@@ -55,6 +57,7 @@ class Regrid2Regridder(BaseRegridder):
         super().__init__(input_grid, output_grid, **options)
 
         self._unmapped_to_nan = unmapped_to_nan
+        self._output_weights = output_weights
 
     def vertical(self, data_var: str, ds: xr.Dataset) -> xr.Dataset:
         """Placeholder for base class."""
@@ -91,13 +94,16 @@ class Regrid2Regridder(BaseRegridder):
         # exclude alternative of NaN values if there are any
         input_data_var = input_data_var.where(input_data_var != nan_replace)
 
+        lat_mapping, lat_weights = _map_latitude(src_lat_bnds, dst_lat_bnds)
+        lon_mapping, lon_weights = _map_longitude(src_lon_bnds, dst_lon_bnds)
+
         # horizontal regrid
         output_data = _regrid(
             input_data_var,
-            src_lat_bnds,
-            src_lon_bnds,
-            dst_lat_bnds,
-            dst_lon_bnds,
+            lat_mapping,
+            lon_mapping,
+            lat_weights,
+            lon_weights,
             src_mask,
             unmapped_to_nan=self._unmapped_to_nan,
         )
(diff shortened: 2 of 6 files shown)


jasonb5:
@pochedls @lee1043 @chengzhuzhang @tomvothecoder I wanted to add one more feature to this PR but want to get a consensus first. I want to update the grid functions to not just output a dataset with spatial coords/dimensions but also include a `mask` variable. I'm curious if this makes sense from a user perspective.

xcdat/regridder/xesmf.py
```diff
@@ -155,6 +160,7 @@ def horizontal(self, data_var: str, ds: xr.Dataset) -> xr.Dataset:
                 f"The data variable '{data_var}' does not exist in the dataset."
             )
 
+        # import pdb; pdb.set_trace()
```
tomvothecoder, line 163:
```suggestion
```

lee1043:
I think this function is useful to have for testing purpose. With the function it would be easier to compare weightings of different regridders. This might be also helpful for high resolution data where regridding can take longer than coarser grid. A following question would be, can the regridder accept re-use of pre-calculated weights? I support this addition unless it significantly drop the performance (which I think it is very unlikely). Thank you for the addition, @jasonb5!

tests/test_regrid.py
```diff
@@ -604,6 +604,27 @@ def test_regrid_4d(self):
 
         assert np.all(output_data.ts == 1)
 
+    def test_output_weigths(self):
```
tomvothecoder, line 607:
```suggestion
    def test_output_weights(self):
```

tomvothecoder (commented):
Hi @jasonb5, this PR looks good to me. Thank you!

RE: Additional enhancements

> I want to update the grid functions to not just output a dataset with spatial coords/dimensions but also include a mask variable. I'm curious if this makes sense from a user perspective.

This could be useful for debugging purposes. We had a situation in `e3sm_diags` where we needed to add a `mask` variable to improve regridding accuracy with `"conservative_normed"` (xESMF [docs](https://xesmf.readthedocs.io/en/latest/notebooks/Masking.html#Regridding-with-a-mask)).  @chengzhuzhang can weigh in too.

> A following question would be, can the regridder accept re-use of pre-calculated weights? I support this addition unless it significantly drop the performance (which I think it is very unlikely). Thank you for the addition, @jasonb5!

I'm curious, what are some scenarios where pre-calculated weights might be used/preferred over internally generated weights?

@pochedls and @chengzhuzhang can weigh in on the usefulness of this enhancement.

tomvothecoder (approved):

pochedls:
> I'm curious, what are some scenarios where pre-calculated weights might be used/preferred over internally generated weights?
> 
> @pochedls and @chengzhuzhang can weigh in on the usefulness of this enhancement.

I don't know if I would use this frequently, but I could see it being useful and a reasonable feature to include. I think if this is a matter of just adding this as an input/output, then it isn't a big deal to add, but if it requires more work it might make sense to tackle this on a longer time horizon with a specific issue/PR.

tomvothecoder:
> > I'm curious, what are some scenarios where pre-calculated weights might be used/preferred over internally generated weights?
> > @pochedls and @chengzhuzhang can weigh in on the usefulness of this enhancement.
> 
> I don't know if I would use this frequently, but I could see it being useful and a reasonable feature to include. I think if this is a matter of just adding this as an input/output, then it isn't a big deal to add, but if it requires more work it might make sense to tackle this on a longer time horizon with a specific issue/PR.

Jill mentioned in our 1:1 meeting that it might be performant to re-use the internally generated weights (outputted via this PR) for variables on the same grid. 

I think xCDAT's xESMF regridder already supports this feature by inheriting xESMF's `weights` parameter to use custom weights ([link](https://github.com/pangeo-data/xESMF/blob/892ac87064d98d98d732ad8a79aa1682b081cdc2/xesmf/frontend.py#L727-L734)). I have not tested this yet.

I suggest we open a separate GitHub issue and PR to tackle this enhancement for `Regrid2` if we think it will be useful.

jasonb5:
I'm going to go ahead and add the `mask` variable to the grid dataset returned from `ds.regridder.grid` property. This will simplify a few lines of code in this PR.

As for re-using weights as Tom mentioned xESMF already supports this and it would be easy to add for Regrid2. I can create an new issue to discuss this.

xcdat/regridder/grid.py
```diff
@@ -532,9 +532,22 @@ def create_grid(
...
 
+def create_mask(ds: xr.Dataset, dims: Optional[List[CFAxisKey]] = None) -> xr.DataArray:
+    if dims is None:
+        dims = ["X", "Y", "Z"]
+
+    dims = list(dims)
+
+    mask_shape = {ds.cf[x].name: ds.cf[x].shape[0] for x in dims if x in ds.cf.axes}
+
+    return xr.DataArray(np.ones(list(mask_shape.values())), dims=list(mask_shape))
```
tomvothecoder, line 548: Since `create_mask()` is a public function, we should add some direct unit tests (in addition to the one you added).

If I understand correctly, this function returns a `mask` DataArray with same dimensions and shape as the grid and filled with ones (via `np.ones`). Is this function the same as the implementation in the xESMF [example](https://xesmf.readthedocs.io/en/latest/notebooks/Masking.html#Regridding-with-a-mask)? In the xESMF example, they create a `mask` variable based on the values of the `zeta` variable, which can include `np.nan`.

tomvothecoder: @chengzhuzhang Can you review the above code too?

tomvothecoder (commented):
Hi @jasonb5, I had some review comments.

xcdat/regridder/grid.py
jasonb5: @tomvothecoder I can add tests specifically for this. This code is only there to create a default mask that passes through all data, it makes no assumptions on what the user intends to do. It's up the user to modify it afterwards or overwrite like the xESMF example. Internally this more of a convenience thing, as it provides the ability to get the shape of the grid using `ds.mask.shape` which was previously more involved with something like `(ds.lat.shape[0], ds.lon.shape[0])`.

I think a better approach to accomplish the xESMF example would be something along the following:
``` python
ds = xc.open_dataset(...)
ds.regridder.create_mask_from_nans("zeta")
ds["mask"].plot()
```
@chengzhuzhang I think it's really just a preference thing. I went with (0, 1) since xESMF/esmf uses this notation. Not sure if this is applicable in this domain and would depend on the implementation but (0, 1) offers the ability to let values "bleed through" but using fractional values.

tomvothecoder: > This code is only there to create a default mask that passes through all data, it makes no assumptions on what the user intends to do. It's up the user to modify it afterwards or overwrite like the xESMF example. Internally this more of a convenience thing, as it provides the ability to get the shape of the grid using `ds.mask.shape` which was previously more involved with something like `(ds.lat.shape[0], ds.lon.shape[0])`.

Got it, I see what you're saying. It also makes it easier to get the shape of the grid compared to `ds.regridder.grid.shape`. 

If the values of the default mask are `np.ones` and the user does not update the mask, does xESMF ignore the `mask` variable as a no-operation? Just curious if xESMF might interpret the `mask` incorrectly somehow with `np.ones` (when the user doesn't update the values).

### Alternative Solution
> ```python
> ds = xc.open_dataset(...)
> ds.regridder.create_mask_from_nans("zeta")
> ds["mask"].plot()
> ```
> 
> @chengzhuzhang I think it's really just a preference thing. I went with (0, 1) since xESMF/esmf uses this notation. Not sure if this is applicable in this domain and would depend on the implementation but (0, 1) offers the ability to let values "bleed through" but using fractional values.

An alternative solution is to add a flag like `create_mask` where the `mask` variable is created directly from the variable being operated on. The user wouldn't have to worry about creating/updating the `mask` variable downstream and it ensures the correct `mask` is created. 

For example:
```python
ds = xc.open_dataset(...)
ds_regrid = ds.regridder.horizontal("zeta", ...create_mask=True)
ds_regrid["mask"].plot()
```

Any thoughts @chengzhuzhang @xCDAT/core-developers?

jasonb5: > Got it, I see what you're saying. It also makes it easier to get the shape of the grid compared to ds.regridder.grid.shape.
>
> If the values of the default mask are np.ones and the user does not update the mask, does xESMF ignore the mask variable as a no-operation? Just curious if xESMF might interpret the mask incorrectly somehow with np.ones (when the user doesn't update the values).

I believe xESM will use it as long as it's there. I don't think there would be any side effect from always having it as ones. I would need to look but if it uses masks like `regrid2` then the math is probably something like `sum(var * mask) / ...` so all ones would let all the values through.

> An alternative solution is to add a flag like create_mask where the mask variable is created directly from the variable being operated on. The user wouldn't have to worry about creating/updating the mask variable downstream and it ensures the correct mask is created.
>
> For example:
> 
> ```
> ds = xc.open_dataset(...)
> ds_regrid = ds.regridder.horizontal("zeta", ...create_mask=True)
> ds_regrid["mask"].plot()
> ```

I think this would be a nice option, I'm between `create_mask` and a little more verbose `create_nan_mask`. If it's common to make a mask from nans we could just enable it by default as well.

tomvothecoder: @jasonb5 Got it. I think we should split out this `"mask"` feature as a another GitHub issue/PR since it is a separate enhancement. We can think more about about how to create the mask variable there too.

That way we can merge this PR focused on outputting regridding weights to release v0.9.0 soon.

tomvothecoder: I just noticed you pushed commits for the `"mask"` feature actually. Since it look almost complete, we can include it in this PR.

jasonb5:
@tomvothecoder Is it fine to merge now?

xcdat/regridder/regrid2.py
```diff
@@ -75,13 +89,21 @@ def horizontal(self, data_var: str, ds: xr.Dataset) -> xr.Dataset:
...
-            src_mask = None
+            src_mask = xr.where(
+                np.isnan(input_data_var.isel(**non_yx_selector)), 0, 1
+            ).values
+        else:
+            # DataArray to np.ndarray, handle error when None
+            try:
+                src_mask = self._input_grid.get("mask", None).values  # type: ignore
+            except AttributeError:
+                src_mask = create_mask(self._input_grid, ["Y", "X"]).values
```
tomvothecoder, line 106: I think the `create_nan_mask` parameter and this logic should also be in the xESMF Regridder for more accurate regridding.

xcdat/regridder/xesmf.py
```diff
@@ -155,6 +160,12 @@ def horizontal(self, data_var: str, ds: xr.Dataset) -> xr.Dataset:
                 f"The data variable '{data_var}' does not exist in the dataset."
             )
 
+        # align the grid dimension to input data ordering
+        if "mask" in self._output_grid:
+            self._output_grid["mask"] = self._output_grid.mask.transpose(
+                *input_da.dims, missing_dims="ignore"
+            )
+
```
tomvothecoder, line 168: This is where the mask logic from Regrid2 should also be added

xcdat/regridder/xesmf.py
```diff
@@ -77,6 +78,9 @@ def __init__(
             regridding methods.
         unmapped_to_nan : bool
             Sets values of unmapped points to `np.nan` instead of 0 (ESMF default).
+        output_weights : Union[bool, str]
+            If True, output weights are added to the output dataset as weights.
+            If str, the name of the variable to store the weights. Default is False.
```
tomvothecoder, line 83: Add `create_nan_from_mask` parameter

tomvothecoder (commented):
Hey @jasonb5, I had some review comments. I think we need to add the same `create_nan_mask` logic in the Regrid2 code to the xESMF regridder too. That way xESMF can use the correct "mask" variable from the source variable for more accurate regridding. Right now it just transposes any existing "mask" variable. 

Once this is implemented with tests, you can merge.

jasonb5:
> Hey @jasonb5, I had some review comments. I think we need to add the same `create_nan_mask` logic in the Regrid2 code to the xESMF regridder too. That way xESMF can use the correct "mask" variable from the source variable for more accurate regridding. Right now it just transposes any existing "mask" variable.
> 
> Once this is implemented with tests, you can merge.

This is fine, I can add it now. Didn't know if we wanted to address this in a later PR.

jasonb5:
@tomvothecoder  This is done once the CI passes.

xcdat/regridder/regrid2.py
```diff
@@ -30,7 +34,15 @@ def __init__(
...
             Dataset containing the destination grid.
-        options : Any
+        unmapped_to_nan : bool
+            If True, unmapped values are set to NaN. Default is True.
+        output_weights : bool | str
+            If True, output weights are added to the output dataset as weights.
+            If str, the name of the variable to store the weights. Default is False.
+        create_nan_mask : bool
+            If True, a mask is created using the nan values from source variable. If
+            a mask already exists in the Dataset it will be ignored.
```
tomvothecoder, line 44:
```suggestion
            If True, a mask is created using the nan values from source variable. If
            a mask already exists in the Dataset it will be ignored. Default is False.
```

xcdat/regridder/xesmf.py
```diff
@@ -77,6 +80,12 @@ def __init__(
             regridding methods.
         unmapped_to_nan : bool
             Sets values of unmapped points to `np.nan` instead of 0 (ESMF default).
+        output_weights : bool | str
+            If True, output weights are added to the output dataset as weights.
+            If str, the name of the variable to store the weights. Default is False.
+        create_nan_mask : bool
+            If True, a mask is created using the nan values from source variable. If
+            a mask already exists in the Dataset it will be ignored.
```
tomvothecoder, line 88:
```suggestion
        create_nan_mask : bool
            If True, a mask is created using the nan values from source variable. If
            a mask already exists in the Dataset it will be ignored. Default is False.
```

xcdat/regridder/grid.py
```diff
@@ -528,9 +529,86 @@ def create_grid(
...
+    """
+    Create a mask as an `xarray.DataArray` based on the specified dimensions.
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+        The input xarray Dataset containing the data and coordinate information.
+    dims : list[CFAxisKey] or None
+        A list of dimension keys to include in the mask. If not provided,
+        defaults to ["X", "Y", "Z"].
```
tomvothecoder, line 547:
```suggestion
        A list of dimension keys to include in the mask. If not provided,
        defaults to None.
```

Issue #748, opened 2025-04-01
[Bug]: Coordinate ValueError using .group_average


pochedls:
### What happened?

I have a dataset that has a dataarray with dimensions of `[member, time]`. When I try to compute the annual average values, I get:

> ValueError: cannot add coordinates with new dimensions to a DataArray

### Minimal Complete Verifiable Example (MVCE)

```python
import xcdat as xc
import xarray as xr

# tutorial data
ds = xc.tutorial.open_dataset("tas_amon_access")
dsc = xr.concat((ds, ds, ds), dim='member')
dsc.temporal.group_average('tas', freq='year')

# other data
fn = '/p/user_pub/climate_work/pochedley1/extreme2324/tlt90_Amon_E3SM-2-0_historical-ssp370_ensemble_gr_187001-202412.nc'
ds = xc.open_dataset(fn)
ds.temporal.group_average('tlt90', freq='year')
```

### Relevant log output

```python
---------------------------------------------------------------------------                                                                                                                                
ValueError                                Traceback (most recent call last)
Cell In[10], line 1
----> 1 dsc.temporal.group_average('tas', freq='year')

File ~/code/xcdat/xcdat/temporal.py:419, in TemporalAccessor.group_average(self, data_var, freq, weighted, keep_weights, season_config)
    253 """Returns a Dataset with average of a data variable by time group.
    254 
    255 Data is grouped into the labeled time point for the averaging operation.
   (...)    415 }
    416 """
    417 self._set_data_var_attrs(data_var)
--> 419 return self._averager(
    420     data_var,
    421     "group_average",
    422     freq,
    423     weighted=weighted,
    424     keep_weights=keep_weights,
    425     season_config=season_config,
    426 )

File ~/code/xcdat/xcdat/temporal.py:879, in TemporalAccessor._averager(self, data_var, mode, freq, weighted, keep_weights, reference_period, season_config)
    877     dv_avg = self._average(ds, data_var)
    878 elif self._mode in ["group_average", "climatology", "departures"]:
--> 879     dv_avg = self._group_average(ds, data_var)
    881 # The original time dimension is dropped from the dataset because
    882 # it becomes obsolete after the data variable is averaged. When the
    883 # averaged data variable is added to the dataset, the new time dimension
    884 # and its associated coordinates are also added.
    885 ds = ds.drop_dims(self.dim)

File ~/code/xcdat/xcdat/temporal.py:1504, in TemporalAccessor._group_average(self, ds, data_var)
[...]

PR #773, opened 2025-06-09
Update `get_bounds_dim()` to dynamically get correct dim

### LINKED ISSUE #772: [Enhancement]: Improve identification of bounds dimension for weight calculation using CF metadata.

Related to https://github.com/E3SM-Project/e3sm_diags/pull/984#issuecomment-2940902776

An integration test with E3SM diagnostics for the `arm_diags` began failing with v0.9.0. During the reference climatology calculation, xCDAT fails to recognize `nv` as a valid bounds dimension name when attempting to calculate weights with the time bounds.
The valid dimension names are listed below:

https://github.com/xCDAT/xcdat/blob/6bf88851213e9dd6fb90f9f241e655f4f7f93a9b/xcdat/bounds.py#L27-L29

The CF Convention lists `nv` as a bounds dimension name, which suggests that it is CF compliant. However, the CF Convention also it "does not standardize any variable or dimension names", which makes hard-coding bounds dimensions might be tricky.

Add `nv` to `VALID_BOUNDS_DIMS` and consider adding other potential common dim names.

_No response_

### Additional context

_No response_

tomvothecoder:
## Description

- Closes #772 

This PR updates `get_bounds_dim()` to dynamically retrieve the correct dimension. It compares the dimensions of the coordinates against coordinate bounds and extracts the difference in dimensions. Coordinate bounds typically have 1 more dimension for the bounds dimension.

For example
```python
time: (time)
time_bnds: (time, nv)

Result
------
"nv" is extracted 
```

This should also work for cases where coordinates have multiple dimensions, related to #748, #749.

```python
time: (member, time)
time_bnds: (member, time, bnds)

Result
--------
"bnds" is extracted
```

### Can't rely on CF Metadata
CF-xarray does not distinguish dimensions using CF metadata, only coordinates and data variables. Also datasets can sometimes have bounds dimensions without bounds coordinates, which means no CF metadata is available.

Example `ds.cf` -- no dimensions identified to use for bounds dim

```python
Coordinates:
             CF Axes: * Z: ['lev']
                      * T: ['time']
                        X, Y: n/a

      CF Coordinates: * time: ['time']
                        longitude, latitude, vertical: n/a

       Cell Measures:   area, volume: n/a

      Standard Names:   n/a

              Bounds:   n/a

       Grid Mappings:   n/a

Data Variables:
       Cell Measures:   area, volume: n/a

      Standard Names:   n/a

              Bounds:   T: ['climatology_bnds']
                        time: ['climatology_bnds']

       Grid Mappings:   n/a
```

## Checklist

If applicable:

### DIFF
--- a/xcdat/bounds.py
+++ b/xcdat/bounds.py
@@ -1005,7 +1005,7 @@ def create_bounds(axis: CFAxisKey, coord_var: xr.DataArray) -> xr.DataArray:
     return bounds
 
 
-def get_bounds_dim(da_coords: xr.DataArray, da_bounds: xr.DataArray) -> str:
+def _get_bounds_dim(da_coords: xr.DataArray, da_bounds: xr.DataArray) -> str:
     """Identify the bounds dimension in the given bounds DataArray.
 
     This function determines which dimension in the bounds DataArray
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -28,7 +28,6 @@ Below is a list of top-level API functions that are available in ``xcdat``.
     get_dim_coords
     get_dim_keys
     get_coords_by_name
-    get_bounds_dim
     create_bounds
     create_axis
     create_gaussian_grid
--- a/tests/test_bounds.py
+++ b/tests/test_bounds.py
@@ -12,7 +12,7 @@ from tests.fixtures import (
     lat_bnds,
     lon_bnds,
 )
-from xcdat.bounds import BoundsAccessor, get_bounds_dim
+from xcdat.bounds import BoundsAccessor, _get_bounds_dim
 
 logger = logging.getLogger(__name__)
 
@@ -1066,7 +1066,7 @@ class TestGetBoundsDim:
         coords = xr.DataArray([0, 1, 2], dims=["lat"])
         bounds = xr.DataArray([[0, 1], [1, 2], [2, 3]], dims=["lat", "bnds"])
 
-        result = get_bounds_dim(coords, bounds)
+        result = _get_bounds_dim(coords, bounds)
 
         assert result == "bnds"
 
@@ -1074,7 +1074,7 @@ class TestGetBoundsDim:
         coords = xr.DataArray([10, 20, 30], dims=["lon"])
         bounds = xr.DataArray([[5, 15], [15, 25], [25, 35]], dims=["lon", "boundaries"])
 
-        result = get_bounds_dim(coords, bounds)
+        result = _get_bounds_dim(coords, bounds)
 
         assert result == "boundaries"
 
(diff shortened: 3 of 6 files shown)


xcdat/bounds.py
```diff
@@ -1037,6 +1043,14 @@ def get_bounds_dim(da_bounds: xr.DataArray) -> str:
...
+    # If only one bounds dimension is found, return it.
+    bounds_dim = [dim for dim in da_bounds.dims if dim not in da_coords.dims]
+
+    if len(bounds_dim) == 1:
+        return str(bounds_dim[0])
+
+    # FIXME: Do we still need this logic, or do we always expect bounds to have
+    # one additional dimension?
     for dim in VALID_BOUNDS_DIMS:
         if dim in da_bounds.dims:
```
tomvothecoder, line 1055: Hey @pochedls, do you recall if there is ever a case where coordinate bounds have >1 dimension more than their coordinates?

For example:

```python
time: (time)
time_bnds: (time, bnds, some_other_dim)
```

If not, then we can remove the old-logic with the hard-coded list.

Refer to [PR description](https://github.com/xCDAT/xcdat/pull/773#issue-3131611627) on how this works now.

tomvothecoder: From my reasoning, bounds should only have 1 more dimension than their coordinates.

For example
```python
time: (time)
time_bnds: (time, nv)

Result
------
"nv" is extracted 
```

```python
time: (member, time)
time_bnds: (member, time, bnds)

Result
--------
"bnds" is extracted
```

pochedls: @tomvothecoder – sorry for the slow response. I don't know how I missed your comment three weeks ago. I did not understand the comment/question, but I now do. I think this logic makes sense. I haven't seen a `bound` dim in `coords` before, but I wonder if it does happen sometimes?

xcdat/bounds.py
```diff
@@ -1037,6 +1043,14 @@ def get_bounds_dim(da_bounds: xr.DataArray) -> str:
...
     representing bounds, which is commonly used in geospatial or temporal data.
     """
+    # If only one bounds dimension is found, return it.
+    bounds_dim = [dim for dim in da_bounds.dims if dim not in da_coords.dims]
+
+    if len(bounds_dim) == 1:
+        return str(bounds_dim[0])
+
+    # FIXME: Do we still need this logic, or do we always expect bounds to have
+    # one additional dimension?
```
pochedls, line 1053: For what it's worth – I looped over a bunch of CMIP data and never made into this if statement.

xcdat/bounds.py
```diff
@@ -1037,6 +1043,14 @@ def get_bounds_dim(da_bounds: xr.DataArray) -> str:
     This function is designed to work with DataArrays that include a dimension
     representing bounds, which is commonly used in geospatial or temporal data.
     """
+    # If only one bounds dimension is found, return it.
+    bounds_dim = [dim for dim in da_bounds.dims if dim not in da_coords.dims]
+
+    if len(bounds_dim) == 1:
+        return str(bounds_dim[0])
```
pochedls, line 1050:
```suggestion
    if len(bounds_dim) == 1:
        # get dim name
        bdim = str(bounds_dim[0])
        # check if it has len(2)
        if len(da_bounds[bdim]) == 2:
            return bdim
```

Should we also enforce that the `bounds_dim` has len(2)?

pochedls (approved):

PR #827, opened 2026-01-28
Replace in-place `.values` assignment with `assign_coords()`

### LINKED ISSUE #826: [Bug]: Assigning values to dim coords with `.values[]` and `.data[]` is breaking axis swapping and time coord shifting
## What happened?

Assigning values to specific indices in a **dimension coordinate** breaks using Xarray `.values[]` or `.data[]`.

`ValueError: assignment destination is read-only`
  
  Simple example:
  ```python
  # Breaks because .lon is a dimension coordinate and the underlying `.values` numpy array is not writeable
  ds.lon.values[-1] = 360
  ```

## What did you expect to happen? Are there are possible answers you came across?

I checked combinations of xarray versions 2025.09.0-2026.01.0, xcdat 0.10.0-0.11.1, python 3.13-3.14, and they all break. It is unclear to me why this is happening now. 

What is clear is that this error comes from NumPy when trying to assign into an array whose memory is read-only (e.g., a view of disk-backed data, memmap, bytes buffer, or xarray/netCDF-loaded array). The observed fix is avoiding in-place mutation of read-only views (especially `.values` from xarray).

Might be related to 
- https://docs.xarray.dev/en/stable/whats-new.html#id14

  > When assigning an indexed coordinate to a data variable or coordinate, coerce it from IndexVariable to Variable ([GH9859](https://github.com/pydata/xarray/issues/9859), [GH10829](https://github.com/pydata/xarray/issues/10829), [PR10909](https://github.com/pydata/xarray/pull/10909)). By [Julia Signell](https://github.com/jsignell).
- https://github.com/pydata/xarray/issues/10670

## Minimal Complete Verifiable Example (MVCE)

Breaks with:
- Python 3.13 - 3.14
- Xarray 2025.09.0 - 2026.01.0
- xCDAT 0.10.0 - 0.11.1
- Not related to Dask as not installing Dask still results in the error

Try:
- [x] Older version of NumPy -- breaks with 2.1.3, 2.2.6, 2.3.5

```python
# %%
import xarray as xr
import numpy as np

# Create a dummy xarray dataset with the specified structure
ds = xr.Dataset({
    "lon": ("lon", [180, 359, 0, 1, 179, 0]),
    "lon_bnds": (("lon", "bnds"), np.random.rand(6, 2)),
})

# .values[] - `ValueError: assignment destination is read-only`
ds["lon"].values[-1] = 360

# .data[] - `ValueError: assignment destination is read-only`
ds["lon"].data[-1] = 360

```

I found that the underlying numpy array for dimension coordinates has a flag called `WRITEABLE` which is set to `False`.
If this flag is set to True, writing is allowed. However, this is not a solution and we must avoid mutating the array this way. 

```python
# %%
arr = ds["lon"].values

# %% -- WRITEABLE is False
print(arr.flags)
# C_CONTIGUOUS : True
# F_CONTIGUOUS : True
# OWNDATA : False
# WRITEABLE : False
# ALIGNED : True
# WRITEBACKIFCOPY : False

# Fails with `ValueError: assignment destination is read-only`
arr[-1] = 360

# Update the write flag to True
arr.setflags(write=True)

# Now works
arr[-1] = 360
```

### **1. Copy + `assign_coords` (best, safest)**

```python
values = ds["lon"].values.copy()
values[-1] = 360.0

new_coord = ds["lon"].copy(data=values)
ds_lon = ds.assign_coords({"lon": new_coord})
```

**Why:** future-proof, xarray-approved, preserves metadata (e.g., attrs)

### **2. `DataArray.copy(data=...)` + `assign_coords`**

```python
lon = ds["lon"].copy(data=ds["lon"].values.copy())
lon[-1] = 360
ds = ds.assign_coords(lon=lon)
```

**Why:** cleaner if you’re already working with a `DataArray`.

### **3. Vectorized replacement (`where`, `assign_coords`)**

```python
ds = ds.assign_coords(lon=ds.lon.where(ds.lon != ds.lon[-1], 360))
```

**Why:** no indexing, works with dask, good for conditional changes.

Affected code
- https://github.com/xCDAT/xcdat/blob/dad3e8045e98cecc863c5660c0157e137056449e/xcdat/axis.py#L515-L519
- https://github.com/xCDAT/xcdat/blob/dad3e8045e98cecc863c5660c0157e137056449e/xcdat/temporal.py#L1314-L1323
- https://github.com/xCDAT/xcdat/blob/dad3e8045e98cecc863c5660c0157e137056449e/xcdat/temporal.py#L1363-L1372

### Environment

Latest xCDAT versions and main branch

tomvothecoder:
## Description

Closes #826

This pull request addresses a bug where assigning values to dimension coordinates using `.values[]` or `.data[]` breaks due to read-only arrays in recent versions of xarray. The fix replaces in-place mutations with proper copy-and-assign patterns using `assign_coords()`.

**Changes:**
- Refactored temporal shifting methods to use `xr.where()` and `xr.apply_ufunc()` instead of in-place array mutations
- Updated axis alignment to use coordinate copying and `assign_coords()` instead of direct value assignment
- Fixed test fixtures to use proper coordinate assignment patterns instead of in-place mutations

## Checklist

If applicable:

### DIFF
--- a/xcdat/temporal.py
+++ b/xcdat/temporal.py
@@ -1357,23 +1357,56 @@ class TemporalAccessor:
         >>>  (2001, "DJF", 1), (2001, "DJF", 2)]
         """
         ds_new = ds.copy()
-        time_coords = ds_new[self.dim].copy()
-        dec_indexes = time_coords.dt.month == 12
+        time_coords = ds_new[self.dim]
+
+        is_december = time_coords.dt.month == 12
 
         if isinstance(time_coords.values[0], cftime.datetime):
-            time_coords.values[dec_indexes] = [
-                time.replace(year=time.year + 1)
-                for time in time_coords.values[dec_indexes]
-            ]
+            shift_func = self._shift_cftime_year
         else:
-            time_coords.values[dec_indexes] = [
-                pd.Timestamp(time) + pd.DateOffset(years=1)
-                for time in time_coords.values[dec_indexes]
-            ]
+            shift_func = self._shift_datetime_year
 
-        ds_new = ds_new.assign_coords({self.dim: time_coords})
+        shifted_time = xr.apply_ufunc(shift_func, time_coords, vectorize=True)
+        shifted_time_dec_only = xr.where(is_december, shifted_time, time_coords)
+        shifted_time_dec_only.encoding = time_coords.encoding.copy()
+
+        ds_new = ds_new.assign_coords({self.dim: shifted_time_dec_only})
 
         return ds_new
+
+    def _shift_cftime_year(self, time: cftime.datetime) -> cftime.datetime:
+        """
+        Shift the year of a cftime.datetime object by 1.
+
+        Parameters
+        ----------
+        time : cftime.datetime
+            The cftime.datetime object to shift.
+
+        Returns
+        -------
+        cftime.datetime
+            The cftime.datetime object with the year incremented by 1.
+        """
+        return time.replace(year=time.year + 1)
+
+    def _shift_datetime_year(self, time) -> pd.Timestamp:
+        """
+        Shift the year of a datetime-like object by 1.
+
+        Parameters
+        ----------
+        time : datetime-like
+            The datetime-like object to shift.
+
+        Returns
+        -------
+        pd.Timestamp
+            The datetime-like object with the year incremented by 1.
+        """
+        ts = pd.Timestamp(time)
+
+        return ts.replace(year=ts.year + 1)
 
     def _drop_incomplete_djf(self, dataset: xr.Dataset) -> xr.Dataset:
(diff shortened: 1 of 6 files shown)


xcdat/temporal.py
```diff
@@ -1357,23 +1354,73 @@ def _shift_djf_decembers(self, ds: xr.Dataset) -> xr.Dataset:
         >>>  (2001, "DJF", 1), (2001, "DJF", 2)]
         """
         ds_new = ds.copy()
-        time_coords = ds_new[self.dim].copy()
-        dec_indexes = time_coords.dt.month == 12
+        time_coords = ds_new[self.dim]
```
tomvothecoder:
```suggestion
        time_coords = ds_new[self.dim].copy()
```

PR #828, opened 2026-02-02
Update `cftime` section in FAQs

### LINKED ISSUE #825: [Doc]: cftime documentation

I believe the [FAQs](https://xcdat.readthedocs.io/en/latest/getting-started-guide/faqs.html) could be updated regarding the advantages of `cftime` over `datetime64`.

> Why does xcdat decode time coordinates as cftime objects instead of datetime64[ns]?
> 
> One unfortunate limitation of using datetime64[ns] is that it limits the native representation of dates to those that fall between the years 1678 and 2262. This affects climate modeling datasets that have time coordinates outside of this range.
> 
> As a workaround, xarray uses the cftime library when decoding/encoding datetimes for non-standard calendars or for dates before year 1678 or after year 2262.

As the references note, pandas has relaxed the `datetime64[ns]` limitation since version 2, so this is no longer the main reason for choosing `cftime`. The FAQ goes on to describe some of the advantages of `cftime` over `datetime`, independent of datetime resolution:

> xcdat opted to decode time coordinates exclusively with cftime because it has no timestamp range limitations, simplifies implementation, and the output object type is deterministic. Another benefit with this approach is that xcdat has its own algorithm for lazily decoding cftime objects, which improves up-front I/O performance.

I think that emphasizing these resolution-independent reasons for using `cftime` up-front would be clearer here.

[I was also wondering -- perhaps there are other benefits like allowing different calendar types?]

tomvothecoder:
## Description

- Closes #825 

## Checklist

If applicable:

### DIFF
--- a/docs/getting-started-guide/faqs.rst
+++ b/docs/getting-started-guide/faqs.rst
@@ -135,26 +135,30 @@ References:
 
 * https://cfconventions.org/cf-conventions/cf-conventions#calendar
 
-Why does ``xcdat`` decode time coordinates as ``cftime`` objects instead of ``datetime64[ns]``?
+Why does ``xcdat`` decode time coordinates as ``cftime`` objects instead of ``datetime64``?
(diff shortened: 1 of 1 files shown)


arfriedman:
Great!

